PR #4, opened 2019-04-19
Feature: Added support for custom geometry column names when reading …

henrymartin1:
…postgis databases.
I also included args/kwargs in the geopandas.from_postgis() call. This allows for example to set a custom crs/srid


henrymartin1:
test

PR #8, opened 2019-12-05
Pull request

henrymartin1:
Large combined pull request:
Highlights: 

- Enhanced tripleg generation (from point to point)
- A working binder setup (from Rene)
-


trackintel/preprocessing/positionfixes.py
```diff
@@ -326,3 +366,49 @@ def extract_triplegs(positionfixes, staypoints=None, *args, **kwargs):
     ret_triplegs['id'] = ret_triplegs['id'].astype('int')
 
     return ret_triplegs
+
+
+def propagate_tripleg(pfs, stp, position_edge_posfix_tl, direction=1):
```
dominikbucher: Don't remember if we have a convention for internal functions (start with _?; is this even an internal function?), but otherwise a small docstring could be helpful, esp. if this function turns up in the (automatically generated) documentation at some point.

trackintel/geogr/distances.py
```diff
@@ -79,13 +79,18 @@ def calculate_distance_matrix(points, dist_metric='haversine', n_jobs=None, *arg
          xy = np.concatenate((x.reshape(-1,1),y.reshape(-1,1)),axis=1)
          D = cdist(xy,xy,metric=haversine_dist_cdist)
         
+        
     else:
         xy = np.concatenate((x.reshape(-1,1),y.reshape(-1,1)),axis=1)
         D = pairwise_distances(xy, metric=dist_metric, n_jobs=n_jobs)
         
+
```
dominikbucher: Many blank lines. See https://www.python.org/dev/peps/pep-0008/#blank-lines:

> Surround top-level function and class definitions with two blank lines.
> 
> Method definitions inside a class are surrounded by a single blank line.
> 
> Extra blank lines may be used (sparingly) to separate groups of related functions. Blank lines may be omitted between a bunch of related one-liners (e.g. a set of dummy implementations).
> 
> Use blank lines in functions, sparingly, to indicate logical sections.

PR #12, opened 2020-03-11
Bugfix

svenruf:
Updated the CRS extraction in io.postgis.write_triplegs_postgis 
to GeoPandas 0.7


henrymartin1:
I think these changes are valid and they are maybe also related to #15?

Thanks for the fix @svenruf

henrymartin1:
Before we can merge this, we will have to adjust the tests so that they will pass again

henrymartin1:
At some point we need to set up a postgis testing environment... then it would be easier to find and include these error...

PR #13, opened 2020-09-16
Similarity Module (trajectroy distances)

svenruf:
- Similarity Module to calculate tajectory distances (DTW, EDR, ...)
- Storing the ids of extracted triplegs also in the original positionfixes -> pfs can be used for further calculations


trackintel/similarity/detection.py
```diff
@@ -0,0 +1,115 @@
+# -*- coding: utf-8 -*-
+from trackintel.similarity import measures
+import numpy as np
+from scipy.sparse import dok_matrix
+from console_progressbar import ProgressBar
+
+
+def e_dist_tuples(c1,c2):
```
henrymartin1, line 8: should be replaced with a real euclidean distance function

trackintel/similarity/detection.py
```diff
@@ -0,0 +1,115 @@
...
+    return np.sqrt((c1[0]-c2[0])**2 + (c1[1]-c2[1])**2)
+
+def similarity_matrix(data, method, field='tripleg_id', trsh=None, eps=None, dist=False, **kwargs):
```
henrymartin1, line 11: I think there are some critical adjustments that have to be made before we can merge this.
- I don't think that tripleg similarity should be based on positionfixes with an ID column but only on the tripleg data (and then on the query points of the tripleg). This will allow us to cover use cases where we have triplegs but no positionfixes (e.g., GreenClass). 
- I don't think that the matrix assignment should be ID based. At the moment we are calculating a full distance matrix (meaning a distance between all pairs) but store it in a sparse matrix. I can see the need to maintain the relation between IDs and rows but it should be done by either returning an ID vector or by telling the user in the documentation that the order in the distance matrix is the same as the order in the dataframes that were given to the function. 
- The function should either call or be merged with the trackintel.geogr.distances.calculate_distance_matrix function.

trackintel/similarity/measures.py
```diff
@@ -0,0 +1,215 @@
...
+    
+    try:
+        traj_dist[int(id_to_compare)]=0 #depending on the method chosen by the user
+    except UnboundLocalError:
+        pass
+    return traj_dist
+
+
+
+def ses_extract(tpl):
```
henrymartin1, line 189: doc missing

trackintel/similarity/measures.py
```diff
@@ -0,0 +1,215 @@
+"""
+Algorithms e_dtw and e_edr adapted from https://github.com/bguillouet/traj-dist
```
henrymartin1, line 2: I like the idea of having a collection of trajectory distance metrics as it is done in measures.py. In calculate_distance_matrix we tried to be as compatible as possible to [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html). I think here it would also be a good idea to try to implement valid metrics so that we can _piggyback _ on the pairwise_distances implementation of scikit (maybe it is even possible to use their parallelization).
I think this is mostly already done here.

henrymartin1 (commented):
Good work but we need to add some changes so that we can integrate it.

henrymartin1:
Update for merge plan:
- We use a single calculate_distance_matrix function (that we might rename at some point...) that is a wrapper around the scikit [pairwise function ](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)with some extra options (such as vectorized haversine). 
- We will have a seperate file for trajectory-trajectory and point-point metrics
- calculate_distance_matrix needs to distinguish the following cases: 
    - XX distance and XY distance
    - check if metric is suitable for points or if metric is suitable for trajectories
- The default will be to calculate the trajectory distances on the query points of the trajectory however in the future it should be possible to also do it based on positionfixes to be able to use the timestamp information.

trackintel/similarity/detection.py
henrymartin1: edit: The function actually supports sparse calculation

requirements.txt
```diff
@@ -9,4 +9,6 @@ geoalchemy2
 osmnx 
 psycopg2
 imageio
-simplification
\ No newline at end of file
+simplification
+console_progressbar
```
hong2223, line 13: Why do we need the `console_progressbar` module? I did not find its usage in the other files. Afaik, this is used for info printing, can we replace it with the more popular `tqdm` module?

henrymartin1: For now I just removed it but we can think about including tqdm (in a different pull request... this one already got quite confusing)

henrymartin1:
@hong2223 : I'd merge it now if you have no other comments

henrymartin1 (commented):
.

PR #28, opened 2020-10-06
Trip generation inspection

henrymartin1:
A review of the trip generation method that lead to the exclusion of the propagate triplegs method that was created a while ago.


henrymartin1 (commented):
ok

PR #42, opened 2020-10-10
Trip generation

henrymartin1:
Feat: Added first version of trip generation. 

I have locally merged this already with my other pull request (https://github.com/mie-lab/trackintel/pull/28)[https://github.com/mie-lab/trackintel/pull/28]. This is probably not a good idea ;-) If this is too messy, you can reject it at first and we'll figure out what happens once the other pull request is accepted. 
Otherwise I think this pull request only alters three files:
- trackintel/preprocessing/triplegs.py
- tests/preprocessing/test_triplegs.py
- tests/data/geolife_long/trips.csv


henrymartin1:
Sorry for the mess... I had to merge in the previous pull request locally...

PR #46, opened 2020-10-15
Add test_distances.py

schalaur:
test for the haversine-distance


henrymartin1:
Great first test! Thanks for the contribution.

PR #53, opened 2020-11-15
Add github action to check documentation syntax

rbuffat:
This PR adds a github action that runs rstcheck on each PR or push to the repository. 

rstcheck is a tool to ["Checks syntax of reStructuredText and code blocks nested within it."](https://pypi.org/project/rstcheck/)

### DIFF
--- a/.github/workflows/rstcheck.yml
+++ b/.github/workflows/rstcheck.yml
@@ -1,0 +1,26 @@
+name: rstcheck
+
+# Run this workflow every time a new commit pushed to your repository
+on: [push, pull_request]
+
+jobs:
+  rstcheck:
+    name: Doc Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


henrymartin1:
Sounds cool, thanks for the PR! So this is basically a quality check for the documentation?
Do we need to add something to the repository?

rbuffat:
> So this is basically a quality check for the documentation?

Yes, basically a CI for the documentation. However, it could be that rstcheck does not support things that are supported by e.g. read the docs. E.g. the autofunction are disabled in the rstcheck command: `--ignore-directives autofunction`

So in the future, there might other things that need to be considered. If the action is more a burden than a help, it can be stopped by simply deleting the yml file. 

> Do we need to add something to the repository?

No, github actions should be enabled by default, and all configurations are contained within the yml file. Once this PR is merged, the action will run for every push or PR.

henrymartin1:
Is there a button for this?

rbuffat:
No, it will show up next to the travis check.

PR #54, opened 2020-11-15
reimplement preprocessing.cluster_staypoints function

hong2223:
- Update the naming of 'places' to 'locations'
- use the `dissolve()` function from `geopandas `to speed up the location data frame creation
- function returns staypoints together with locations 
- update the test functions accordingly


trackintel/preprocessing/staypoints.py
```diff
@@ -78,34 +78,42 @@ def cluster_staypoints(staypoints, method='dbscan',
...
+        ret_loc.loc[~point_idx, 'center'] = ret_loc.loc[~point_idx, 'geom'].apply(lambda p: Point(np.array(p)[:,0].mean(), 
+                                                                                                np.array(p)[:,1].mean()))
+        
+        # extent is the convex hull of the geometry
+        ret_loc['extent'] = ret_loc['geom'].apply(lambda p: p.convex_hull)                                                                                        
+        # convex_hull of one point would be a Point and two points a Linestring, 
+        # we change them into Polygon by creating a buffer of epsilon around them.
+        pointLine_idx = (ret_loc['extent'].geom_type == 'LineString') | (ret_loc['extent'].geom_type == 'Point')
+        
+        # TODO: how to know if the crs is geodetic or projection?
```
henrymartin1, line 104: I would say that we don't do any projections here. We should expect the user to know the data and then the epsilon that we use for the buffer should be the same unit as we use it in DBScan.
I think at the moment there is still a mistake in the code if I am not getting this wrong. 
Example: 
At the moment it could be the case that if the data is in WGS84 and the user _wants_ to cluster the data in degree (for what ever reason...), he would pass _epsilon=0.01_ to the method. The places then get clustered using epsilon=0.01 degree.
For the extent of places with only one point you read the 0.01 degrees as meters and use the meters_to_decimal_degrees function to transform it into something very small...
You could use this approach if the user adds the haversine keyword for the distance matrix metric as we can then expect that the distance is in meters.

henrymartin1 (changes requested):
Address my comments in trackintel/preprocessing/staypoints.py line 105

henrymartin1:
Looks good!

PR #60, opened 2020-11-18
Update location extraction method

hong2223:
- change the naming of 'places' to 'locations' throughout the repo
- add dataset level location extraction
- add tests to the location extraction method


AUTHORS.md
```diff
@@ -6,6 +6,8 @@ Henry Martin: [@henrymartin1 (Github)](https://github.com/henrymartin1).
 
 René Buffat: [@rbuffat (Github)](https://github.com/rbuffat).
 
+Ye Hong: [@hong2223 (Github)](https://github.com/hong2223).
```
henrymartin1, line 9: I think we also need to add you as an author in the documentation somewhere...

henrymartin1 (approved):
Great work, thanks for the refactoring and the additional dataset_level option!

PR #65, opened 2020-11-20
feat(analysis): add simple coarse mode predictions

dominikbucher:
To be fair, the predictions here are _really_ simple (and coarse). Maybe this can just serve as a basis for discussion, especially considering the function signatures. We can always improve the simple/coarse prediction model later on.


trackintel/analysis/__init__.py
```diff
@@ -0,0 +1 @@
+from .transport_mode_identification import predict_transport_mode
```
henrymartin1, line 1: is this consistent everywhere in trackintel. I have never included statements like this... maybe you'll give me a quick start on init statements at some point...

henrymartin1 (commented):
Looks good. I would agree that it is quite simple but having the signature makes it much easier to extend the framework. 
You could check if you can change the function length calculation.

trackintel/analysis/__init__.py
dominikbucher: I think it mostly depends on what you want to be exposed when importing the module. Like this, you can say `trackintel.analysis.predict_transport_mode`. I think the main functions within a module are often exposed like this. 

In addition, it prevents errors that a module (or rather a function defined in another file in the module folder) is not imported yet (I think). If you remove the line and let the tests run the `test_transport_mode_identification` test fails as the module is not imported anywhere. 

For the other tests it seems to work as they are (luckily (?)) imported along the way somewhere (most in `preprocessing` which in turn is imported in the `model` files, and those are imported in the `trackintel/__init__.py`).

dominikbucher:
> Looks good. I would agree that it is quite simple but having the signature makes it much easier to extend the framework.
> You could check if you can change the function length calculation.

What do you mean by function length calculation? The distance and resulting speed? Is there an (implemented) way we can compute it in meters (globally)?

trackintel/analysis/transport_mode_identification.py
```diff
@@ -0,0 +1,55 @@
...
+
+    For additional documentation, see 
+    :func:`trackintel.analysis.transport_mode_identification.predict_transport_mode`.
+    """
+    triplegs = triplegs.copy()
+
+    def identify_mode(tripleg):
+        """Identifies the mode based on the (overall) tripleg speed."""
+        # Computes distance over whole tripleg geometry (using the Haversine distance).
+        distance = sum([haversine_dist(pt1[0], pt1[1], pt2[0], pt2[1]) for pt1, pt2 \
```
henrymartin1, line 42: Is this the same as: https://geopandas.org/reference.html#geopandas.GeoSeries.length ?

henrymartin1:
> 
> 
> > Looks good. I would agree that it is quite simple but having the signature makes it much easier to extend the framework.
> > You could check if you can change the function length calculation.
> 
> What do you mean by function length calculation? The distance and resulting speed? Is there an (implemented) way we can compute it in meters (globally)?

Ah sorry... I somehow didn't submit the related comment in the code... I was thinking if we could use the geopandas.geometry.length function directly. 

https://geopandas.org/reference.html#geopandas.GeoSeries.length

dominikbucher:
For the length, we agreed to use the current implementation if the coordinate system is WGS84 (as a convenient default), and use the `GeoSeries.length` function otherwise, which _should_ generate a warning if the coordinate system is _not_ a projected one. In this case, it might actually make sense to capture the warning and improve the message (the speed will _never_ be correct, even if a user is aware that he is using a geographic (and not projected) coordinate system).

henrymartin1:
> 
> 
> For the length, we agreed to use the current implementation if the coordinate system is WGS84 (as a convenient default; we will provide a helper function `haversine_dist_linestring` though), and use the `GeoSeries.length` function otherwise, which _should_ generate a warning if the coordinate system is _not_ a projected one. In this case, it might actually make sense to capture the warning and improve the message (the speed will _never_ be correct, even if a user is aware that he is using a geographic (and not projected) coordinate system).

The current plan of @svenruf is: 
- Create a function 'haversine_dist_linestring'
- make a case distinction
   - If WGS84 --> use 'haversine_dist_linestring'
   - Else --> use gpd.length()
- Adjust tests accordingly

svenruf:
I merged this into my development fork and added some cases, see PR #103. Once accepted we could close this PR. Thanks @dominikbucher for the good base :)

henrymartin1:
This PR will be replaced by #103

PR #66, opened 2020-11-20
feat(filter): add spatial filter function

hong2223:
- add `spatial filter` function under `filter` module.
- register the function under the `staypoints`, `locations`, and `triplegs` accessors.
- add tests: the results are compared with ArcGIS manually obtained results


henrymartin1 (approved):
Great contribution! I really like the spatial filtering.

PR #68, opened 2020-11-22
douglas peucker function and test

zero31415:
- replaced the simplify function in simplification.cutil to the one in shapely

- modified the test function


henrymartin1 (approved):
Looks great, thanks for the contribution!

PR #72, opened 2020-12-02
Experimental/efficient trip generation

hong2223:
- a more efficient implementation of the trip generation function using `groupby().apply()`.
- add 'tqdm' as a required library for progress printing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ osmnx
 psycopg2
 imageio
 simplification
+tqdm
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,5 +9,4 @@ geoalchemy2
 osmnx 
 psycopg2
 imageio
-simplification
 tqdm
--- a/trackintel/preprocessing/triplegs.py
+++ b/trackintel/preprocessing/triplegs.py
@@ -3,7 +3,7 @@ import datetime
 
 import numpy as np
 import pandas as pd
-
+from tqdm import tqdm
 
 def smoothen_triplegs(triplegs, method='douglas-peucker', tolerance=1.0):
     """reduces number of points while retaining structure of tripleg
@@ -47,7 +47,7 @@ def _temp_trip_stack_has_tripleg(temp_trip_stack):
     return has_tripleg
 
 
-def _create_trip_from_stack(temp_trip_stack, origin_activity, destination_activity, trip_id_counter):
+def _create_trip_from_stack(temp_trip_stack, origin_activity, destination_activity):
     """
     Aggregate information of trip elements in a structured dictionary
 
@@ -60,8 +60,6 @@ def _create_trip_from_stack(temp_trip_stack, origin_activity, destination_activi
                     Either dictionary or pandas series
     destination_activity : dictionary like
                     Either dictionary or pandas series
-    trip_id_counter : int
-            current trip id
 
     Returns
     -------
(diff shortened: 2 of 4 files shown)


trackintel/preprocessing/triplegs.py
```diff
@@ -200,124 +155,167 @@ def generate_trips(stps_input, tpls_input, gap_threshold=15, id_offset=0, print_
...
+    
+    # assign trip_id to tpls
+    ids = tpls.index
+    trip2tpl_map = trips[['id', 'tpls']].set_index('id').to_dict()['tpls']
+    ls = []
+    for key, values in trip2tpl_map.items():
+        for value in values:
+            ls.append([value, key])
+    temp = pd.DataFrame(ls, columns=['tplsIndex', 'trip_id'])
+    tpls = tpls.merge(temp, how='left', left_on='id', right_on='tplsIndex')
```
henrymartin1, line 178: left_of should be tpls.index.name to allow for flexible index names. 
If we are using the index instead of column names we could switch to pd.join() instead of pd.merge()

trackintel/preprocessing/triplegs.py
```diff
@@ -200,124 +155,167 @@ def generate_trips(stps_input, tpls_input, gap_threshold=15, id_offset=0, print_
...
+    tpls.drop(columns='tplsIndex', inplace=True)
+    tpls['id'] = ids
+    tpls.set_index('id', inplace=True)
```
henrymartin1, line 181: I think the index of `tpls` is left untouched by the transformations above. If you can confirm, than we should delete these lines in order to not to change the index name of the user (I think at the moment we do not require that the index has the name 'id' but I am not sure...).

trackintel/preprocessing/triplegs.py
```diff
@@ -200,124 +155,167 @@ def generate_trips(stps_input, tpls_input, gap_threshold=15, id_offset=0, print_
...
-            if in_trip is False:
-                # If there are several activities in a row, we skip until the last one
-                if row['activity'] and row['activity_next']:
-                    continue
-
-                # if this is the last activity before the trip starts, reset the origin
-                elif row['activity']:
+def _generate_trips_user(df, gap_threshold):
+
+    unknown_activity = {'user_id': df['user_id'].unique()[0], 'activity': True, 'id': np.nan}
```
henrymartin1, line 228: I think the `df['user_id'].unique()[0]` is a bit confusing. I am ware that we enter this function after grouping by user id, so couldn't we simply tyke `df['user_id'].iloc[0]` ?
On the other way, we could use the `df['user_id'].unique()[0]` for an assert that could help the readability. 
```python
user_id = df['user_id'].unique()
assert len(user_id) == 1
user_id = user_id[0]
unknown_activity = {'user_id': user_id, 'activity': True, 'id': np.nan}
```

henrymartin1 (changes requested):
Hey @hong2223, 
great work! Really nice extension of the trip generation. I had some questions/comments regarding the update of the id's after the trip generation that you should have a look at before we can merge it.

requirements.txt
```diff
@@ -9,4 +9,5 @@ geoalchemy2
 osmnx 
 psycopg2
 imageio
-simplification
\ No newline at end of file
+simplification
```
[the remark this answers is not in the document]
hong2223: the dependency is deleted

trackintel/preprocessing/triplegs.py
hong2223: Replaced and solved

requirements.txt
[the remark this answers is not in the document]
henrymartin1: I am pretty sure that I also already deleted it. I was wondering if it could be that it sneaked back in because of a pull request that was not up to date with the origin.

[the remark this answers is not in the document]
hong2223: Yes, I also remember you deleted it. 
I was careful in sync every time I create a new branch, but maybe at some point I overlooked

trackintel/preprocessing/triplegs.py
hong2223: solved, now index names can be arbitrary.

trackintel/preprocessing/triplegs.py
hong2223: solved. now using 'join' the indexes of tpls and spts are left untouched

hong2223:
the function now accepts flexible index names of 'stps' and 'tpls'; the generated 'trips' has 'id' as the index name.

henrymartin1 (approved):
Nice!

Issue #70, opened 2020-12-02
Improve dtype consistency for trackintel objects


henrymartin1:
I think it would be a good Idea if we somehow improve how we handle data types in our functions. I am raising this issue because I noticed that when `dtype(pfs.user_id)` is integer, then:
`pfs.as_positionfixes.extract_staypoints()` returns float user ids while `pfs.as_positionfixes.extract_triplegs(spts)` creates int user ids.

I think the best practice would be that input dtype = output dtype especially because user ids could be strings too.

Issue #56, opened 2020-11-16
Check if id preservation + id propagation works for all datatypes


henrymartin1:
We should check (and write a test) if ids are properly stored in all datatypes. 

E.g., 
positionfixes -> triplegs (tripleg_id should exist in positionfixes)
positionfixes -> staypoints (staypoint_id should exist in positionfixes)
triplegs/staypoints -> trips (trip_id, prev_trip_id, next_trip_id in staypoints/triplegs)
staypoins->locations (staypoints have location id)
trips -> tours (trips should have tour id)

henrymartin1:
At the moment I can confirm the following problems:
- `as_positionfixes.extract_triplegs` does return a dataframe with non unique index (but a unique column 'id'). 
- `as_positionfixes.extract_staypoints` does still return the column `id` (but with unique index)

hong2223:
> positionfixes -> triplegs (tripleg_id should exist in positionfixes) 
In Case 1 of the `generate_triplegs` function for positionfixes, no 'tripleg_id' field is assigned to the new positionfixes

henrymartin1:
Originally this issue was concerning the problem with altering the input data and with not having IDs as an index. It now only concerns the problem of changing the input dataframe. The problem of having IDs as pandas index is described in Issue #97

PR #88, opened 2021-01-26
Fix: dtype consistency and function outputs

hong2223:
For generate functions:
- functions do not change the original input df but instead returns the modified one (with the corresponding generated ids).
- The generated id is in int64. Exception: the dtype for [`trip_id` `prev_trip_id` and `next_trip_id`] for spts and [`trip_id`] for triplegs in 'generate_trips' function is float, as int64 does not support missing values ('NaN').
- generated df has column 'id', while original df receives a new column '{generated}_id'. Example: loc_df has ['id'] field, while original stps_df has ['location_id'] field. (Partly solves Issue #56 )
- 'user_id' dtype of the generated df is consistent with the original one. (Issue #70  )
- modified tests to account for these changes


henrymartin1 (approved):
Another step towards consistency in the framework! 
For IDs we originally agreed on always setting them as the index of the dataframe but I think it is easier to merge this request and treat the ID problem in issue #97.

Thanks for the input!

Issue #23, opened 2020-09-25
Add rules for test structure


henrymartin1:
I think it would be beneficial to have a common organization for the layout of test functions. We can add this in the development section in the readme. I thought [this stackoverflow discussion ](https://stackoverflow.com/a/57532692) on this topic is quite interesting. Personally I think that the "Put all tests for each function in a class" approach is very convincing

henrymartin1:
I summarize: 
- The test folder copies the folder structure that the trackintel.trackintel folder has. 
- Every python module has a single test file
- Every function has 1 test class
- Every method of this function should test a single property

henrymartin1:
Rules are added to [contributing.md](https://pypi.org/project/trackintel/)

PR #93, opened 2021-02-02
Feat: import GeoDataFrames

svenruf:
added trackintel/io/from_geopandas.py: Created functions that warp pd.rename to import GeoDataFrames into the trackintel model
added tests for these functions in tests/io/test_io.py and datasets for these tests in tests/data (geojson files)

the commits regarding the test_pypi file can be ignored as they were deleted. Unfortunately I could not get rid of these commits (also the hard reset did not remove them...) 

the issue with the locations (multiple geometry) is related to the issues [867](https://github.com/geopandas/geopandas/issues/867) / [1275](https://github.com/geopandas/geopandas/issues/1275) in Geopandas


tests/io/test_io.py
```diff
@@ -79,3 +123,14 @@ def test_trips_from_to_postgis(self):
...
 
+    def test_trips_from_gpd(self): 
+        csv_file = os.path.join('tests','data','trips.csv')
+        df = pd.read_csv(csv_file, sep=';')
+        trps = ti.io.from_geopandas.trips_from_gpd(df)
+        trps_csv = ti.read_trips_csv(csv_file,sep=';')
+        trps_csv['started_at'] = trps_csv['started_at'].apply(lambda d: d.isoformat().replace(' ', 'T'))
+        trps_csv['finished_at'] = trps_csv['finished_at'].apply(lambda d: d.isoformat().replace(' ', 'T'))
+        trps_csv['started_at'] = trps_csv['started_at'].apply(lambda d: d.replace('+00:00', 'Z'))
+        trps_csv['finished_at'] = trps_csv['finished_at'].apply(lambda d: d.replace('+00:00', 'Z'))
```
hong2223, line 134: same as above, here it is a bit more confusing..

trackintel/io/from_geopandas.py
```diff
@@ -0,0 +1,203 @@
...
+    
+    lcs = gdf.rename(columns=columns)
+    lcs = lcs.set_geometry('center')
+    
+    assert lcs.as_locations
+    
+
+    return lcs
+
+def tours_from_gpd(gdf, user_id='user_id',started_at='started_at', finished_at='finished_at', origin_destination_location_id='origin_destination_location_id', journey='journey', mapper={}):
```
hong2223, line 170: for this function, there are no tests available? Maybe add an issue saying that test for tours are missing in both read_csv and from_geopandas

hong2223 (commented):
Nice job! Before we can merge, maybe we need to add some more descriptions to the functions (see the specific comments).
We are also trying to organize the file structures in the 'test' folder. The general idea is to imitate the file structure of the corresponding modules and to create a class for each function. You can see the file 'test_positionfixes.py' and issue #23 as examples. 
So here, maybe create a separate test file named `test_from_geopandas.py`.

svenruf:
Updated the PR with the points discussed in our meeting:
- comments on the hardcoded datetime modifications
- cleaned up the structure of the tests
Anything to add?

hong2223:
Nice job!

PR #100, opened 2021-02-07
Feat: flexible csv column names

svenruf:
Allows the user to rename csv columns directly while using the e.g. read_triplegs_csv functions
modified tests accordingly
added test data with modified column names


trackintel/io/file.py
```diff
@@ -10,6 +10,9 @@ def read_positionfixes_csv(*args, **kwargs):
     builds a geopandas GeoDataFrame. This also validates that the ingested data
     conforms to the trackintel understanding of positionfixes (see 
     :doc:`/modules/model`). 
+    
+    In addition to the pandas read_csv keyword arguments, this function provides to rename columns directly while importing
+    by specifying the columnnames in a dictionary, see second example.
```
hong2223, line 15: I am wondering if we need to explicitly write out the `column` as a parameter and describe how to use it (mainly just original col as key and standard col names as value), an example of comment see [here ](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/positionfixes.py).

trackintel/io/file.py
```diff
@@ -58,12 +64,22 @@ def read_triplegs_csv(*args, **kwargs):
...
+    by specifying the columnnames in a dictionary, see example.
+    
     Returns
     -------
     GeoDataFrame
         A GeoDataFrame containing the triplegs.
+        
+    Examples
+    --------
+    >>> trackintel.read_triplegs_csv('data.csv', columns={'start_time':'started_at'})
```
hong2223, line 78: maybe add an example without columns here.

tests/io/test_file.py
```diff
@@ -61,13 +70,17 @@ def test_staypoints_from_to_postgis(self):
...
+        mod_file = os.path.join('tests','data','locations_mod_columns.csv')
         tmp_file = os.path.join('tests', 'data', 'locations_test.csv')
-        locs = ti.read_locations_csv(orig_file, sep=';')
-        locs.as_locations.to_csv(tmp_file, sep=';',
-                                 columns=['user_id', 'elevation', 'center', 'extent'])
+        mod_plcs = ti.read_locations_csv(mod_file, columns={'geom':'center'},sep=';')
+        plcs = ti.read_locations_csv(orig_file, sep=';')
+        assert mod_plcs.equals(plcs)
+        plcs.as_locations.to_csv(tmp_file, sep=';',
+                              columns=['user_id', 'elevation', 'center', 'extent'])
```
hong2223, line 79: change the `plcs` (places) to `locs` (locations). We have changed the wording "places" into "locations".

trackintel/io/file.py
```diff
@@ -58,12 +64,22 @@ def read_triplegs_csv(*args, **kwargs):
     builds a geopandas GeoDataFrame. This also validates that the ingested data
     conforms to the trackintel understanding of triplegs (see :doc:`/modules/model`).
 
+
+    In addition to the pandas read_csv keyword arguments, this function provides to rename columns directly while importing
+    by specifying the columnnames in a dictionary, see example.
+
```
hong2223, line 70: Comment explicitly, see above

trackintel/io/file.py
```diff
@@ -94,13 +111,18 @@ def read_staypoints_csv(*args, **kwargs):
...
     the ingested data conforms to the trackintel understanding of staypoints 
     (see :doc:`/modules/model`).
+    
+    In addition to the pandas read_csv keyword arguments, this function provides to rename columns directly while importing
+    by specifying the columnnames in a dictionary, see examples in read_triplegs_csv or read_positionfixes_csv.
 
     Returns
     -------
     GeoDataFrame
         A GeoDataFrame containing the staypoints.
```
hong2223, line 121: Maybe also add the two examples here

trackintel/io/file.py
```diff
@@ -132,13 +154,17 @@ def read_locations_csv(*args, **kwargs):
     validates that the ingested data conforms to the trackintel understanding 
     of locations (see :doc:`/modules/model`).
 
+    In addition to the pandas read_csv keyword arguments, this function provides to rename columns directly while importing
+    by specifying the columnnames in a dictionary, see examples in read_triplegs_csv or read_positionfixes_csv.
+
```
hong2223, line 159: Comment explicitly, see above

trackintel/io/file.py
```diff
@@ -132,13 +154,17 @@ def read_locations_csv(*args, **kwargs):
...
     Returns
     -------
     GeoDataFrame
         A GeoDataFrame containing the locations.
     """
```
hong2223, line 164: Maybe also add the two examples here

trackintel/io/file.py
```diff
@@ -170,19 +196,25 @@ def read_trips_csv(*args, **kwargs):
     """Wraps the pandas read_csv function and extraces proper datetimes. This also 
     validates that the ingested data conforms to the trackintel understanding 
     of trips (see :doc:`/modules/model`).
+    
+    In addition to the pandas read_csv keyword arguments, this function provides to rename columns directly while importing
+    by specifying the columnnames in a dictionary, see examples in read_triplegs_csv or read_positionfixes_csv.
```
hong2223, line 202: Comment explicitly, see above

trackintel/io/file.py
```diff
@@ -170,19 +196,25 @@ def read_trips_csv(*args, **kwargs):
...
 
     Returns
     -------
     DataFrame
         A DataFrame containing the trips.
     """
```
hong2223, line 207: Maybe also add the two examples here

hong2223 (changes requested):
Nice functionality! I am mostly concerned with these issues:
- Comment explicitly about the usage of `column`.
- Complete the example line of code in the comments.
- Change variable name `plcs` to `locs`.

Some of them are legacy issues; it will be nice if you can solve them

trackintel/io/file.py
svenruf: Yes I thought about it too, but was not sure if this is "clean" to explicitly describe keyword arguments. But as this is the only one that is used by our function, it makes sense to add this.

svenruf:
included the reviews in the last commit

Issue #18, opened 2020-09-24
Timestamp format should be defined in the trackintel data model


henrymartin1:
The timestamp of the columns started_at and finished_at should be in a specific format (e.g., pandas timestamps) so that later algorithms can use these columns in a safe way

henrymartin1:
@svenruf I agree that this should be done rather sooner than later.

hong2223:
Currently, the `TestFromGeopandas` would fail in my local environment but pass the CI (and also pass in Sven's environment). Guess the default timestep format is different. 
This would be a very important issue to fix.

henrymartin1:
I agree that this is very important and I think it is not too hard to implement. 
Basically we would agree that all timestamps in trackintel objects have pandas [datetime64](https://pbpython.com/pandas_dtypes.html) as dtype. Then we would check it in the accessors and change it in the model documentation.
If you agree to that I could try implementing it.

hong2223:
> I agree that this is very important and I think it is not too hard to implement.
> Basically we would agree that all timestamps in trackintel objects have pandas [datetime64](https://pbpython.com/pandas_dtypes.html) as dtype. Then we would check it in the accessors and change it in the model documentation.
> If you agree to that I could try implementing it.

Sure. Go ahead!

henrymartin1:
I think this is a good point to introduce timezone awareness. I found this [blog entry](https://howchoo.com/g/ywi5m2vkodk/working-with-datetime-objects-and-timezones-in-python#enter-timezones) which had some very usefull advise on timezones: 
- always use "aware" datetime objects.
- always work in UTC and do timezone conversion as a last step.

So the goal is to have timezone aware timestamps and then we can advise the user to have them in UTC. This will require all io functions to have an optional timezone input which defaults to UTC. If the user does not set the timezone I would throw a warning that we assume the timezone to be UTC.

henrymartin1:
solved with #101

PR #101, opened 2021-02-09
Fix datetime type

henrymartin1:
Trackintel is now timezone aware
- all required datetime columns are now forced to have a timezone and a fixed datatype using an assert statement in the accessors
- all io functions (except for postgis!) now import timezone aware using an optional timezone argument. If nothing is provided they warn and assume UTC
- Tests are updated according to these changes
- Reformated files according to PEP8 + numpy docstring standards

This solves issue #18


hong2223:
Great work!
I quickly fixed the typos and this is now ready to merge

PR #102, opened 2021-02-10
Update Readme_developers.md

henrymartin1:
@hong2223 Maybe you have an idea if something is missing here


henrymartin1:
Add line length of 120?

hong2223:
I think the file now contains the conventions we agreed on. We could always add points afterward.

PR #103, opened 2021-02-10
Feat: simple coarse mode prediction

svenruf:
adapted from #65 
added cases for different CRS:
- WGS84 -> haversine distance (with UserWarning)
- None -> WGS84 is assumed (with UserWarning)
- geographic coordinate system (other than WGS84) -> raises UserWarning, no calculation done
- other (projected) system -> geoSeries.length is used

modified test accordingly

Is the setting of the warnings etc. ok like this?


trackintel/analysis/transport_mode_identification.py
```diff
@@ -0,0 +1,62 @@
...
+
+    For additional documentation, see 
+    :func:`trackintel.analysis.transport_mode_identification.predict_transport_mode`.
+    """
+    triplegs = triplegs.copy()
+    
+    wgs = ti.geogr.distances.check_crs(triplegs)
+
+
+    def identify_mode(tripleg,wgs):
```
henrymartin1, line 43: This looks like an indentation error or is this because of the online viewer?

trackintel/analysis/transport_mode_identification.py
```diff
@@ -0,0 +1,62 @@
...
+        # Computes distance over whole tripleg geometry (using the Haversine distance).
+        if wgs:
+            distance = sum([haversine_dist(pt1[0], pt1[1], pt2[0], pt2[1]) for pt1, pt2 \
+                in zip(tripleg.geom.coords[:-1], tripleg.geom.coords[1:])])
+        else:
+            distance = tripleg.geom.length
+        duration = (tripleg['finished_at'] - tripleg['started_at']).total_seconds()
+        speed = distance / duration
+
+        if speed < 15 / 3.6:
```
henrymartin1, line 54: I think it would be nice to have these values as arguments in the function with defined default values.
We also have to add a comment about the unit of the speed. I guess m/s?

tests/analysis/test_transport_mode_identification.py
```diff
@@ -0,0 +1,53 @@
...
+        """Assert that the method does not work for empty DataFrames (but
+        that the rest works fine, e.g., method signature)."""
+        tpls = ti.read_triplegs_csv('tests/data/triplegs_transport_mode_identification.csv', sep=';')
+        empty_frame = tpls[0:0]
+        with pytest.raises(AssertionError):
+            empty_frame.as_triplegs.predict_transport_mode(method='simple-coarse')
+   
+
+    def test_check_correct_simple_coarse_identification(self):
+        """Assert that the simple-coarse transport mode identification yields the correct results."""
```
henrymartin1, line 26: I think it would be better to split this into different tests

trackintel/analysis/transport_mode_identification.py
svenruf: I think Dominik implemented the function inside the other function. Idk if this is correct, but it seems to be working. Indent error would have caused the tests to fail, wouldn't it?

trackintel/analysis/transport_mode_identification.py
svenruf: Ok, yes this makes sense. Should we also offer to define more categories (e.g. by passing a dict to the function?)

tests/analysis/test_transport_mode_identification.py
svenruf: Ok so one test per crs case?

henrymartin1: One per crs case could be a good idea. 
A bit hard to say how to split the function as it is a bit flexible. The idea is to have 1 function per functionality that you want to test. E.g., you could have one case to test correct warning behavior, one to test correct classification behavior (for different crs cases) or one per crs. A good rule of thumb is that you should be able to explain the test in a single short sentence (this should then be used to document the test). 
E.g., "Testing correct warning behavior" or "Testing correct behavior for crs=4326"

trackintel/analysis/transport_mode_identification.py
henrymartin1: Yes, I think [nested functions](https://realpython.com/inner-functions-what-are-they-good-for/) are OK. They are not mentioned in pep8 so why not

trackintel/analysis/transport_mode_identification.py
henrymartin1: Yes, if this is not too much effort. Just make sure that the units are documented and the validity of the input dictionary is tested.

trackintel/geogr/distances.py
```diff
@@ -161,4 +162,17 @@ def meters_to_decimal_degrees(meters, latitude):
...
-    return meters / (111.32 * 1000.0 * cos(latitude * (pi / 180.0)))
\ No newline at end of file
+    return meters / (111.32 * 1000.0 * cos(latitude * (pi / 180.0)))
+
+
+def check_crs(data):
+    wgs = False
+    if data.crs == 4326:
+        wgs=True
+        warnings.warn('Your data is in WGS84, for length calculation the haversine distance is used')
```
henrymartin1, line 172: Hm... if the data is in wgs84 than using haversine seems like the expected behavior. In this case we should not throw a warning.

trackintel/geogr/distances.py
```diff
@@ -161,4 +162,17 @@ def meters_to_decimal_degrees(meters, latitude):
     float
         An approximation of a distance (given in meters) in degrees.
     """
-    return meters / (111.32 * 1000.0 * cos(latitude * (pi / 180.0)))
\ No newline at end of file
+    return meters / (111.32 * 1000.0 * cos(latitude * (pi / 180.0)))
+
+
+def check_crs(data):
```
henrymartin1, line 168: - Missing docstring
- I think the function name should be more specific and that a "check" function should not throw errors or warnings. At the moment the function is very specific and it is unlikely that we will find another use-case that needs exactly this behavior. So we could either make the function more general (e.g., check_if_wgs84) or move this check directly into the calculate length function.

trackintel/geogr/distances.py
```diff
@@ -161,4 +162,17 @@ def meters_to_decimal_degrees(meters, latitude):
...
+    wgs = False
+    if data.crs == 4326:
+        wgs=True
+        warnings.warn('Your data is in WGS84, for length calculation the haversine distance is used')
+    elif data.crs == None:
+        wgs=True
+        warnings.warn('Your data is not projected. WGS84 is assumed and for length calculation the haversine distance is used')
+    elif data.crs.is_geographic:
+        raise UserWarning('Your data is in a geographic coordinate system, length calculation fails')
```
henrymartin1, line 177: I think a "check" function should not raise exceptions. It should rather return the result of the check and leave the raising of warnings or exceptions to the specific application (the need for exceptions might be different for different applications).

henrymartin1 (changes requested):
I have added some comments. Most are only minor changes but after going through the code I think it might be better to do the checks in "check_crs" directly in the distance function. I have a hard time imagining how this function could be of use for other applications at the moment.

trackintel/geogr/distances.py
svenruf: ok, I will move this functionality back to the length function. I only placed in geogr.distances because I noted this from our last meeting. But I agree, that it makes only sense if it is a general function.

henrymartin1: Yeah... sorry for the back and forth... In the meeting it sounded like a good idea 🤷‍♂️

trackintel/analysis/transport_mode_identification.py
svenruf: See my latest commit. Was not sure where to define the default values, but it should be in the "top-level" function to offer a proper documentation to the user, right? I'm open for feedback :)

tests/analysis/test_transport_mode_identification.py
svenruf: see last commit

henrymartin1 (approved):
Great work! Thanks for the changes. One step closer to 1.0 :-)

PR #109, opened 2021-02-10
Update README.md

henrymartin1:
Here is a first draft. We can maybe do several iterations in the pull request. 
It is not yet fully consistent especially with the activities...


hong2223:
This is by now a very nice and comprehensive readme file. I think we could merge already merge it. What do you think?

Issue #105, opened 2021-02-10
UserWarning because of empty Geometry column


henrymartin1:
Our tests currently throw the following warning: 

> UserWarning: Geometry column does not contain geometry. warnings.warn("Geometry column does not contain geometry.")

I am not sure where exactly this originates but it is thrown for the following tests:

> tests/io/test_file.py::TestFile::test_triplegs_from_to_csv
> tests/io/test_file.py::TestFile::test_staypoints_from_to_csv
> tests/io/test_file.py::TestFile::test_locations_from_to_csv

hong2223:
Nice! Already fixed this issue and will be integrated with the next pull request

henrymartin1:
:D

Issue #104, opened 2021-02-10
DeprecationWarning because of geodataframe initialization


henrymartin1:
Our tests throw the following warning: 

> DeprecationWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.

The reason is the initialization of a Geodataframe in trackintel/preprocessing/staypoints.py in [line 62](https://github.com/mie-lab/trackintel/blob/8067ee064dd23e20b35584f9c4427488a21a26d3/trackintel/preprocessing/staypoints.py#L62) without defining the types of the columns.

PR #110, opened 2021-02-10
Solve test warning

hong2223:
Solves test warnings:
- #104 is caused by using .apply method for empty Series. Add control statement to fix, readability is reduced though.
- #105 is caused by assigning an 'str' column to gpd.geometry column that requires valid geometries. Solved by first convert to pd dataframe.


henrymartin1 (approved):
Nice, as far as I can see there are now 0 warnings left :-)

PR #112, opened 2021-02-12
update version number for new release

hong2223:
update the version number to 0.5.0 for new pypi release


docs/conf.py
```diff
@@ -47,7 +47,6 @@
   'sklearn',
   'sqlalchemy',
   'geoalchemy2',
-  'simplification',
```
henrymartin1, line 50: Ah, good that this is finally done ;-)

henrymartin1 (approved):
step-by-step closer to 1.0

Issue #97, opened 2021-02-03
ID should consistently be used as an index for trackintel datatypes


henrymartin1:
All trackintel objects should have their unique identifier as the geopandas index. This should also be preserved if we process data. The reason we agreed on this was that pandas dataframes can lead to unexpected behavior if the index does not correspond to the ID (e.g., when merging or joining). 

This has to be considered at several places. At the moment I can think of the documentation, the accessors, i/o and functions that generate the outputs (e.g., that generate trips). Once it is implemented as a check in the accessors we will probably find most of the other places were we have to take care of this with the help of failing tests.

In the [documentation](https://trackintel.readthedocs.io/en/latest/modules/model.html ), we have already included the phrase _"the index of the GeoDataFrame will be treated as unique identifier of the Positionfixes"_ for all data types except for `Users`.

Issue #91, opened 2021-01-27
Use groupby.apply method in preprocessing.positionfixes 


hong2223:
At the moment the generate function in `preprocessing.positionfixes` is implemented using for loops for each user. This can be enhanced by using `df.groupby("user_id").apply()` for better performance

PR #115, opened 2021-02-15
PERF: Groupby for gen_loc function

hong2223:
- the `sliding` and `dbscan` methods in `generate_staypoints()` function are now implemented using `.groupby('user_id')` for better performance and readability. The functions `_generate_staypoints_sliding_user` and `_generate_staypoints_dbscan_user` are created accordingly. Link to #91 
- the original function is moved to `test_posifionfixes` and the `test_generate_staypoints_groupby` is created to compare the results using `groupby` and for loop.

Additionally:
- The `id` of `pfs` and `spts` are now set as index. For that the corresponding `from_gpd` and `from_csv` function is changed to assign an `id` field if not provided by the user. If this is ok, I'll create another pull request to change the other read functions. Link to #97 
- The `generate_staypoints` function now accepts a `pfs` with `id` as index, and returns `pfs` and `spts` with `id` as index

Sorry for the mixed pull requests, but it is hard to change functions without changing the `id` field


trackintel/preprocessing/positionfixes.py
```diff
@@ -60,114 +66,54 @@ def generate_staypoints(positionfixes, method='sliding',
...
+        # Assign staypoint_id to ret_pfs if spts is detected 
+        if not ret_spts.empty:
+            stps2pfs_map = ret_spts[['pfs_id']].to_dict()['pfs_id']
+            
+            ls = []
+            for key, values in stps2pfs_map.items():
+                for value in values:
+                    ls.append([value, key])
+            temp = pd.DataFrame(ls, columns=['id', 'staypoint_id']).set_index('id')
+            ret_pfs = ret_pfs.join(temp, how='left')
```
henrymartin1, line 96: This is the beauty if we have strict index requirements :-)

trackintel/preprocessing/positionfixes.py
```diff
@@ -60,114 +66,54 @@ def generate_staypoints(positionfixes, method='sliding',
...
+            ret_spts.drop(columns={'pfs_id'}, inplace=True)
+        # if no staypoint is identified
```
henrymartin1, line 98: This means if no staypoint at all is identified ? Ah, I see the default values create by the join is nan. Good ;-)

trackintel/preprocessing/positionfixes.py
```diff
@@ -179,17 +125,20 @@ def generate_staypoints(positionfixes, method='sliding',
...
+    ret_pfs = gpd.GeoDataFrame(ret_pfs, geometry=name_geocol,crs=ret_pfs.crs)
+    ret_spts = gpd.GeoDataFrame(ret_spts, geometry=name_geocol,crs=ret_pfs.crs)
     
     ## ensure dtype consistency 
     # stps id (generated by this function) should be int64
-    ret_spts['id'] = ret_spts['id'].astype('int64')
-    ret_pfs['staypoint_id'] = ret_pfs['staypoint_id'].astype('int64')
+    ret_spts.index = ret_spts.index.astype('int64')
+    # ret_pfs['staypoint_id'] should be float
+    ret_pfs['staypoint_id'] = ret_pfs['staypoint_id'].astype('float')
```
henrymartin1, line 141: Hm... it is a bit weird to have these different types for the different indices.... Maybe we can either set everything to float or set everything to int and use "None" for missing values...

henrymartin1: Anyways this is no problem that seems important at the moment...

trackintel/preprocessing/positionfixes.py
```diff
@@ -179,17 +125,20 @@ def generate_staypoints(positionfixes, method='sliding',
                 staypoint['id'] = staypoint_id
 
                 # point geometry of staypoint
-                staypoint.geometry = Point(group[name_geocol].x.mean(),
+                staypoint[name_geocol] = Point(group[name_geocol].x.mean(),
```
henrymartin1, line 128: Is there a reason why this should be changed? 
If so, it might be useful to introduce a convention.

trackintel/preprocessing/positionfixes.py
```diff
@@ -386,3 +336,91 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+                        staypoint['pfs_id'] = [pfs[j]['id']]
+
+                        ret_spts = ret_spts.append(staypoint, ignore_index=True)
+                i = j
+                break
+            j = j + 1
+    
+    return ret_spts
+
+def _generate_staypoints_dbscan_user(df,
```
henrymartin1, line 407: I think this function should get the metric parameter from the main function.

trackintel/preprocessing/positionfixes.py
```diff
@@ -386,3 +336,91 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
     # todo: triplegs dataframe has use the index as id
     # todo: proposed fix: ret_triplegs = ret_triplegs.set_index('id')
     return ret_pfs, ret_tpls
+
+
+def _generate_staypoints_sliding_user(df, 
+                                      name_geocol,
```
henrymartin1, line 342: Why do you need to have an explicit parameter for "name_geocol"? It is in `df` (which probably should be gdf) isn't it?

trackintel/preprocessing/positionfixes.py
```diff
@@ -386,3 +336,91 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+    transform_crs = None
+    if transform_crs is not None:
+        pass
+
+    # get staypoint matching
+    p = np.array([[radians(g.y), radians(g.x)] for g in df[name_geocol]])
+    labels = db.fit_predict(p)
+
+    # add positionfixes - staypoint matching to original positionfixes
+    df['staypoint_id'] = labels
```
henrymartin1, line 424: maybe you could rename "df" to "pfs" as you did in the sliding function. I think this might be easier to read (if the function is only used for pfs...)

trackintel/preprocessing/positionfixes.py
```diff
@@ -60,114 +66,54 @@ def generate_staypoints(positionfixes, method='sliding',
...
-                pass
-
-            # get staypoint matching
-            coordinates = np.array([[g.x, g.y] for g in user_positionfixes['geom']])
-            labels = db.fit_predict(coordinates)
-
-            # add positionfixes - staypoint matching to original positionfixes
-            ret_pfs.loc[user_positionfixes.index, 'staypoint_id'] = labels
-
+        ret_pfs = ret_pfs.groupby("user_id").apply(_generate_staypoints_dbscan_user,
```
henrymartin1, line 110: I didn't try it out but doesn't this generate non-unique staypoint ids? E.g. the staypoint ID starts at 0 for every user?

henrymartin1 (changes requested):
Major: 
- I am not so sure how to implement the "id" requirement in the csv function. It might be best to add the requirement that the index name is "id" to the accessors. I have left a longer comment in  trackintel/io/file.py
- is staypoint ID unique for DBScan method? (line 110 in  trackintel/preprocessing/positionfixes.py)
Minor:
- Naming/parameter stuff for generate staypoint methods
- Some missing documentation for tests (comments in file)

henrymartin1:
Nice PR! Especially for positionfixes performance is really important... 
I added some comment but most of them where questions about naming or documentation. The only issue that we might to discuss further is the question with the requirements for the index name. Is it necessary to require the index to have the name "id" or could we just go on and use whatever is the index?

tests/preprocessing/test_positionfixes.py
```diff
@@ -2,30 +2,60 @@
...
+        pfs_for, spts_for = _generate_staypoints_original(pfs_ori, 
+                                                          method='sliding', 
+                                                          dist_threshold=25, 
+                                                          time_threshold=300)
+        
+        pd.testing.assert_frame_equal(spts_groupby, spts_for, check_dtype=False)
+        pd.testing.assert_frame_equal(pfs_groupby, pfs_for, check_dtype=False)
+        
+        ### dbscan method
+        # stps detection using groupby
```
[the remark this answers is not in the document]
hong2223: I quickly fixed the dbscan method for this test. 
mainly 
num_samples -> default 1 
staypoint.geometry -> staypoint[name_geocol]

trackintel/io/file.py
```diff
@@ -81,6 +82,12 @@ def read_positionfixes_csv(*args, **kwargs):
             # dateutil parser timezones are sometimes not compatible with pandas (e.g., in asserts)
             tz = df[col].iloc[0].tzinfo.tzname(df[col].iloc[0])
             df[col] = df[col].dt.tz_convert(tz)
+    # set id as index
+    if "id" in df.columns:
+        df.set_index("id", inplace=True)
+    else:
+        df.index = list(df.index)
```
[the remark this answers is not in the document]
hong2223: changing the default [rangeindex](https://pandas.pydata.org/docs/reference/api/pandas.RangeIndex.html#pandas.RangeIndex) into [int64index](https://pandas.pydata.org/docs/reference/api/pandas.Int64Index.html#pandas.Int64Index), as index created by our generate function is Int64Index and they are sometimes not compatible (e.g., using the join method)

trackintel/preprocessing/positionfixes.py
```diff
@@ -24,19 +29,20 @@ def generate_staypoints(positionfixes, method='sliding',
...
 
     time_threshold : float, default 300 (seconds)
         The time threshold for the 'sliding' method in seconds, i.e., how long someone has to 
         stay within an area to consider it as a staypoint.
 
-    epsilon : float, default 100
-        The epsilon for the 'dbscan' method.
-
     dist_func : function, defaut haversine_dist
         A function that expects (lon_1, lat_1, lon_2, lat_2) and computes a distance in meters.
```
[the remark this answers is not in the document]
hong2223: this docstring comes from old age :). I think the docs and the way we handle the distance function in the `generate_location` function are quite good and we can also use them here

trackintel/preprocessing/positionfixes.py
hong2223: here 0 staypoint is identified. Otherwise there will be an error in `ret_spts[['pfs_id']].to_dict()['pfs_id']`, because there is no 'pfs_id'

trackintel/preprocessing/positionfixes.py
hong2223: Yes! you caught me 💯. The original function has also this behaviour and this is left as a todo (also because there are no tests )... We can also use the code in `generate_location`, they are really similar.

trackintel/preprocessing/positionfixes.py
hong2223: `staypoint` is a dict and has no `.geometry`

trackintel/preprocessing/positionfixes.py
hong2223: I think we cannot use int for mixed int and np.nan columns, [here](https://pandas.pydata.org/pandas-docs/stable/user_guide/gotchas.html#support-for-integer-na). Simple workaround is to use float. I should add more comments here

trackintel/preprocessing/positionfixes.py
hong2223: I encountered this strange behaviour of not having `gdf.geometry` after `gdf.groupby()`, [here ](https://github.com/geopandas/geopandas/issues/1777)for open bugs.

trackintel/preprocessing/positionfixes.py
hong2223: Yes, I agree. In the previous version the DBSCAN just used L2 distance and ignores the distance function. This will be fixed in the fix DBSCAN pull request

hong2223:
I had a closer look at using id as index name. It might not be necessary to have a name for the index (we never used the imported index name), and we are also not interested in the actual meaning of the index. The index provides matching between generated level (e.g., spts) and the original level (e.g., pfs), and this link is deleted just in our function. So now: 
- We no longer set 'id' as index in read functions; the original index is used in the generate function. For that I changed the linking in the big `while` (positionfixes.py line 385 and 401).
- We still set 'id' as index in our newly generated gdf (e.g., stps).

This solves the problem of managing the index while importing, I am not sure if I missed some cases though...

Additionally, docstrings and comments are added. Some other comments are related to the dbscan method and I think the best is to solve them in a new pull request.

hong2223:
As discussed in the meeting, users are now suggested to pass the `index_col` argument in the `read_from_csv` functions of `pfs` and `spts`. If not specified, a warning will be raised. 
The `from_geopandas` functions are unchanged, so it would require the user to use `set_index` before using the function, see changes in `test_from_geopandas.py`

The tons of changes in files mainly fix the warning in tests

trackintel/preprocessing/positionfixes.py
henrymartin1: Ah ok. There I didn't notice iin the places case ;-)
Can you document this/these problem/s somehow before we forget it. I think it would be enough to add a #todo note in the code and open an issue for it.

henrymartin1 (changes requested):
Hey Ye, 
sorry that it took me so long... Only a small change before we can merge it. I think it would be important to document the dbscan user_id bug/behavior somehow before we forget about it. I think it would be enough to add a todo note in the code that explains the bevhaviour and maybe gives a hint on how to solve it (last thing is optional).

The other comments are optional.

tests/model/test_staypoints.py
```diff
@@ -0,0 +1,22 @@
...
+        assert stps.as_staypoints
+
+        stps = stps.drop(['geom'], axis=1)
+        with pytest.raises(AttributeError):
+            stps.as_staypoints
+
+    def test_staypoints_center(self):
+        stps_file = os.path.join('tests', 'data', 'staypoints.csv')
+        stps = ti.read_staypoints_csv(stps_file, sep=';', index_col='id')
+        assert len(stps.as_staypoints.center) == 2
```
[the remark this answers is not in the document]
hong2223: I think it checks whether stps has both lat and lon. Here `.center` is like calculating the mean (lat, lon) pairs of the df.

henrymartin1 (approved):
41 comments, 16 commits and 23 files changed. This pull request was quite a project, nice work!👍

Issue #89, opened 2021-01-26
Test for preprocessing.positionfixes


hong2223:
Tests are needed for the different cases in 'generate_triplegs' function from preprocessing.positionfixes.
Case 1 and case 2 do not assign 'tripleg_id' to the original positionfixes df.

henrymartin1:
I agree that the generate_tripleg function is currently very complex and partially untested. 
Important steps would be:
- Break up the different cases using a "method" argument that is given by the user
- Encapsulated the different cases into functions 
- Create tests for the different functions

PR #124, opened 2021-02-26
CLN: Rework tripleg generation #89

henrymartin1:
- tripleg_generation is now encapsulated and fully tested
- tripleg_generation now includes the trip before the first and after the last staypoint
- all tripleg_generation variants return tripleg_ids to positionfixes


tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+    pfs_file = os.path.join('tests', 'data', 'geolife', 'geolife_positionfixes_with_ids.csv')
+    pfs = ti.read_positionfixes_csv(pfs_file, index_col='id')
+    pfs.geometry = pfs.geometry.set_crs(epsg=4326)
+    return pfs
+
+
+def assert_geodataframe_equal(gdf1, gdf2, **kwargs):
+    """assert equal geometry and data seperatly for two geodataframes"""
+    assert (gdf1.geometry.geom_almost_equals(gdf2.geometry)).all()
+    pd.testing.assert_frame_equal(gdf1.drop('geom', axis=1), gdf2.drop('geom', axis=1),
```
hong2223, line 59: minor: can we access the `geom` column via `.geometry`?

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
 from shapely.geometry import Point
-from math import radians
+from sklearn.cluster import DBSCAN
 
 import trackintel as ti
 from trackintel.geogr.distances import haversine_dist
 
+
+@pytest.fixture
+def pfs_geolife():
```
hong2223, line 18: should these `fixture` functions also be placed at the end of the file (following convention)? And their name with prefix `_`?

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+
+@pytest.fixture
+def testdata_pfs_ids_geolife():
+    pfs_file = os.path.join('tests', 'data', 'geolife', 'geolife_positionfixes_with_ids.csv')
+    pfs = ti.read_positionfixes_csv(pfs_file, index_col='id')
+    pfs.geometry = pfs.geometry.set_crs(epsg=4326)
+    return pfs
+
+
+def assert_geodataframe_equal(gdf1, gdf2, **kwargs):
```
hong2223, line 56: same as the fixture functions. Move to the end?

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+
+        assert (tpls.geometry.geom_almost_equals(testdata_tpls_geolife.geometry)).all()
+        pd.testing.assert_frame_equal(tpls.drop('geom', axis=1), testdata_tpls_geolife.drop('geom', axis=1),
+                                      check_like=True)
+
+    def test_generate_triplegs_case3(self, geolife_pfs_spts_short, testdata_tpls_geolife):
+        """Test tripleg generation using the 'between_staypoints' method for case 3.
+        case 3: triplegs from positionfixes with column 'staypoint_id' but without staypoints."""
+
+        pfs, spts = geolife_pfs_spts_short
```
hong2223, line 147: Here `spts` is not used, maybe use `_` to catch the return instead

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
-        tpls_file = os.path.join('tests', 'data', 'geolife', 'geolife_triplegs_short.csv')
-        tpls_test = ti.read_triplegs_csv(tpls_file, tz='utc')
+        pfs, spts = geolife_pfs_spts_short
+        pfs, tpls = pfs.as_positionfixes.generate_triplegs(spts, method='between_staypoints')
 
-        assert len(tpls) > 0
-        assert len(tpls) == len(tpls)
+        assert (tpls.geometry.geom_almost_equals(testdata_tpls_geolife.geometry)).all()
+        pd.testing.assert_frame_equal(tpls.drop('geom', axis=1), testdata_tpls_geolife.drop('geom', axis=1),
+                                      check_like=True)
```
hong2223, line 129: Line 127-129 doing the same thing as function `assert_geodataframe_equal`?
Also Line 139-141, Line 150-152, Line 164-166, Line 181-183, Line 194-196

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+
+    def test_tripleg_ids_in_positionfixes_case1(self, geolife_pfs_spts_short, testdata_pfs_ids_geolife):
+        """
+        checks if positionfixes are not altered and if staypoint/tripleg ids where set correctly during tripleg
+        generation using the 'between_staypoints' method with positionfixes with 'staypoint_id' and staypoints as
+        input (case 2).
+        """
+        pfs, spts = geolife_pfs_spts_short
+        pfs, tpls = pfs.as_positionfixes.generate_triplegs(spts, method='between_staypoints')
```
hong2223, line 161: `tpls` is not used, use `_`.
Also Line 176,Line  191.

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+        assert (pfs.geometry.geom_almost_equals(testdata_pfs_ids_geolife.geometry)).all()
+        pd.testing.assert_frame_equal(pfs.drop('geom', axis=1), testdata_pfs_ids_geolife.drop('geom', axis=1),
+                                      check_like=True)
+
+    def test_tripleg_ids_in_positionfixes_case3(self, geolife_pfs_spts_short, testdata_pfs_ids_geolife):
+        """
+        checks if positionfixes are not altered and if staypoint/tripleg ids where set correctly during tripleg
+        generation using the 'between_staypoints' method with only positionfixes as input (case 3).
+        """
+        pfs, spts = geolife_pfs_spts_short
```
hong2223, line 190: `spts` not used

tests/preprocessing/test_positionfixes.py
```diff
@@ -1,116 +1,225 @@
...
+        """
+        checks if the results are same if different variants of the tripleg_generation method 'between_staypoints'
+        are used
+        """
+
+        pfs, spts = geolife_pfs_spts_long
+
+        pfs_case1, tpls_case1 = pfs.as_positionfixes.generate_triplegs(spts, method='between_staypoints')
+        pfs_case2, tpls_case2 = pfs.drop('staypoint_id', axis=1).as_positionfixes.generate_triplegs \
+            (spts, method='between_staypoints')
```
hong2223, line 208: For me using `\` to split the line reduces the readability. Is this common in the numpy standards?

trackintel/preprocessing/positionfixes.py
```diff
@@ -163,6 +162,11 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
     positionfixes : GeoDataFrame
         The positionfixes have to follow the standard definition for positionfixes DataFrames.
 
+    method: str
+        Method to create tripelgs. Can be one of the following: ('between_staypoints', )
+        'between_staypoints': A tripleg is then defined as all positionfixes between two staypoints. This method
+            requires either a column ``staypoint_id`` on the positionfixes or passing staypoints as an input.
+
     staypoints : GeoDataFrame, optional
```
hong2223, line 170: description of `staypoints` should be before `method`

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+
     Examples
     --------
     >>> pfs.as_positionfixes.generate_triplegs(staypoints)
     """
+
     # copy the original pfs for adding 'staypoint_id' column
     ret_pfs = positionfixes.copy()
-    
     name_geocol = ret_pfs.geometry.name
```
hong2223, line 196: name_geocol is never used

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
-                pfs_tripleg = pfs.iloc[position_first_posfix_tl:position_last_posfix_tl + 1]
+            # case 3: Only positionfixes with staypoint id for tripleg generation
+            elif case == 3:
+                generated_triplegs.extend(_triplegs_between_staypoints_case3(positionfixes_user_this, user_id_this))
 
-                # include every positionfix that brings you closer to the center 
-                # of the staypoint
+        # create tripleg dataframe
+        if len(generated_triplegs) == 0:
+            ret_tpls = gpd.GeoDataFrame(columns=['id', 'user_id', 'started_at', 'finished_at', 'geom', 'pfs_ids'],
```
hong2223, line 237: Here we have 'id' as column, and later also as index. I think the best is to not include `id` to ensure consistency

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+    position_last_posfix_last_stp = positionfixes.index.get_loc(index_last_posfix_last_stp)
+
+    pfs_tripleg = positionfixes.iloc[position_last_posfix_last_stp:]
+    generated_triplegs_list.append(_get_tripleg_record_from_psfs(pfs_tripleg, user_id_this, min_nb_of_points=2))
+
+    # filter None values
+    return list(filter(None, generated_triplegs_list))
+
+
+def _get_tripleg_record_from_psfs(pfs_tripleg, user_id_this, min_nb_of_points):
```
hong2223, line 322: This function is used by both case1 and case2? Would be nice to place it after the two functions, and maybe add two `_` (i.e., `__`) as prefix.

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+        elif staypoints is not None:
+            case = 2
+        elif "staypoint_id" in ret_pfs:
+            case = 3
+        else:
+            raise Exception('unknown case')
 
-        positionfixes_user_this = ret_pfs.loc[ret_pfs['user_id'] == user_id_this]  # this is no copy
-        pfs = positionfixes_user_this.sort_values('tracked_at')
         generated_triplegs = []
```
hong2223, line 215: It's a bit inconvenient here that `generated_triplegs` is a list of dict whose keys are defined in `_get_tripleg_record_from_psfs` (two functions inside). Later the keys are converted to columns and directly used in this function. It may cause issues in the later maintenance stage. Is it possible to define the columns just in this function? (or at least we can add a comment)

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
-                if len(coords) > 1:
-                    generated_triplegs.append({
-                        'id': curr_tripleg_id,
-                        'user_id': user_id_this,
-                        'started_at': started_at,  # pfs_tripleg['tracked_at'].iloc[0],
-                        'finished_at': finished_at,  # pfs_tripleg['tracked_at'].iloc[-1],
-                        'geom': LineString(coords)
-                    })
-                    curr_tripleg_id += 1
+        for tripleg_id, tripleg in ret_tpls.iterrows():
```
hong2223, line 247: Here I would assume iterations are quite slow, is it possible to use join or merge to directly assign the ids? (pfs on index and tpls on pfs_id). 
Check Line 97 in `trackintel\preprocessing\positionfixes.py` or Line 135-138 in `trackintel\preprocessing\triplegs.py`

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
-                    curr_tripleg_id += 1
-
-        # case 3: Only positionfixes with staypoint id for tripleg generation
+        ## ensure dtype consistency
+        # tpls id (generated by this function) should be float (missing values)
+        ret_pfs['tripleg_id'] = ret_pfs['tripleg_id'].astype('float')
+        # user_id of tpls should be the same as ret_pfs
+        ret_tpls['user_id'] = ret_tpls['user_id'].astype(ret_pfs['user_id'].dtype)
+
+        # todo: proposed fix: ret_triplegs = ret_triplegs.set_index('id')
```
hong2223, line 258: ret_triplegs already has the id in the index? line 242

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+        return ret_pfs, ret_tpls
+
+    else:
+        raise NameError('Chosen method is not defined')
+
+
+def _triplegs_between_staypoints_case1(positionfixes, staypoints, user_id_this):
```
hong2223, line 265: These functions should be placed after the `_generate_staypoints_sliding_user` functions

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+        # - store or skip the current positionfix based on the state of 'status'
+
+        # During the loop the status of a positionfix can be {'in_tripleg', 'in_staypoint', 'tripleg_starts',
+        # 'tripleg_ends'}
+        if prev_pf is None and np.isnan(pf['staypoint_id']):
+            status = 'in_tripleg'
+        elif prev_pf is None and ~np.isnan(pf['staypoint_id']):
+            status = 'in_staypoint'
+
+        elif ~np.isnan(prev_pf['staypoint_id']):
```
hong2223, line 438: I am not able to understand what happened here as the logic is quite complicated - -!. But Vscode told me that `Value 'prev_pf' is unsubscriptable`, maybe it's because prev_pf is initialized as None and only assigned as the previous pf in the first loop (at the end). It might not cause an error, but I think it is not safe in later stages. Is there a way to fix this?

trackintel/preprocessing/positionfixes.py
```diff
@@ -173,187 +177,335 @@ def generate_triplegs(positionfixes, staypoints=None, *args, **kwargs):
...
+def _triplegs_between_staypoints_case1(positionfixes, staypoints, user_id_this):
+    """
+    This function uses the staypoints and the column 'staypoint_id' in the positionfixes, to identify all
+    positionfixes that lie in between two staypoints.
+
+    Parameters
+    ----------
+    positionfixes: trackintel positionfixes
+    staypoints: trackintel staypoints
+    user_id_this:
```
hong2223, line 274: This has no description, also in later functions. But it is already nice that you provide docstrings for these functions as well 👍

hong2223 (changes requested):
Nice and complicated work! It surprised me that you can sort this out (and nicely!). 
I posted some comments before we can merge this, most of them are style issues.

tests/preprocessing/test_positionfixes.py
henrymartin1: I think for fixtures the `_` convention doesn't really apply because you wouldn't import anything from a test script anyways. I tried to follow the [pytest documentation for fixtures](https://docs.pytest.org/en/stable/fixture.html) and there the fixtures are at the top. I think the idea is to organize the file similar to the life cycle of a test (Arrange, Act, Assert, Cleanup)

tests/preprocessing/test_positionfixes.py
henrymartin1: 👍

tests/preprocessing/test_positionfixes.py
henrymartin1: 👍

tests/preprocessing/test_positionfixes.py
henrymartin1: 👍 
I also changed it so that all (new) tests use it where possible

tests/preprocessing/test_positionfixes.py
henrymartin1: Yes ;-) I wrote the `assert_geodataframe_equal` later when it annoyed me but I never used it for the tests that where already written. I fixed that now.

trackintel/preprocessing/positionfixes.py
henrymartin1: 👍

trackintel/preprocessing/positionfixes.py
henrymartin1: Ah no, actually not. I fixed it 👍

trackintel/preprocessing/positionfixes.py
henrymartin1: 👍

tests/preprocessing/test_positionfixes.py
henrymartin1: 👍

tests/preprocessing/test_positionfixes.py
henrymartin1: 👍

tests/preprocessing/test_positionfixes.py
henrymartin1: yes, but splitting within parentheses (which is possible here) should be preferred: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

trackintel/preprocessing/positionfixes.py
henrymartin1: 👍

trackintel/preprocessing/positionfixes.py
henrymartin1: 👍

trackintel/preprocessing/positionfixes.py
henrymartin1: After the if/else is now a ` ret_tpls = ret_tpls.set_index('id')` statement which was missing before. So that should fix that

trackintel/preprocessing/positionfixes.py
henrymartin1: 👍

trackintel/preprocessing/positionfixes.py
henrymartin1: Generally I like this json-style creation of dataframes but I see the point that it is hard to see what happens without diving deeper into the code. I have added some comments and two asssert statements (244 + 246) that might give the reader some security while going through the code. 
Other than that I am not sure how to change this without breaking up all the functions so I would rather refrain from it...

trackintel/preprocessing/positionfixes.py
henrymartin1: Yes, it is actually missing on purpose. After the : follows the datatype and the user_id could be anything, so I didn't knew what to put there and left it empty. 
I could also write 'various' or something

trackintel/preprocessing/positionfixes.py
henrymartin1: I adjusted the code a bit but these state-machine like implementations are never really pretty (nor safe super but that's what tests are for^^). I have added an extra condition with an extra flag for the first iteration (that's what the None was for) but I guess that the Vscode will now complain about that the variable might be called before being defined. I thought about giving it some initialization but I did not have a good idea that does not make it more dangerous instead of more safe.

henrymartin1:
Thanks for the comments! I think I could address all of them except for the bad-style issue in the case3 implementation

trackintel/preprocessing/positionfixes.py
henrymartin1: actually you were right and I got confused somehow. I have adjusted it as you said.

hong2223 (approved):

hong2223:
Thanks for the change and this big improvement. A huge increase in test coverage! 
The case3 implementation is now cleaner and I am able to understand what happened. vscode also did not complain!
I will now merge this pull request.

PR #125, opened 2021-03-01
TST: add tests for index starting from 0

### LINKED ISSUE #119: Ensure generated id starts from 0
`Agg_level='user'` option in `generate_locations` function generates `id` starting from 1, which is not consistent with `Agg_level='dataset'`. Need to have ids starting from 0.
Tests for this issue are also needed.

hong2223:
closes #119


tests/preprocessing/test_triplegs.py
```diff
@@ -53,14 +55,34 @@ def test_generate_trips_dtype_consistent(self):
...
-        stps, tpls, trips = ti.preprocessing.triplegs.generate_trips(stps, 
-                                                                     tpls, 
+        stps, tpls, trips = ti.preprocessing.triplegs.generate_trips(stps, tpls,
                                                                      gap_threshold=15, 
                                                                      id_offset=0)
         
         assert stps['user_id'].dtype == trips['user_id'].dtype
         assert trips.index.dtype == "int64"
         
+        assert stps['trip_id'].dtype == "float"
```
henrymartin1, line 65: I think this is what I mean in #131. Here we have the index as int64 and the references to that index as float...

henrymartin1 (approved):
Nice and clean! Easy to review :-)

tests/preprocessing/test_triplegs.py
hong2223: I had a look at the issue and I agree we could change into `Int64`. Now (I think) the framework is complete and all we need is to change `float` into `Int64` in generate functions and the dtype_consistent tests.

henrymartin1: You can notice that the framework gets more _complete_ because the pull requests are getting easier :D

PR #127, opened 2021-03-02
DOC: illustration of model

svenruf:
Added images to show model hierarchy.


hong2223:
Thanks for the nice example image!

PR #128, opened 2021-03-02
DOC: Added Geopandas import functions to documentation + cleanup

svenruf:
-included io.from_geopandas in documentation
-cleanup / bugfixes in io documentation


henrymartin1:
Thanks for taking care of the documentation! It's really important. 
We should be more aware of this when reviewing pull requests.

PR #134, opened 2021-03-03
ENH: Same distance unit as input #24

### LINKED ISSUE #24: Same distance unit as input
The `extract_staypoints` and `plotting` functions accept different distance units as input: the former is in meters while the latter in decimal degrees. It would be good if we could agree on one unit and use it consistently. IMO, meters would be a more user-friendly choice, partly because calling the function `meters_to_decimal_degrees` repeatedly would decrease readability.

BTW, the input units are not specified in the documentation.

bifbof:
Closes #24

### DIFF
--- a/examples/preprocess_trajectories.py
+++ b/examples/preprocess_trajectories.py
@@ -29,24 +29,24 @@ tpls = pfs.as_positionfixes.generate_triplegs(staypoints=spts)[1]
 tpls.as_triplegs.plot(out_filename='examples/out/gpsies_trajectory_triplegs.png', staypoints=spts,
                       staypoints_radius=100, plot_osm=True)
 
-sys.exit(0)
+# sys.exit(0)
 
 # Geolife trajectory.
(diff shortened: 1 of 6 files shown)


henrymartin1:
Was this closed on purpose? If it is closed we can not merge it. If you want to still do some changes that you could also leave it open as a draft.

bifbof:
Yes, I will move it to a different branch. So that I can work on that depository in the meantime.

Von: Henry Martin <notifications@github.com>
Gesendet: Mittwoch, 3. März 2021 16:44
An: mie-lab/trackintel <trackintel@noreply.github.com>
Cc: Christof Leutenegger <christof.leutenegger@hotmail.com>; State change <state_change@noreply.github.com>
Betreff: Re: [mie-lab/trackintel] ENH: Same distance unit as input #24 (#134)


Was this closed on purpose? If it is closed we can not merge it. If you want to still do some changes that you could also leave it open as a draft.

PR #135, opened 2021-03-03
ENH: Updated plotting functions to be more versatile

### LINKED ISSUE #24: Same distance unit as input

bifbof:
Closes #24 
Hope that it works now. 😅


tests/test_visualization.py
```diff
@@ -39,3 +39,13 @@ def test_staypoints_plot(self):
         stps.as_staypoints.plot(out_filename=tmp_file, radius=0.01, positionfixes=pfs, plot_osm=False)
         assert os.path.exists(tmp_file)
         os.remove(tmp_file)
+
+    def test_locations_plot(self):
+        tmp_file = 'tests/data/locations_plot.png'
```
henrymartin1, line 44: Use os.path module to construct operating system independent paths. I am actually surprised that this works on windows but for consistency you should construct the path using os.path anyways (as done in the other functions).

trackintel/visualization/positionfixes.py
```diff
@@ -4,28 +4,42 @@
...
     """
-    _, ax = regular_figure()
+    if axis is None:
+        _, ax = regular_figure()
+    else:
+        ax = axis
+
+    crs_wgs84 = 'EPSG:4326'
+    if positionfixes.crs is None:
+        Warning("Coordinate System (CRS) is not set, default to WGS84.")
```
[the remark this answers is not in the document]
henrymartin1: Here is an example: [https://github.com/mie-lab/trackintel/blob/master/tests/io/test_file.py#L26](https://github.com/mie-lab/trackintel/blob/master/tests/io/test_file.py#L26)

trackintel/visualization/triplegs.py
```diff
@@ -20,36 +20,42 @@ def plot_triplegs(triplegs, out_filename=None, positionfixes=None, staypoints=No
...
-            south = positionfixes.geometry.y.min()
-        else:
-            triplegs_bounds = triplegs.bounds 
-            west = min(triplegs_bounds.minx) - 0.03 #TODO: maybe a relative value instead of 0.03
-            east = max(triplegs_bounds.maxx) + 0.03
-            north = max(triplegs_bounds.maxy) + 0.03
-            south = min(triplegs_bounds.miny) - 0.03
-        plot_osm_streets(north, south, east, west, ax)
+    if axis is None:
+        _, ax = regular_figure()
```
henrymartin1, line 35: use the same style in  examples/preprocess_trajectories.py

trackintel/visualization/triplegs.py
```diff
@@ -20,36 +20,42 @@ def plot_triplegs(triplegs, out_filename=None, positionfixes=None, staypoints=No
...
+        The radius in meter with which circles around staypoints should be drawn.
+
     plot_osm : bool
         If this is set to True, it will download an OSM street network and plot 
         below the triplegs.
+
+    axis : matplotlib.pyplot.Artist, optional
+        axis on which to draw the plot
+
     """
```
henrymartin1, line 33: Please add example in docstring (see point 15 in the [docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html#sections)

henrymartin1 (changes requested):
Great work! Nice to add axis management and the cascading style that uses the existing functions. Like this it is much cleaner than before. 
I ran the code and everything looks good. I have added some minor syntax/documentation comments that need to be addressed and please double check if the plots in the documentation are still up to date (I think nothing changed there)

bifbof:
> please double check if the plots in the documentation are still up to date (I think nothing changed there)
It looks like everything works

Two things are different:
1. Aspect ratio but that wasn't changed in this pull request.
2. It seems through the unequal aspect ratio the circles are rather ellipses.

bifbof:
gpsies_trajectory_locations

bifbof:
Okay, this is strange. The warnings for `staypoints` and `positionfixes` do not work, but the warnings for `triplegs` and `locations` do work, although it is exactly the same test.

henrymartin1:
- I think the ellipses are rather a projection issue than an aspect ratio issue. If so, we could add an option to draw the staypoints in a projected coordinate system in a later pull request. 
- I'll have a look at the warnings to see where the problem might be.

henrymartin1:
- I could fix the warnings issue by using "warnings.warn" instead of "Warning". I think "Warning" would need to be raised (using the "raise" keyword) but would then count as an exception rather than a warning.
Anyways, it is a good example why it is a good Idea to test the warnings.
- I agree now that the distortion of the circles is not because of the projection but because of the scaling of the figure. It would be nice to have an option (that would probably be on by default) that makes sure that x and y axis are in the same scale. However I would rather have this as an additional issue so that we can merge this PR. 

There were two additional (unrelated) bugs in the example file that probably were there for a long time.

bifbof:
Nice thanks for the fix!
A fix for the same aspect in the figure should be quite easy.
It would be a good idea if we put the transformation to wgs84 into a separate function.
So are the plotting functions cleaner and the test functions not so messy.

henrymartin1 (approved):

henrymartin1:
Both ideas sound great! I think having clean and easy to understand plot functions is very valuable. Depending on the scope of the transformation function it could either be in visualization.util or somewhere in geogr. Could you open a new issue for these? I will then merge this PR.

henrymartin1:
Ready to merge. Thanks for the contribution! The visualizations are extremely important for the attractiveness of the framework and have been neglected for quite a while. So this was a much needed update.

PR #136, opened 2021-03-04
ENH: Int64 for missing datatypes

### LINKED ISSUE #131: Inconsistent datatypes for integer IDs with missin values in trackintel.
The IDs we are using are often integers that are represented as floats in order to handle missing data. While this always gave me a bad feeling because theoretically `float(1) != float(1) != 1` in practice this seems not to be an issue in pandas. However it remains bad style and it get's worse when the dtype of the IDs becomes inconsistent. This is currently the case in the trip generation where the resulting trips have the columns `origin_staypoint_id` and `destination_staypoint_id` which are float even though the `id` of the corresponding staypoints DataFrame is integer.
In general I would not recommend to change the datatypes given by the user as some use cases (e.g., postgresql/postgis) are not flexible in terms of datatypes. Meaning that they would reject the attempt to insert floats into an integer column.

As a fix I would follow the pandas guidelines and use [Nullable integer datatypes](https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html#integer-na) in the case of integer indices. 
see also the discussion [here](https://stackoverflow.com/questions/11548005/numpy-or-pandas-keeping-array-type-as-integer-while-having-a-nan-value). 
I think our main goal should then be to not alter the datatype given by the user and use true integer representations (for integers) otherwise.

hong2223:
closes #131 

`~` is a Bitwise NOT in python. Now replaced to `not`.


tests/preprocessing/test_positionfixes.py
```diff
@@ -87,7 +89,7 @@ def test_generate_staypoints_dtype_consistent(self):
                                                              dist_threshold=25,
                                                              time_threshold=5 * 60)
         assert pfs['user_id'].dtype == stps['user_id'].dtype
-        assert pfs['staypoint_id'].dtype == "float"
+        assert pfs['staypoint_id'].dtype == "Int64"
         assert stps.index.dtype == "int64"
```
henrymartin1, line 93: It seems weird on the first glance to have two different datatypes but it is probably the right thing to do, as there are no missing values in the index.
We just have to be careful if we evaluate index.dtype == dtype

henrymartin1 (approved):
Nice and clean! Thanks for the quick fix.

PR #137, opened 2021-03-05
ENH: Added option to set CRS in read_xx_csv function #123

### LINKED ISSUE #123: read_csv function should have an option to set the crs
At the moment it is not possible to define a crs while reading trackintel data directly from .csv. I think this would be a nice enhancement. 

Example code:
```python 
tpls_file = os.path.join('tests', 'data', 'geolife', 'geolife_triplegs_short.csv')
tpls = ti.read_triplegs_csv(tpls_file, tz='utc', index_col='id')
print(tpls.crs)
```

bifbof:
Closes #123

### DIFF
--- a/tests/io/test_file.py
+++ b/tests/io/test_file.py
@@ -8,12 +8,12 @@ import trackintel as ti
 class TestFile:
     def test_positionfixes_from_to_csv(self):
         orig_file = os.path.join('tests', 'data', 'positionfixes.csv')
-        mod_file = os.path.join('tests','data', 'positionfixes_mod_columns.csv')
+        mod_file = os.path.join('tests', 'data', 'positionfixes_mod_columns.csv')
         tmp_file = os.path.join('tests', 'data', 'positionfixes_test.csv')
 
         pfs = ti.read_positionfixes_csv(orig_file, sep=';', index_col="id")
         
-        column_mapping = {'lat':'latitude', 'lon':'longitude', 'time':'tracked_at'}
+        column_mapping = {'lat': 'latitude', 'lon': 'longitude', 'time': 'tracked_at'}
         mod_pfs = ti.read_positionfixes_csv(mod_file, sep=';', index_col="id", columns=column_mapping)
         assert mod_pfs.equals(pfs)
         pfs['tracked_at'] = pfs['tracked_at'].apply(lambda d: d.isoformat().replace('+00:00', 'Z'))
@@ -40,7 +40,7 @@ class TestFile:
 
     def test_triplegs_from_to_csv(self):
         orig_file = os.path.join('tests', 'data', 'triplegs.csv')
-        mod_file = os.path.join('tests','data','triplegs_mod_columns.csv')
+        mod_file = os.path.join('tests', 'data', 'triplegs_mod_columns.csv')
         tmp_file = os.path.join('tests', 'data', 'triplegs_test.csv')
         tpls = ti.read_triplegs_csv(orig_file, sep=';', tz='utc', index_col="id")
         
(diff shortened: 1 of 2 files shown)


bifbof:
What should a good test look like to check if the crs are working?

henrymartin1:
I think a good test could be that you read a csv file with the crs option set and then check afterwards that the crs of the Geodataframe is set correctly.

tests/io/test_file.py
```diff
@@ -22,6 +22,11 @@ def test_positionfixes_from_to_csv(self):
         pfs.as_positionfixes.to_csv(tmp_file, sep=';', columns=columns)
         assert filecmp.cmp(orig_file, tmp_file, shallow=False)
         os.remove(tmp_file)
+
+        crs = "EPSG:2056"
+        assert pfs.crs is None
+        pfs = ti.read_positionfixes_csv(orig_file, sep=';', index_col="id", crs=crs)
+        assert pfs.crs == crs
```
henrymartin1, line 30: Move code block into its own test.

For tests it is best if every test only tests a single specific functionality (see [atomic](https://defragdev.com/blog/2012/11/02/the-fundamentals-of-automated-testing-atomic.html) if you are interested). Our tests don't always adhere to this but we try our best ;-) 
This being said it would be great if you could paste the code into a new test. For this you probably have to copy parts of the test into the new one.
You could also replace the first assert `assert pfs.crs is None` with something that checks if there is a warning. (e.g., using the `with pytest.warns(UserWarning)` environment but it is up to you how exactly you create the test.

tests/io/test_file.py
```diff
@@ -52,6 +55,11 @@ def test_triplegs_from_to_csv(self):
         tpls.as_triplegs.to_csv(tmp_file, sep=';', columns=columns)
         assert filecmp.cmp(orig_file, tmp_file, shallow=False)
         os.remove(tmp_file)
+
+        crs = "EPSG:2056"
+        assert tpls.crs is None
+        tpls = ti.read_triplegs_csv(orig_file, sep=';', tz='utc', index_col="id", crs=crs)
+        assert tpls.crs == crs
```
henrymartin1, line 63: same as for test_positionfixes_from_to_csv

tests/io/test_file.py
```diff
@@ -79,17 +85,21 @@ def test_staypoints_from_to_csv(self):
         stps.as_staypoints.to_csv(tmp_file, sep=';', columns=columns)
         assert filecmp.cmp(orig_file, tmp_file, shallow=False)
         os.remove(tmp_file)
+
+        crs = "EPSG:2056"
+        assert stps.crs is None
+        stps = ti.read_staypoints_csv(orig_file, sep=';', tz='utc', index_col="id", crs=crs)
+        assert stps.crs == crs
```
henrymartin1, line 93: same as for test_positionfixes_from_to_csv

tests/io/test_file.py
```diff
@@ -101,7 +111,12 @@ def test_locations_from_to_csv(self):
         locs.as_locations.to_csv(tmp_file, sep=';', columns=['user_id', 'elevation', 'center', 'extent'])
         assert filecmp.cmp(orig_file, tmp_file, shallow=False)
         os.remove(tmp_file)
-    
+
+        crs = "EPSG:2056"
+        assert locs.crs is None
+        locs = ti.read_locations_csv(orig_file, sep=';', index_col="id", crs=crs)
+        assert locs.crs == crs
+
```
henrymartin1, line 119: same as for test_positionfixes_from_to_csv

henrymartin1 (changes requested):
Only a last detail, sorry for being picky...
- Move test code into a new test
[optional]: Add warnings again if no crs is provided. Just out of curiosity, why did you remove the warnings? I think it is a good idea not to assume WGS84 by default but I think the warning was nice.

bifbof:
The warning for the crs is in the other pull request #135.
I would only add a warning if we want to enforce a crs I think it not necessary to always provide a crs.
Else the ``pyproj.CRS.from_user_input()`` function will handle any exceptions.
But I'll move the crs testing to a self contained function.

henrymartin1:
Ah ok, than I got confused ;-) 
Great, once the tests are moved we can merge I think.

bifbof:
Thanks for the advice. :)

henrymartin1 (approved):
Great work, thanks to the addition to trackintel!

PR #138, opened 2021-03-05
BUG: Corrected wrong error message in trips accessor. #122

### LINKED ISSUE #122: Wrong error message in trips accessor
The _validate method in the trips accessor raises the error message of staypoints: 
> "To process a DataFrame as a collection of staypoints, "

This should be updated to an error message that fits trips

bifbof:
Closes #122


henrymartin1 (approved):
Nice work, thanks for the fix!

PR #143, opened 2021-03-07
BUG: fix time_threshold bug in generate_staypoints

### LINKED ISSUE #132: generate_staypoints produce stps with duration smaller than time_threshold
Some of the `stps` generated by the function have very short durations (even 0). The problem is in file `trackintel\trackintel\preprocessing\positionfixes.py`

I am not very sure the cause, but it seems `delta_t` and the `if` statement in Line 308 is correct, but in Line 319 `staypoint['finished_at'] = pfs[j - 1]['tracked_at']` the `j-1` is different than `j` used elsewhere.

hong2223:
closes #132, added test for checking time_threshold.
Now our algorithm has exactly the same behaviour as Li et al. (2008).

I checked the commit history and the reason we changed Line 316 into `j-1` is because
`# TODO: should this not be j-1? because j is not part of the staypoint. DB: Changed.`
I fully agree with this and think this is a bug in the original algorithm. As the `jth` point is not part of the stps, it should not be considered in the coordinate and time calculation. Returning back to `j-1` might also not be desired: in the first Stay point example of losing GPS signal we will create a stps with duration 0. 

A proposed fix could be to exclude the `jth` point in coordinate and include the `jth` point as the `finished_at` time (by that we assume the user 'teleport' from the stp to the `jth` point)


hong2223:
The change in `stp` generation is fundamental and affects many of our tests in preprocessing module. 
Currently, I have not changed the other tests. So our tests will not pass.

hong2223:
I would also question the need to regard the last point as staypoint (Line 331), which creates a point with duration 0. The condition is inside the distance and time threshold, which means we already add the `[i, j]` (or more correctly `j-1`) points to stps, so the `j == num_pfs - 1`th point is the first point outside the last stp. 
We could of course add a stp for the point here, but to ensure consistency (in the distance and time threshold) I would choose to discard this last point.

henrymartin1:
Hey, 
I finally managed to have a closer look at the code.

### The i/j problem
The more I think about it, the more what is currently implemented makes sense to me. The original algorithm has a problem if you consider gaps in the tracking data. Without gaps (and a high enough sample frequency) the algorithm should work fine. With gaps the problem that I tried sketching in the figure below could arise. 
From the image it should be clear that the points 1 to 6 should be part of the staypoint while point 7 should not. The _only_ problem now is that we use point 7 to determine if the staypoint is long enough (`delta_t = pfs[j]['tracked_at'] - pfs[i]['tracked_at']`) which leads to the case that the staypoints could have less duration than the threshold.

I think this actually sounds worse than it is if we alter the definition a little. We could say that this algorithm finds staypoints of minimal duration of `time_threshold` +/- the sample frequency and the maximum gap size.
I am very skeptical about including any information from point 7 in the staypoint generated from points 1-6. In reality, the true temporal end of the staypoint lies somewhere between the start of trackpoint 6 and trackpoint 7 but we have no means to find out where this could be.

image

### The last staypoint problem
I did not look at the current implementation as detailed as the i/j problem so I might miss something but still some comments.
I think we should not have staypoints of duration 0 ;-) but we should still include the last staypoint. The problem of the algorithm is that it only detects staypoints if a person _steps out of it_. So if you arrive at home at 8pm and stop tracking at midnight the last 4 hour staypoint would be lost. The idea of this last check was to include this case (see figure below) but it can very well be that the implementation did not achieve what we wanted. 

image

henrymartin1:
We agreed on the following:
- strictly implementing the algorithm as it is in the original paper. 
- Adding a flag to include (or not) the last staypoint.

We did not yet discuss the default value of this last_staypoint flag but to be consistent it should probably be off by default.

henrymartin1:
Well... it get's worse. 
The current implementation has a Bug that makes it deviate from the original algorithm implementation (see commit 61dc89e). However, if we fix it and strictly use the implementation as it is in the paper (except you find an error in my commit 61dc89e) then this generates super weird behavior in some special cases (actually happens for the first geolife user, so probably not so special after all).

See image below:
If you go into a building and lose the signal (posfix 1-3) and then you get your signal again if you are again in a staypoint (posfix 4-6) then the first positionfix of the second staypoint belongs to both staypoints. This leads to many problems because our implementation (and actually our data model) assume a unique matching of staypoints and positionfixes. And yeah... Point 4 does not belong to the second staypoint... 
In the second commit ced645a I added some asserts where this causes problems. One problem is that if we use joins with non-unique indices it will duplicate the positionfix but I am sure that there will be many more problems. 

At this point I can not say how much effort it is to get this algorithm working for us... 
As an alternative we could also go back to accepting gaps as what they are and return to the j-1 solution but add a note that it returns shorter staypoins if there is a gap afterwards.
Another alternative would be another algorithm... 


image

hong2223:
Yes indeed that will produce a lot of problems... wondering how did they make this work.
As a proposed fix consider the following:
image

We change the coordinate calculation into [i, j) and did not include the `jth` point into current stps (red arrow). Honestly I think this is a typo in their paper...
The time calculation stays the same (red line)

This will create inconsistent behavior in whether including `jth` point in the time and geo dimension, but I think this makes sense to me and maybe is the solution with the minimum change required. Otherwise excluding j also in the time dimension is too confusing for me...
Or we can see how to implement DBSCAN and make that our main option.

henrymartin1:
Good question whether or not this is a typo 🤷‍♂️. The good old story of looking at papers too closely ;-)
I think we'll need another meeting to discuss what we do with this algorithm but I honestly think that none of the choices will be good. We'll probably should go for either DBSCAN or find another algorithm. DBSCAN has the advantage that it is way faster. At the moment it really is a pain to work with this algorithm...

I think the usual approach with DBSCAN is to normalize coordinates and time to the same scale and fit it on all three dimensions (meaning x,y,t). We usually used chebyshev distance for that to ensure that no dimensions is too far away but there will be some thinking necessary to translate the input parameters (max time + max distance) into the DBSCAN parameters. With this approach the `min_pts` parameter would probably fixed to a low number (2?) and everything else is done by `chebyshev` + `eps`.

An alternative I know of is to use the `min_pts` parameter to model time but this only works if the sampling rate is very regular so it is probably not suitable for us.

hong2223:
I will close this pull request and open a new one

PR #147, opened 2021-03-09
ENH: Added function for transforming into WGS84.

### LINKED ISSUE #146: Put transformation of crs for plotting in designated function.
With #135 we introduced a check to transform the data to WGS84 for plotting.
Since this method is repeated four times, it would be good to create a function for that in `visualization.util`. This way we can test for bugs in one place and the plot function will be cleaner.

bifbof:
Closes #146


bifbof:
> The tests contain some warnings, can you try to fix the warnings before we continue?
 
Some are from older commits, I'll fix them.

Should we move all the other tests for visualization also in this PR or should we rather do that in a seperate one?

henrymartin1 (commented):
Thanks for the contribution!

PR #148, opened 2021-03-09
CLN: Added explicit parameters for read_csv functions.

### LINKED ISSUE #111: explicit arguments for read_positionfixes_csv
Can [read_positionfixes_csv](https://github.com/mie-lab/trackintel/blob/master/trackintel/io/file.py#L39) have `columns`, `tz` as an explicit argument? Main goal would be to improve readability. 

I think this applies also to the other read_csv functions

bifbof:
Closes #111

### DIFF
--- a/trackintel/io/file.py
+++ b/trackintel/io/file.py
@@ -38,7 +38,7 @@ def localize_timestamp(dt_series, pytz_tzinfo, col_name):
     return dt_series.apply(pd.Timestamp, tz=timezone)
 
 
-def read_positionfixes_csv(*args, columns=None, tz=None, index_col=None, crs=None, **kwargs):
+def read_positionfixes_csv(*args, columns=None, tz=None, index_col=object(), crs=None, **kwargs):
     """Wraps the pandas read_csv function, extracts longitude and latitude and
     builds a geopandas GeoDataFrame. This also validates that the ingested data
     conforms to the trackintel understanding of positionfixes (see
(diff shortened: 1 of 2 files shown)


trackintel/io/file.py
```diff
@@ -74,14 +74,14 @@ def read_positionfixes_csv(*args, **kwargs):
...
     """
-    columns = kwargs.pop('columns', {})
-    tz = kwargs.pop('tz', None)
-    crs = kwargs.pop('crs', None)
-    
+    columns = {} if columns is None else columns
+
     # Warning if no 'index_col' parameter is provided
-    if 'index_col' not in kwargs:
+    if index_col is None:
```
henrymartin1, line 80: I think the implementation now has a problem because there are some cases where you can not get rid of the warning.
The problem is that there are cases where there is no index in the .csv file and a new index should be generated and cases where there is an index in the .csv file but the user forgot to define the column. Our Idea now was to check if the argument `index_col` was set explicitly to None by the user. In a way it was like if there are two different 'None' cases. 

There is a discussion about this on [stackoverflow](https://stackoverflow.com/questions/14749328/how-to-check-whether-optional-function-parameter-is-set) but most of the solutions are rather wild...
I think the best option would be [this](https://stackoverflow.com/a/14749388) where the default value is something that the user would never pick. E.g., `index_col=object()` and then check if the parameter is an object and set it to None afterwards. 
If we do it like this, then the warning could stay the same as it would be prevented if `index_col` is set to None explicitly.

hong2223, line 80: I think here the behaviour is not what we expected, There are three cases:
1. User pass `index_col` explicitly, we pass it to pd.read_csv().
2. User do not pass `index_col`, we raise a warning.
3. User pass `index_col=None`, no warning should be raised as the user should be aware that we assume index as the unique identifier.

I think the third case is not covered here. I am not sure how to implement this though, is there a way to distinguish user passed None vs. the default None?

hong2223 (changes requested):
Nice fix to the issue. The index_col argument needs some modification

henrymartin1 (changes requested):
Looks good. There is one issue with the index_col argument and the related warning that I commented on in the code. I only left the comment at one position but it applies to all functions.

henrymartin1 (approved):
Nice, thanks for the changes! Ready to merge

PR #153, opened 2021-03-10
ENH: black CI

hong2223:
CI of black which checks if the code from the current pull request (and repo) adheres to the black format. 
Would fail if some files need to be changed. **It will not automatically convert code into black style**

Additional parameter `-l 120` sets the line length to 120


henrymartin1:
Ah, I see. I think it is still a very valuable workflow if we decide to go black. I would say we merge this as soon as the test passes and then it will pass forever ;-) 
We should add a section to code-style and black to contributing.md as well then.

PR #156, opened 2021-03-15
BUG: update image path

### LINKED ISSUE #149: pypi doesn't show images in readme
Fix: use absolute url for images in readme instead of relative paths

[trackintel-pypi](https://pypi.org/project/trackintel/)

hong2223:
shall close #149 according to [here](https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github)


henrymartin1 (approved):
Then let's hope for the best ;-)

PR #159, opened 2021-03-15
Wgs84 length

### LINKED ISSUE #158: Support length calculation of wgs84 triplegs
It would be very useful if we would have a function to directly calculate the length of WGS84 triplegs as it is done inside the `predict_transport_mode_simple_coarse` function

henrymartin1:
fixes #158


henrymartin1:
This pull request has synchronization errors and was mixed with another branch. It is replaced by #161

PR #160, opened 2021-03-15
ENH/BUG: gaps in triplegs

svenruf:
Added the possibility to pass a gap_threshold to the generate_triplegs method. If the temporal difference between two positionfixes is greater than gap_threshold, a new tripleg will be started (without creating a staypoint).

Cases 1+2 work properly (tested with the long and the short geolife data set). In case 3 is really tricky, to integrate the funcionality as many diferent cases (staypoint followed by a gap, staypoint followed by tripleg with too less points, ...) have to be covered and the architecture is completely different to 1+2. All cases in the short geolife data set are covered, but in the long data set there are situations where case 3 behaves different than 1+2. 

The solution once mentionnned in the meeting - to check the triplegs after the generation - is not possible as the timestamps are not stored in the triplegs, therefore the gaps cannot be recognized. 

If you have a good idea how to integrate the functionality cleaner in case 3, I'm very open to it.


svenruf:
Will be covered by #170

PR #161, opened 2021-03-15
ENH: Wgs84 length

henrymartin1:
Functionality to calculate the length of triplegs (or linestrings) using the haversine distance.
(The pull request before got mixed with another pull request)

### DIFF
--- a/tests/analysis/test_transport_mode_identification.py
+++ b/tests/analysis/test_transport_mode_identification.py
@@ -1,15 +1,11 @@
 # -*- coding: utf-8 -*-
 
-import datetime
 import os
 
-import pytest
 import numpy as np
-import geopandas as gpd
-from numpy.testing import assert_almost_equal
+import pytest
 
 import trackintel as ti
-from trackintel.io.dataset_reader import read_geolife
 
 
 class TestTransportModeIdentification:
@@ -30,9 +26,8 @@ class TestTransportModeIdentification:
         tpls_file = os.path.join('tests', 'data', 'triplegs_transport_mode_identification.csv')
         tpls = ti.read_triplegs_csv(tpls_file, sep=';', index_col="id")
 
-        with pytest.warns(UserWarning, 
-                          match='Your data is not projected. WGS84 is assumed and for length ' + 
-                          'calculation the haversine distance is used'):
+        with pytest.warns(UserWarning, match='Your data is not projected. WGS84 is assumed and for length calculation'
+                                             ' the haversine distance is used'):
             tpls_transport_mode = tpls.as_triplegs.predict_transport_mode(method='simple-coarse')
 
         assert tpls_transport_mode.iloc[0]['mode'] == 'slow_mobility'
@@ -62,13 +57,12 @@ class TestTransportModeIdentification:
         assert tpls_transport_mode_3.iloc[2]['mode'] == 'fast_mobility'
 
     def test_simple_coarse_identification_geographic(self):
-        """Asserts the correct behaviour with data in geographic coordinate systems."""
+        """Asserts that a warning is thrown if data in non-WGS geographic coordinate systems."""
         tpls_file = os.path.join('tests', 'data', 'triplegs_transport_mode_identification.csv')
         tpls = ti.read_triplegs_csv(tpls_file, sep=';', index_col="id")
         tpls_2 = tpls.set_crs(epsg=4326)
         tpls_4 = tpls_2.to_crs(epsg=4269)
-        with pytest.raises(UserWarning,
-                           match='Your data is in a geographic coordinate system, length calculation fails'):
+        with pytest.warns(UserWarning):
             tpls_4.as_triplegs.predict_transport_mode(method='simple-coarse')
 
     def test_check_categories(self):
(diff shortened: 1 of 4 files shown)


henrymartin1:
fixes #158

tests/analysis/test_transport_mode_identification.py
```diff
@@ -30,9 +26,8 @@ def test_simple_coarse_identification_no_crs(self):
         tpls_file = os.path.join('tests', 'data', 'triplegs_transport_mode_identification.csv')
         tpls = ti.read_triplegs_csv(tpls_file, sep=';', index_col="id")
 
-        with pytest.warns(UserWarning, 
-                          match='Your data is not projected. WGS84 is assumed and for length ' + 
-                          'calculation the haversine distance is used'):
+        with pytest.warns(UserWarning, match='Your data is not projected. WGS84 is assumed and for length calculation'
+                                             ' the haversine distance is used'):
```
henrymartin1, line 30: I just changed this in the process because I had trouble getting this test to work. In the end the problem was something else

trackintel/analysis/transport_mode_identification.py
```diff
@@ -110,21 +101,15 @@ def identify_mode(tripleg, wgs, categories):
...
         duration = (tripleg['finished_at'] - tripleg['started_at']).total_seconds()
-        speed = distance / duration  # The unit of the speed is m/s
+        speed = tripleg['distance'] / duration  # The unit of the speed is m/s
 
         for bound in categories:
             if speed < bound:
                 return categories[bound]
 
-    triplegs['mode'] = triplegs.apply(lambda l: identify_mode(l, wgs, categories), axis=1)
+    triplegs['mode'] = triplegs.apply(lambda l: identify_mode(l, categories), axis=1)
```
hong2223, line 112: This is equivalent to `triplegs.apply(identify_mode, args=(categories), axis=1)`? see [here](https://stackoverflow.com/questions/12182744/python-pandas-apply-a-function-with-arguments-to-a-series)

trackintel/geogr/distances.py
```diff
@@ -170,3 +169,96 @@ def meters_to_decimal_degrees(meters, latitude):
...
+    >>> triplegs['length'] = calc_haversine_length_of_linestrings(triplegs)
+    """
+
+    assert all(gdf.geom_type == "LineString")
+
+    length = gdf.geometry.apply(calc_haversine_length_of_single_linestring)
+    return length
+
+
+def calc_haversine_length_of_single_linestring(linestring):
```
hong2223, line 236: This could be named to `_calc_haversine_length_of_linestring_single` to highlight the functions relation

trackintel/geogr/distances.py
```diff
@@ -170,3 +169,96 @@ def meters_to_decimal_degrees(meters, latitude):
...
+        is_wgs = True
+
+    elif crs is None:
+        is_wgs = True
+        warnings.warn('Your data is not projected. WGS84 is assumed and for length calculation the haversine '
+                      'distance is used')
+    return is_wgs
+
+
+def calc_haversine_length_of_linestrings(gdf):
```
hong2223, line 210: We have `calculate_distance_matrix` at the top, should we either use `calculate` or `calc` in both places?
Also, since linestring requirement is specified both in docstring and the assert, can be simplify this function to `calc_haversine_length`?

tests/geogr/test_distances.py
```diff
@@ -99,8 +137,17 @@ def test_meters_to_decimal_degrees(self):
...
         for degree, lat_output in input_result_dict.items():
             for lat, meters in lat_output.items():
-
                 decimal_degree_output = meters_to_decimal_degrees(meters, lat)
                 assert np.isclose(decimal_degree_output, degree, atol=0.1)
 
 
+class Testcalc_haversine_length_of_linestring:
+    def test_distance(self, gdf_ls):
+        dist = calc_haversine_length_of_linestrings(gdf_ls)
```
hong2223, line 146: Are there any asserts or checks here?

hong2223 (changes requested):
A nice cleanup of the original function. Just some minor naming issues before we can merge.

trackintel/geogr/distances.py
henrymartin1: I agree for the naming in general. But I think this function will potentially be used outside this file so I would take your name without the leading underscore

trackintel/geogr/distances.py
henrymartin1: In theory I would agree but in this case I would say that this function is really only useful for wgs84/haversine as you can use shapely/geopandas build-ins `linestring.length` or `gdf.length`.
So I would go with your first option.

tests/geogr/test_distances.py
henrymartin1: I added a weak one but it is more to see if it runs through. (length calculation is tested in the function below).

trackintel/analysis/transport_mode_identification.py
henrymartin1: 👍

henrymartin1 (commented):
Addressed all comments

hong2223 (approved):
Nice and neat 👍

Issue #27, opened 2020-10-06
Tripleg extraction is vulnerable to gaps in tracking data


henrymartin1:
At the moment, the triplegs are defined as all trackpoints between two staypoints. While this is often correct, we should make sure, that gaps in the tracking data are dealt with. Otherwise, this could lead to triplegs that span over multiple days. Here is an example: 
![grafik](https://user-images.githubusercontent.com/31069329/95238990-9ce6d200-080a-11eb-9e21-10f288000f48.png)

![grafik](https://user-images.githubusercontent.com/31069329/95238926-82acf400-080a-11eb-8774-74a06b290789.png)

Here is code to reproduce the results (to look at it in QGIS)
```python
pfs = read_geolife(os.path.join('examples', 'example_geolife', 'rawdata_geolife'))
spts = pfs.as_positionfixes.extract_staypoints(method='sliding', dist_threshold=25, time_threshold=5 * 60)
tpls = pfs.as_positionfixes.extract_triplegs(spts)
pfs.to_file("posfix.geojson", driver='GeoJSON')
spts.to_file("stp.geojson", driver='GeoJSON')
tpls.to_file("triplegs.geojson", driver='GeoJSON') ```

svenruf:
@henrymartin1 should it be possible to pass a threshold or do we define a fixed value for the max gap allowed?

henrymartin1:
I think it would be better to pass a threshold. Anybody has an idea for a default threshold value? Whats the typical duration for long tunnels in Switzerland? :D

svenruf:
Ok, then I will implement this. I think with 20 min the long tunnels are covered (gotthard base tunnel 17 min, gotthard street tunnel 13min).

svenruf:
I just saw that for the trips, the gap_threshold default is 15min... I think we can talk about this, the default value is also easy to change after the implementation :)

henrymartin1:
We agreed on reformulating the trip generation code in a simpler way.

PR #170, opened 2021-03-18
ENH: More elegant tripleg generation

henrymartin1:
Here is a first draft of a pull request that proposes a new and simpler tripleg generation.


henrymartin1:
fixes #27

henrymartin1:
I had a problem with the last commit message in `d1a7344` but this commit finishes the code however due to shifting the tripleg definition by 1 point (tripleg starts with the first point after a staypoint instead of the shared last point of the staypoint) the tests have to updated

svenruf:
I assume that you will also update the tests as you assigned yourself, right? If there's anything to do for me, let me know.

trackintel/preprocessing/positionfixes.py
```diff
@@ -261,26 +260,30 @@ def generate_triplegs(positionfixes_in, spts=None, method='between_staypoints',
...
+        #  `cond_all = cond_all & pd.isna(pfs['staypoint_id'])`
         tpls_starts = np.where(cond_all)[0]
         tpls_lengths = np.diff(tpls_starts)
-        cond_to_remove = np.take(tpls_starts,np.where(tpls_lengths<3)[0])
-        for i in cond_to_remove:
-            cond_all[i] = False
-            pfs.loc[i,'tripleg_id'] = -1
+        cond_to_remove = np.take(tpls_starts, np.where(tpls_lengths < 2)[0])  # a valid linestring needs 2 points
+        cond_all.iloc[cond_to_remove] = False
+        pfs.iloc[cond_to_remove].loc['tripleg_id'] = -1
```
svenruf, line 286: This causes a warning when I execute it, as it is trying to set the -1 on a copy of pfs

trackintel/preprocessing/positionfixes.py
```diff
@@ -204,52 +204,51 @@ def generate_triplegs(positionfixes_in, spts=None, method='between_staypoints',
...
+                # step 2
+                # Identify first positionfix after a staypoint
+                # find index of closest positionfix with equal or greater timestamp.
                 tracked_at_sorted = pfs_u['tracked_at'].sort_values()
                 insert_position_psf_u = tracked_at_sorted.searchsorted(spts_u['finished_at'])
                 insert_index_psf = tracked_at_sorted.iloc[insert_position_psf_u].index
 
                 # store the insert insert_position_spts_u in an array
-                insert_index_psf_list.extend(list(tracked_at_sorted.iloc[insert_position_psf_u].index))
+                insert_index_psf_list.extend(list(tracked_at_sorted.iloc[insert_index_psf].index))
```
svenruf, line 241: This causes a out of bound error when testing it with the geolife_long. removing the .iloc solves the error

svenruf:
@hong2223 as discussed I had a look and ran the tripleg generation. At the moment there are two things that need to have closer look, I commented directly into Henrys commit. I did not want to push the changes directly, as Henry did change this from my code and I don't know if it has a special reason and needs another solution.

trackintel/preprocessing/positionfixes.py
hong2223: I changed this now to `pfs.loc[pfs.index.isin(cond_to_remove), "tripleg_id"] = -1`

trackintel/preprocessing/positionfixes.py
hong2223: I think this is actually `insert_index_ls.extend(list(insert_index_psf))`

hong2223:
> @hong2223 as discussed I had a look and ran the tripleg generation. At the moment there are two things that need to have closer look, I commented directly into Henrys commit. I did not want to push the changes directly, as Henry did change this from my code and I don't know if it has a special reason and needs another solution.

Thanks for the check! I had a look at the function and completed its functionality. 
I messed up with the commits, so I will close this one (again!) and open a new one - -..

PR #171, opened 2021-03-18
CLN: Added missing docstring

### LINKED ISSUE #133: create_activity_flag missing correct docstring

henrymartin1:
fixes #133


docs/modules/preprocessing.rst
```diff
@@ -30,6 +30,7 @@ transfer between two transport modes). We can cluster these into locations that
 frequently visits.
 
 .. autofunction:: trackintel.preprocessing.staypoints.generate_locations
+.. autofunction:: trackintel.preprocessing.staypoints.create_activity_flag
```
hong2223, line 33: should we include some description of this function? for example just simply adding:

> Staypoints are points where someone stayed for a longer period of time (e.g., during a transfer between two transport modes). We can cluster these into locations that a user frequently visits **and/or infer if they correspond to activities.**

hong2223 (commented):
Just an optional comment on the preprocessing documentation :). The others are fine and ready to merge

docs/modules/preprocessing.rst
henrymartin1: Yes! We should show them what we have ;-) 
Is this in the readme?

hong2223: This is just 4 lines above :).

Issue #129, opened 2021-03-02
Inconsistent naming for io functions


henrymartin1:
At the moment we have:
- `read_positionfixes_csv`
- `read_positionfixes_postgis`
- `positionfixes_from_gpd`
So I would propose that we rename `positionfixes_from_gpd` to `read_positionfixes_gpd`

svenruf:
I originally named it like this, but in a meeting the argument (don't remember from who) was that it isn't actually a reading function, as it does only import/rename the GeoDataFrame. So I changed the name 😄

henrymartin1:
I think i can remember this 😬. I just tried using it and found it weird that it is named differently...

henrymartin1:
We agreed on double checking if pandas makes a distinction between reading and loading if it is not external. If there is no distinction, then we can rename the function in `read_positionfixes_gpd`

PR #172, opened 2021-03-19
CLN: Updated function name (#129)

abcnishant007:
fixes issue #129


henrymartin1 (approved):
Perfect on first try! You are now officially a contributer to trackintel 🥳 
Thanks for the fix.

hong2223:
thanks for the fix! I guess the change needs to be propagated to other from_geopandas function?
image

PR #175, opened 2021-03-21
PERF: datatime loading

hong2223:
When loading pfs from csv, the bottleneck is in `df['tracked_at'] = df['tracked_at'].apply(dateutil.parser.parse)`. This per-row operation is expensive and can be changed to `df["tracked_at"] = pd.to_datetime(df["tracked_at"])`, which is much faster for common datatime formats and will call `dateutil.parser.parse` if receiving non-common ones. Since we work directly with pd, line 96-99 is no longer needed.

Additional tests for setting tz is added.
The test stricture of test_file.py is incorrect, but I only changed the `TestPositionfixes` class.

### DIFF
--- a/trackintel/io/file.py
+++ b/trackintel/io/file.py
@@ -93,14 +93,10 @@ def read_positionfixes_csv(*args, columns=None, tz=None, index_col=object(), crs
     # transform to datatime
     df["tracked_at"] = pd.to_datetime(df["tracked_at"])
 
-    # check and/or set timezone
+    # set timezone if none is recognized
     for col in ['tracked_at']:
         if not pd.api.types.is_datetime64tz_dtype(df[col]):
             df[col] = localize_timestamp(dt_series=df[col], pytz_tzinfo=tz, col_name=col)
-        else:
-            # dateutil parser timezones are sometimes not compatible with pandas (e.g., in asserts)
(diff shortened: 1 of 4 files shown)


trackintel/io/file.py
```diff
@@ -85,18 +85,18 @@ def read_positionfixes_csv(*args, columns=None, tz=None, index_col=object(), crs
...
+    df["tracked_at"] = pd.to_datetime(df["tracked_at"])
 
-    # check and/or set timezone
+    # set timezone if none is recognized
     for col in ['tracked_at']:
         if not pd.api.types.is_datetime64tz_dtype(df[col]):
             df[col] = localize_timestamp(dt_series=df[col], pytz_tzinfo=tz, col_name=col)
-        else:
-            # dateutil parser timezones are sometimes not compatible with pandas (e.g., in asserts)
-            tz = df[col].iloc[0].tzinfo.tzname(df[col].iloc[0])
```
henrymartin1, line 98: Hm... I don't really remember why I put this there. There was a problem that `dateutil` and `pytz` timezones are sometimes not compatible. This line essentially transforms `dateutil` into `pytz` timezones or something like that. But maybe the error is not coming up because we are now using better asserts (e.g., is_datetime64tz_dtype)?

henrymartin1 (approved):
Looks good! Funny what single lines of code can change ;-)

PR #178, opened 2021-03-22
BUG: a proposed fix for stps generation

hong2223:
I checked [their](https://github.com/scikit-mobility/scikit-mobility/blob/master/skmob/preprocessing/detection.py) implementation, essentially they are also using the 2008 method, but considering [i, j] in time and [i, j) in space (median is used instead of mean for coordinates). I also went for this and maybe we add some description in the docs.
They added two minor improvements:
-  a gap threshold between consecutive pfs to determine whether to add a stp here. (Line 300-304).
- a speed threshold to delete the pfs with too high speed at the end of a stp. (not implemented here).

I slightly modified the structure (now using only one _for_) and tested using geolife. For one user it takes approx. 15s, which is quite fast I guess.

Known problems:
- the method only detects stps after the user steps out, so the last stp is always discarded. for `test_generate_staypoints_sliding_min` test this means it always generate `n-1` stp for `n` pfs. Proposed fix would be to add the last stp using a flag.
- we still have the matching problem between pfs and stp (they don't). Now the [i, j-1] pfs is considered as stp, although stp receives the [i, j] time. This means if we include the last pfs in stp in the tripleg generation (`generate_triplegs`), we will create overlapping time. See below (green stp ending time, red tpl starting time). The proposed fix would be not including the last pfs.
image

other minor changes:
- time_threshold parameter now accepts minutes instead of seconds, some tests are changed.
- print_progress parameter is added to show a progress bar for user-level generation.
- DBSCAN method is deleted for now.

some tests still fails and I will wait until we discussed a solution for the above problems.

### DIFF
--- a/tests/analysis/test_modal_split.py
+++ b/tests/analysis/test_modal_split.py
@@ -17,7 +17,7 @@ def read_geolife_with_modes():
 
     pfs, spts = pfs.as_positionfixes.generate_staypoints(method='sliding',
                                                          dist_threshold=25,
-                                                         time_threshold=5 * 60)
+                                                         time_threshold=5.0)
     _, tpls = pfs.as_positionfixes.generate_triplegs(spts, method='between_staypoints')
 
     tpls_with_modes = geolife_add_modes_to_triplegs(tpls, labels)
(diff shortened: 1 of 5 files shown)


henrymartin1:
I currently working on the tripleg generation and the more I work with the algorithm (current implementation, not this PR) the worse it gets. To me it is really nonsense to include the [i, j] staypoint in the staypoint because there are basically always gaps.... 
(see example picture). I think the additional gap threshold is a good work around until we find a better algorithm at some point
image

hong2223:
The new staypoint generation algorithm is ready!
The failed tests are for triplegs and trip generation, which need to wait for #170, and the change in trip_generation related test data

PR #182, opened 2021-03-22
TST: added tests for spatial_filter

hong2223:
added two tests for increasing test coverage.

### DIFF
--- a/tests/preprocessing/test_filter.py
+++ b/tests/preprocessing/test_filter.py
@@ -1,13 +1,33 @@
 import os
-
+import pytest
 import geopandas as gpd
+from geopandas.testing import assert_geodataframe_equal
 
 import trackintel as ti
 
 
+@pytest.fixture
+def locs_from_geolife():
+    """Create locations from geolife staypoints."""
+    # read staypoints
+    spts_file = os.path.join('tests', 'data', 'geolife', 'geolife_staypoints.csv')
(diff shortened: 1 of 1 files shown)


tests/preprocessing/test_filter.py
```diff
@@ -24,11 +44,13 @@ def test_filter_staypoints(self):
         
         assert len(within_spts) == gis_within_num, "The spatial filtered sp number should be the same as" + \
             "the one from the result with ArcGIS"
-        assert all(within_spts.geometry == intersects_spts.geometry), "For sp the result of within and" + \
-            "intersects should be the same"
         assert len(crosses_spts) == 0, "There will be no point crossing area"
         
+        # For staypoints the result of within and intersects should be the same
```
abcnishant007, line 49: @hong2223 within the comments, what is the meaning of _within_ and _intersects_ here?

abcnishant007 (commented):
@hong2223 I tested the pytests locally, they are fine; 
will approve after addressing that minor concern on the comments

tests/preprocessing/test_filter.py
hong2223: These correspond to the `method` parameter in `spatial_filter` [function](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/filter.py). These correspond to the shapely provided so-called _binary predictors_, see for example [here ](https://shapely.readthedocs.io/en/stable/manual.html#object.intersects)

PR #185, opened 2021-03-26
DOC: trackintel basic tutorial

svenruf:
- jupyter notebook that shows the core functionality of trackintel
- test that checks if the notebook runs without errors

Problem: To ensure that eventual errors of the notebook will cause the test to fail, nbconvert has to be executed with the flag --stdout. This will also print out figures when running python -m pytest in the command line, which is not so nice. If anyone knows a fixture to disable figure outputs during the test session, I would be happy to integrate this.


svenruf:
In addition, any feedback on the tutorial is very welcome :)

svenruf:
Does anyone know why the coverage is decreased, although I did not manipulate the other tests?

henrymartin1:
I think you shouldn't worry too much about this. The coverage can decrease if you delete lines of code that are covered by tests because then the ratio of tested and untested lines of code changes.

hong2223:
I cannot run the initial solution, the `read_positionfixes_gpd` function name is not updated in the jupyter notebook, yet the (original) test still passed. Also, I need to add the parent folder of "example" in order to import trackintel `(sys.path.append(os.path.abspath(os.path.join(os.getcwd(), os.pardir))))`

I changed the `test_tutorial.py` file, which works in my local environment but failed in the github tests. I will fix the problem

svenruf:
Yes, it runs and does the expected things. thank you!
Yes the function was not called correctly, I created the tutorial branch before this changes and forgot to update it...

hong2223:
the tests are working fine. I changed some typos in the tutorial and this is now ready to merge

PR #186, opened 2021-03-26
DOC/BUG: correct datamodels for temporal_tracking_quality

### LINKED ISSUE #184: temporal tracking quality doesn't work on positionfixes
Should either be implemented or return a correct warning (returns key error 'finished_at' at the moment) and at least be in the documentation of the method.

hong2223:
a quick fix that closes #184


hong2223:
Does this solve the problem you mentioned? @svenruf

svenruf (approved):
Thanks for the fix!

PR #190, opened 2021-03-31
DOC: provide new tutorial on binder

svenruf:
updated the binder badge, it is now linked to the new tutorial.

TODO: Decide what happens to the example geolife jupyter notebook


hong2223:
Thanks for the work! The tutorial works fine in binder (also with the `import trackintel` statement).
I will have a look at the old tutorial and probably delete it.

PR #212, opened 2021-05-05
ENH: Add simple location identifier.

bifbof:
Added simple location identifier freq. That algorithm simply selects the location that the user spend the most time on.
Other "recipes" will be much easier to program as they build all on the same principle.

### DIFF
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -37,7 +37,7 @@ def location_identifier(spts, method="FREQ", pre_filter=True, **pre_filter_kwarg
 
     References
     ----------
-    Chen, Qingqing, and Ate Poorthuis. 2021.
+    [1] Chen, Qingqing, and Ate Poorthuis. 2021.
     ‘Identifying Home Locations in Human Mobility Data: An Open-Source R Package for Comparison and Reproducibility’.
     International Journal of Geographical Information Science 0 (0): 1–24.
     https://doi.org/10.1080/13658816.2021.1887489.
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -46,7 +46,7 @@ def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs
     >>> ti.analysis.location_identifier(sps, pre_filter=True, recipe="FREQ")
     """
     assert sps.as_staypoints
-    sps.copy()
+    sps = sps.copy()
     if "location_id" not in sps.columns:
         raise KeyError(
             (
@@ -57,7 +57,7 @@ def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs
     if pre_filter:
         f = pre_filter_locations(sps, **pre_filter_kwargs)
     else:
-        f = pd.Series([True for _ in sps.index])
+        f = pd.Series(np.full(len(sps.index), True))
 
     if recipe == "FREQ":
         recipe_val = freq_recipe(sps[f], "home", "work")
@@ -115,7 +115,8 @@ def pre_filter_locations(
 
     Examples
     --------
-    >> sps = sps[ti.analysis.pre_filter_locations(sps)]
+    >> mask = ti.analysis.pre_filter_locations(sps)
+    >> sps = sps[mask]
     """
     assert sps.as_staypoints
     sps = sps.copy()
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -3,16 +3,16 @@ import pandas as pd
 import trackintel as ti
 
 
-def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs):
-    """Assign "home" and "work" activity label for each user with different recipes.
+def location_identifier(spts, method="FREQ", pre_filter=True, **pre_filter_kwargs):
+    """Assign "home" and "work" activity label for each user with different methods.
 
     Parameters
     ----------
-    sps : Geodataframe (as trackintel staypoints)
+    spts : Geodataframe (as trackintel staypoints)
         Staypoints with column "location_id".
 
-    recipe : {'FREQ'}, default "FREQ"
-        Choose which recipe to use.
+    method : {'FREQ'}, default "FREQ"
+        Choose which method to use.
 
         - FREQ: "Generate a activity label per user by assigning the most visited location the label "home"
           and the second most visited location the label "work". The remaining locations get no label.
@@ -31,7 +31,7 @@ def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs
 
     Note
     ----
-    The recipe are adapted from [1]. The original algorithms count the distinct hours at a
+    The methods are adapted from [1]. The original algorithms count the distinct hours at a
     location as the home location is derived from geo-tagged tweets. We directly sum the time
     spent at a location as our data model includes that.
 
@@ -44,33 +44,33 @@ def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs
 
     Examples
     --------
-    >>> ti.analysis.location_identifier(sps, pre_filter=True, recipe="FREQ")
+    >>> ti.analysis.location_identifier(spts, pre_filter=True, method="FREQ")
     """
-    assert sps.as_staypoints
-    sps = sps.copy()
-    if "location_id" not in sps.columns:
+    assert spts.as_staypoints
+    spts = spts.copy()
+    if "location_id" not in spts.columns:
         raise KeyError(
             (
                 "To derive activity labels the GeoDataFrame (as trackintel staypoints) must have a column "
-                f"named 'location_id' but it has [{', '.join(sps.columns)}]"
+                f"named 'location_id' but it has [{', '.join(spts.columns)}]"
             )
         )
     if pre_filter:
-        f = pre_filter_locations(sps, **pre_filter_kwargs)
+        f = pre_filter_locations(spts, **pre_filter_kwargs)
     else:
-        f = pd.Series(np.full(len(sps.index), True))
+        f = pd.Series(np.full(len(spts.index), True))
 
-    if recipe == "FREQ":
-        recipe_val = freq_recipe(sps[f], "home", "work")
+    if method == "FREQ":
+        method_val = freq_method(spts[f], "home", "work")
     else:
-        raise ValueError(f"Recipe {recipe} does not exist.")
+        raise ValueError(f"Method {method} does not exist.")
 
-    sps.loc[f, "activity_label"] = recipe_val["activity_label"]
-    return sps
+    spts.loc[f, "activity_label"] = method_val["activity_label"]
+    return spts
 
 
 def pre_filter_locations(
-    sps,
+    spts,
     agg_level="user",
     thresh_sp=10,
     thresh_loc=10,
@@ -84,7 +84,7 @@ def pre_filter_locations(
 
     Parameters
     ----------
-    sps : GeoDataFrame (as trackintel staypoints)
+    spts : GeoDataFrame (as trackintel staypoints)
         Staypoints with the column "location_id".
 
     agg_level: {"user", "dataset"}, default "user"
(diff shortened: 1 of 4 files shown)


trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+    user_loc = user["location_id"] >= thresh_min_loc
+    user_filter_agg = user_sp & user_loc
+    user_filter_agg.rename("user_filter", inplace=True)  # rename for merging
+    user_filter = pd.merge(sps["user_id"], user_filter_agg, left_on="user_id", right_index=True)["user_filter"]
+
+    # filtering locations
+    sps["duration"] = sps["finished_at"] - sps["started_at"]
+    if agg_level == "user":
+        groupby_loc = ["user_id", "location_id"]
+    elif agg_level == "dataset":
```
hong2223, line 124: Is there a reason why we want to distinguish between these two types here (user and dataset)? 
The generation algorithm did this because sometimes locations may have meanings for several users. Here, `location_identifier` is user-specific (e.g., home) and we need to filter on the individual level? 

I think operationally it would be okay if we just include `user` agg_level. In the generation function, the location are assigned to each user even if `agg_level == "dataset"`

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+        groupby_loc = ["location_id"]
+    else:
+        raise ValueError(f"Unknown agg_level '{agg_level}' use instead {{'user', 'dataset'}}.")
+    loc = sps.groupby(groupby_loc).agg({"started_at": [min, "count"], "finished_at": max, "duration": sum})
+    loc.columns = loc.columns.droplevel(0)  # remove possible multi-index
+    loc.rename(columns={"min": "started_at", "max": "finished_at", "sum": "duration"}, inplace=True)
+    loc["period"] = loc["finished_at"] - loc["started_at"]
```
hong2223, line 131: If I understand correctly `period` considers the time span of visits and `duration` considers only the sum of stay time?
Sometimes this gets a bit confusing. Maybe more comments and description inside the code is needed.

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+
+def pre_filter_locations(
+    sps,
+    agg_level="user",
+    thresh_min_sp=10,
+    thresh_min_loc=10,
+    thresh_sp_at_loc=10,
+    thresh_loc_time=1,
+    thresh_loc_period=pd.Timedelta("5h"),
+):
```
hong2223, line 72: IMO, this `pre_filter` is quite specific to the `freq_recipe` function that we use, and sometimes too strict? would it be possible to **not** using some of the filters? for example, if I do not want a `thresh_min_sp` filter, what should I input here?

I realize this is the design in the original article and might be too minor for this first pull request

EDIT: I notice in the test that we can disable the filters by setting them to 0. Maybe state this in the docstring

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+    if "location_id" not in sps.columns:
+        raise KeyError(
+            (
+                "To derive location activities the GeoDataFrame (as trackintel staypoints)must have a column "
+                f"named 'location_id' but it has [{', '.join(sps.columns)}]"
+            )
+        )
+    # then hand it to to the filter function if necessary.
+    if pre_filter:
+        f = pre_filter_locations()
```
hong2223, line 56: `pre_filter_locations()` accepts quite some parameters, do we need to give users the freedom to alter them?

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+        Staypoints with column "location_id".
+
+    pre_filter : bool, default True
+        Prefiltering the staypoints to exclude locations with not enough data.
+        The filter function can also be accessed via `pre_filter`.
+
+    recipe : {'FREQ'}, default "FREQ"
+        Choose which recipe to use.
+        - FREQ: Select the most visited location
+        - PASS: just for decoration.
```
hong2223, line 20: delete this

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+    loc_filter = pd.merge(sps[groupby_loc], loc_filter_agg.reset_index(), on=groupby_loc, how="left")["loc_filter"]
+
+    total_filter = user_filter & loc_filter
+
+    return total_filter
+
+
+def freq_recipe(sps, *labels):
+    """Generate a location activity per user by assigning the most visited location the label "home"
+    and the second most visited location the label "work". The remaining locations get no label.
```
hong2223, line 150: This description is very nice and should be incorporated into the recipe of the main function.

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+    group_agg = group.groupby("location_id").agg({"duration": sum})
+    group_agg["activity_label"] = _freq_assign(group_agg["duration"], *labels)
+    group_merge = pd.merge(
+        group["location_id"], group_agg["activity_label"], how="left", left_on="location_id", right_index=True
+    )
+    return group_merge["activity_label"]
+
+
+def _freq_assign(duration, *labels):
+    """Assign labels to array with longest duration.
```
hong2223, line 200: Is this `longest duration` or `k longest duration` (here 1st is home and 2nd is work)?

trackintel/analysis/location_identification.py
hong2223, line 150: In theory, this is not restricted to home and work but can accept arbitrary labels and length? e.g., I can define my labels as ["1st", "2nd", "3rd"...]?

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+        raise KeyError(
+            (
+                "To derive location activities the GeoDataFrame (as trackintel staypoints)must have a column "
+                f"named 'location_id' but it has [{', '.join(sps.columns)}]"
+            )
+        )
+    # then hand it to to the filter function if necessary.
+    if pre_filter:
+        f = pre_filter_locations()
+        sps = sps[f]
```
hong2223, line 57: After this filtering, the stps that we return is only a subset of the original stps? 
What do you think would be better?
- return filtered stps (like now)
- return original stps with an additional column "activity_label", the filtered records received pd.NA.

This is mostly a design issue

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+
+    recipe : {'FREQ'}, default "FREQ"
+        Choose which recipe to use.
+        - FREQ: Select the most visited location
+        - PASS: just for decoration.
+
+    Returns
+    -------
+    Geodataframe (as trackintel staypoints)
+        With additional column assigning one of three location identies {'home', 'work', None}.
```
hong2223, line 25: might be good to state the name of the additional column. I think it's `activity_label`?

hong2223 (changes requested):
Many thanks for this **BIG** and complete pull request! Overall the framework is quite nice.

I only have some design issue comments that I think is worth discussing before merging this into trackintel. You can check them on each line

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,214 @@
...
+    thresh_min_sp : int, default 10
+        Minimum staypoints a user must have to be included.
+
+    thresh_min_loc : int, default 10
+        Minimum locations a user must have to be included.
+
+    thresh_sp_at_loc : int, default 10
+        Minimum number of staypoints at a location must have to be included.
+
+    thresh_loc_time : int, default 1
```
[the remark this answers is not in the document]
bifbof: Yes, I was a bit undecided what to choose, a good range for `thresh_loc_period` could vary from hours to days. So first I chose a fixed integer unit for `thresh_loc_time` and then which I couldn`t use for the period.
I will change both to the same interface so that it is either a string for conversion to a timedelta, e.g. `5H` or a pd.Timedelta.

trackintel/analysis/location_identification.py
bifbof: Added a comment to clarify that.

trackintel/analysis/location_identification.py
bifbof: Added `**pre_filter_kwargs` in function definition to enable direct managment.

trackintel/analysis/location_identification.py
bifbof: I think the second variant is better as staypoints don't disappear without a notice.

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,229 @@
+import numpy as np
+import pandas as pd
+import trackintel as ti
+
+
+def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs):
```
hong2223, line 6: would be great if you could also add `location_identifier` and/or `pre_filter_locations` to the documentation.

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,229 @@
...
+    ‘Identifying Home Locations in Human Mobility Data: An Open-Source R Package for Comparison and Reproducibility’.
+    International Journal of Geographical Information Science 0 (0): 1–24.
+    https://doi.org/10.1080/13658816.2021.1887489.
+
+    Examples
+    --------
+    >>> ti.analysis.location_identifier(sps, pre_filter=True, recipe="FREQ")
+    """
+    assert sps.as_staypoints
+    sps.copy()
```
hong2223, line 49: I think you need to write `sps = sps.copy()`
`.copy()` is not an inplace operation

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,229 @@
...
+        raise KeyError(
+            (
+                "To derive activity labels the GeoDataFrame (as trackintel staypoints) must have a column "
+                f"named 'location_id' but it has [{', '.join(sps.columns)}]"
+            )
+        )
+    if pre_filter:
+        f = pre_filter_locations(sps, **pre_filter_kwargs)
+    else:
+        f = pd.Series([True for _ in sps.index])
```
hong2223, line 60: Maybe too nitpicking: can we replace this `forloop` with:
`pd.Series(np.full(len(sps.index), True))`

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,229 @@
...
+    -------
+    pd.Series
+        Boolean series containing the filter as a mask.
+
+    Examples
+    --------
+    >> sps = sps[ti.analysis.pre_filter_locations(sps)]
+    """
+    assert sps.as_staypoints
+    sps = sps.copy()
```
hong2223, line 121: we do not need this now, as spt is already copied in the main function?

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,229 @@
...
+        If str must be parsable by pd.to_timedelta.
+
+    Returns
+    -------
+    pd.Series
+        Boolean series containing the filter as a mask.
+
+    Examples
+    --------
+    >> sps = sps[ti.analysis.pre_filter_locations(sps)]
```
hong2223, line 118: might be better to write this in two lines

hong2223 (approved):
The test file looks great and almost covered everything!
I left some minor comments. It would be great to add this function to the documentation.

trackintel/analysis/location_identification.py
bifbof: Yes, but it should be possible to access the function independently. Maybe that's a bit too much.

trackintel/analysis/location_identification.py
bifbof: I have tried it, but I am not quite sure if it is correct. I just copied the code from the older docs.

abcnishant007:
@hong2223  
once the first reviewer has completed the review, what are the action items for the second/third reviewer? In case of disagreement, I see options to un-resolve the discussion and suggest changes. 
But, if the second reviewer agrees to the comments and discussions from the first reviewer, what is the action item in that case?

henrymartin1:
> 
> 
> @hong2223
> once the first reviewer has completed the review, what are the action items for the second/third reviewer? In case of disagreement, I see options to un-resolve the discussion and suggest changes.

I'd say you create your review independently and then everybody is free to reply to your requests and discuss them in case of disagreement.

trackintel/analysis/location_identification.py
hong2223: I checked and added a commit to fix the warnings in the docstring. You can test yourself by referring to our [guide ](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#documentation).

henrymartin1 (approved):
Great work! I only had some comments that probably don't really fit in this pull requests and I opened new issues for them.

trackintel/analysis/location_identification.py
```diff
@@ -0,0 +1,231 @@
...
+
+def location_identifier(sps, recipe="FREQ", pre_filter=True, **pre_filter_kwargs):
+    """Assign "home" and "work" activity label for each user with different recipes.
+
+    Parameters
+    ----------
+    sps : Geodataframe (as trackintel staypoints)
+        Staypoints with column "location_id".
+
+    recipe : {'FREQ'}, default "FREQ"
```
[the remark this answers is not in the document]
bifbof: The paper we get it from called it that way.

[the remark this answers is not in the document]
bifbof: We agreed upon changing it to "method".

trackintel/analysis/location_identification.py
```diff
@@ -143,33 +143,57 @@ def pre_filter_locations(
...
+        if "duration" not in group.columns:
+            group["duration"] = group["finished_at"] - group["started_at"]
+        # pandas keeps inner order or groups
+        sps.loc[sps["user_id"] == name, "activity_label"] = _freq_transform(group, *labels)
+    return sps
 
 
-def _freq_transform(duration, *labels):
+def _freq_transform(group, *labels):
     """Transform function that assigns the longest visited data point the labels in order.
```
abcnishant007, line 197: possibly confusing statement: _assigns the longest visited data point the labels in order_

abcnishant007 (changes requested):
@hong2223 during the review, are we allowed to request changes for trivial issues like typos/confusing inline comments?

trackintel/analysis/location_identification.py
```diff
@@ -1,41 +1,74 @@
-def location_identifier(sps, pre_filter=True, recipe="FREQ"):
-    """
+def location_identifier(sps, pre_filter=True, receipt="FREQ"):
+    """Finde home and work/location for each user with different receipts.
```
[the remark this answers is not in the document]
bifbof: Was already changed in a earlier commit.

trackintel/analysis/location_identification.py
bifbof: Tried to improve the docstring.

abcnishant007 (approved):

PR #217, opened 2021-05-19
BUG: extra columns #203

### LINKED ISSUE #203: BUG: generate_staypoints and generate_trips
if the input pfs already has a column `staypoint_id`, the following error will occur on Line 144

`ValueError: columns overlap but no suffix specified: Index(['staypoint_id'], dtype='object')`

abcnishant007:
closes #203

### DIFF
--- a/tests/preprocessing/test_positionfixes.py
+++ b/tests/preprocessing/test_positionfixes.py
@@ -30,10 +30,10 @@ class TestGenerate_staypoints:
         # not cause any problems in the second run
         pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife"))
         pfs_run_1, stps = pfs.as_positionfixes.generate_staypoints(
-            method="sliding", dist_threshold=0, time_threshold=0, include_last=True
+            method="sliding", dist_threshold=100, time_threshold=5.0, include_last=True
         )
         pfs_run_2, stps = pfs_run_1.as_positionfixes.generate_staypoints(
-            method="sliding", dist_threshold=0, time_threshold=0, include_last=True
+            method="sliding", dist_threshold=100, time_threshold=5.0, include_last=True
         )
         assert set(pfs_run_1.columns) == set(pfs_run_2.columns)
 
--- a/tests/preprocessing/test_positionfixes.py
+++ b/tests/preprocessing/test_positionfixes.py
@@ -29,10 +29,10 @@ class TestGenerate_staypoints:
         # we run generate_staypoints twice in order to check that the extra column(tripleg_id) does
         # not cause any problems in the second run
         pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife"))
-        pfs_run_1, stps = pfs.as_positionfixes.generate_staypoints(
+        pfs_run_1, _ = pfs.as_positionfixes.generate_staypoints(
             method="sliding", dist_threshold=100, time_threshold=5.0, include_last=True
         )
-        pfs_run_2, stps = pfs_run_1.as_positionfixes.generate_staypoints(
+        pfs_run_2, _ = pfs_run_1.as_positionfixes.generate_staypoints(
             method="sliding", dist_threshold=100, time_threshold=5.0, include_last=True
         )
         assert set(pfs_run_1.columns) == set(pfs_run_2.columns)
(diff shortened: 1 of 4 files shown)


abcnishant007:
@hong2223 
I added two tests to make sure that the actual bug does not exist anymore. 
do you have any suggestions why my the code coverage decreased after my changes?

abcnishant007:
@hong2223 ignore the previous comment about coverage. Actually the coverage decrease was probably because of some older results being displayed. It is fine now. Please feel free to review.

trackintel/preprocessing/positionfixes.py
```diff
@@ -214,6 +218,10 @@ def generate_triplegs(pfs_input, stps_input, method="between_staypoints", gap_th
     # copy the original pfs for adding 'staypoint_id' column
     pfs = pfs_input.copy()
 
+    # if the positionfixes already have a column "tripleg_id", we drop it
+    if "tripleg_id" in pfs:
```
hong2223, line 222: The original bug probably would not occur in this function: We wrote `pfs["tripleg_id"] = pd.NA` in Line [264](https://github.com/mie-lab/trackintel/blob/ddd0a9df13d5e4e0ffe962ccf261655e1c814aa4/trackintel/preprocessing/positionfixes.py#L264).

But it won't hurt to add this either :)

tests/preprocessing/test_positionfixes.py
```diff
@@ -23,6 +23,20 @@ def geolife_pfs_stps_long():
...
     """Tests for generate_staypoints() method."""
 
+    def test_duplicate_columns(self):
+        """Test if running the function twice, the generated column does not yield exception in join statement"""
+
+        # we run generate_staypoints twice in order to check that the extra column(tripleg_id) does
+        # not cause any problems in the second run
+        pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife"))
+        pfs_run_1, stps = pfs.as_positionfixes.generate_staypoints(
+            method="sliding", dist_threshold=0, time_threshold=0, include_last=True
```
hong2223, line 33: Here may be using a "normal" parameter set would be better. Like ` dist_threshold=100, time_threshold=5.0`

tests/preprocessing/test_positionfixes.py
```diff
@@ -23,6 +23,20 @@ def geolife_pfs_stps_long():
 class TestGenerate_staypoints:
     """Tests for generate_staypoints() method."""
 
+    def test_duplicate_columns(self):
+        """Test if running the function twice, the generated column does not yield exception in join statement"""
+
+        # we run generate_staypoints twice in order to check that the extra column(tripleg_id) does
+        # not cause any problems in the second run
+        pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife"))
+        pfs_run_1, stps = pfs.as_positionfixes.generate_staypoints(
```
hong2223, line 32: `stps` is not used, consider replacing it with `_`. Same in Line 35

tests/preprocessing/test_positionfixes.py
```diff
@@ -123,6 +137,17 @@ def test_temporal(self):
...
 
+    def test_duplicate_columns(self, geolife_pfs_stps_long):
+        """Test if running the function twice, the generated column does not yield exception in join statement"""
+
+        # we run generate_triplegs twice in order to check that the extra column (tripleg_id) does
+        # not cause any problems in the second run
+        pfs, stps = geolife_pfs_stps_long
+        # case 1
+        pfs_run_1, tpls = pfs.as_positionfixes.generate_triplegs(stps, method="between_staypoints")
+        pfs_run_2, tpls = pfs_run_1.as_positionfixes.generate_triplegs(stps, method="between_staypoints")
```
hong2223, line 148: Line 147-148, `tpls` is not used, replace with `_`

tests/preprocessing/test_positionfixes.py
```diff
@@ -123,6 +137,17 @@ def test_temporal(self):
 class TestGenerate_triplegs:
     """Tests for generate_triplegs() method."""
 
+    def test_duplicate_columns(self, geolife_pfs_stps_long):
+        """Test if running the function twice, the generated column does not yield exception in join statement"""
+
+        # we run generate_triplegs twice in order to check that the extra column (tripleg_id) does
+        # not cause any problems in the second run
+        pfs, stps = geolife_pfs_stps_long
+        # case 1
```
hong2223, line 146: what is case 1?

hong2223 (changes requested):
Thanks for the fix. 

Issue #203 states the bug is also observed in the `generate_trips` function, would be good if you can add the fix and a test function in that function as well.

tests/preprocessing/test_positionfixes.py
abcnishant007: done

tests/preprocessing/test_positionfixes.py
abcnishant007: done

tests/preprocessing/test_positionfixes.py
abcnishant007: it was a typo, removed

tests/preprocessing/test_positionfixes.py
abcnishant007: done

trackintel/preprocessing/positionfixes.py
abcnishant007: noted :) I have retained this

hong2223:
> @hong2223 once the contributor has made the requested changes according to the review, is the contributor supposed to click on "Resolve conversation" for each sub-comment? I just wrote "done" in each, not sure if that is the right way 😆

As you wish, we are flexible on this. I will now merge this PR, great work!

PR #232, opened 2021-05-31
ENH: Change for loop to groupby

### LINKED ISSUE #180: ENH/PERF for generate_locations
The performance of per-user location generation can be enhanced using .groupby() (Line 81-99). Pay attention to the unique labels across users (Line 92-96). 
An additional parameter `print_progress` can be added to monitor the per-user progress.
See `generate_trips` function for examples.

abcnishant007:
closes #180

### DIFF
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -1,14 +1,22 @@
 from math import radians
 
 import numpy as np
+import geopandas as gpd
 from shapely.geometry import Point
 from sklearn.cluster import DBSCAN
+from tqdm import tqdm
 
 from trackintel.geogr.distances import meters_to_decimal_degrees
 
 
 def generate_locations(
-    staypoints, method="dbscan", epsilon=100, num_samples=1, distance_metric="haversine", agg_level="user"
+    staypoints,
+    method="dbscan",
+    epsilon=100,
+    num_samples=1,
+    distance_metric="haversine",
+    agg_level="user",
+    print_progress=False,
 ):
     """
     Generate locations from the staypoints.
@@ -36,9 +44,11 @@ def generate_locations(
 
     agg_level: {'user','dataset'}
         The level of aggregation when generating locations:
-
         - 'user'      : locations are generated independently per-user.
         - 'dataset'   : shared locations are generated for all users.
+
+    print_progress : bool, default False
+        If print_progress is True, the progress bar is displayed
 
     Returns
     -------
@@ -59,6 +69,8 @@ def generate_locations(
 
     # initialize the return GeoDataFrames
     ret_stps = staypoints.copy()
+    ret_stps = ret_stps.sort_values("started_at")
+    geo_col = ret_stps.geometry.name
 
     if method == "dbscan":
 
@@ -72,27 +84,34 @@ def generate_locations(
             db = DBSCAN(eps=epsilon, min_samples=num_samples, algorithm="ball_tree", metric=distance_metric)
 
         if agg_level == "user":
-            location_id_counter = 0
-            # TODO: change into groupby
-            for user_id_this in ret_stps["user_id"].unique():
-                # Slice staypoints array by user. This is not a copy!
-                user_staypoints = ret_stps[ret_stps["user_id"] == user_id_this]
 
-                if distance_metric == "haversine":
-                    # the input is converted to list of (lat, lon) tuples in radians unit
-                    p = np.array([[radians(g.y), radians(g.x)] for g in user_staypoints.geometry])
-                else:
-                    p = np.array([[g.x, g.y] for g in user_staypoints.geometry])
-                labels = db.fit_predict(p)
+            # we create a new column: 'serial_num' where we assign serial numbers to each user id
+            ret_stps = ret_stps.assign(serial_num=(ret_stps["user_id"]).astype("category").cat.codes)
+            if print_progress:
+                tqdm.pandas(desc="User location generation")
+                ret_stps = ret_stps.groupby("user_id", as_index=False).progress_apply(
+                    _generate_locations_per_user,
+                    geo_col=geo_col,
+                    distance_metric=distance_metric,
+                    db=db,
+                    num_total_stps=len(ret_stps),
+                )
+            else:
+                ret_stps = ret_stps.groupby("user_id", as_index=False).apply(
+                    _generate_locations_per_user,
+                    geo_col=geo_col,
+                    distance_metric=distance_metric,
+                    db=db,
+                    num_total_stps=len(ret_stps),
+                )
 
-                # enforce unique lables across all users without changing noise labels
-                max_label = np.max(labels)
-                labels[labels != -1] = labels[labels != -1] + location_id_counter
-                if max_label > -1:
-                    location_id_counter = location_id_counter + max_label + 1
+            ret_stps = ret_stps.drop(columns={"serial_num"})
 
-                # add staypoint - location matching to original staypoints
-                ret_stps.loc[user_staypoints.index, "location_id"] = labels
+            # we remove the holes in the staypoint ids to ensure sequential location ids
+            noise_labels = ret_stps["location_id"] == -1
+            ret_stps = ret_stps.assign(location_id=(ret_stps["location_id"]).astype("category").cat.codes)
+            ret_stps.loc[noise_labels, "location_id"] = -1
+
         else:
             if distance_metric == "haversine":
                 # the input is converted to list of (lat, lon) tuples in radians unit
@@ -171,3 +190,26 @@ def generate_locations(
     ret_loc["user_id"] = ret_loc["user_id"].astype(ret_stps["user_id"].dtype)
 
     return ret_stps, ret_loc
+
+
+def _generate_locations_per_user(user_staypoints, distance_metric, db, geo_col, num_total_stps):
+    """function called after groupby: should only contain records of one user; see generate_locations() function for parameter meaning."""
+
+    # ensuring unique labels for each user, by reserving a new block of ids (size of block = num_total_stps)
+    offset = user_staypoints["serial_num"].iloc[0] * num_total_stps
+
+    if distance_metric == "haversine":
+        # the input is converted to list of (lat, lon) tuples in radians unit
+        p = np.array([[radians(q.y), radians(q.x)] for q in (user_staypoints[geo_col])])
+    else:
+        p = np.array([[q.x, q.y] for q in (user_staypoints[geo_col])])
+    labels = db.fit_predict(p)
+
+    # enforce unique lables across all users without changing noise labels
+    labels[labels != -1] = labels[labels != -1] + offset
+
+    # add staypoint - location matching to original staypoints
+    user_staypoints["location_id"] = labels
+    user_staypoints = gpd.GeoDataFrame(user_staypoints, geometry=geo_col)
+
+    return user_staypoints
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -2,6 +2,7 @@ from math import radians
 
 import numpy as np
 import geopandas as gpd
+import pandas as pd
 from shapely.geometry import Point
 from sklearn.cluster import DBSCAN
 from tqdm import tqdm
@@ -69,7 +70,7 @@ def generate_locations(
 
     # initialize the return GeoDataFrames
     ret_stps = staypoints.copy()
-    ret_stps = ret_stps.sort_values("started_at")
+    ret_stps = ret_stps.sort_values(["user_id", "started_at"])
     geo_col = ret_stps.geometry.name
 
     if method == "dbscan":
@@ -84,9 +85,6 @@ def generate_locations(
             db = DBSCAN(eps=epsilon, min_samples=num_samples, algorithm="ball_tree", metric=distance_metric)
 
         if agg_level == "user":
-
-            # we create a new column: 'serial_num' where we assign serial numbers to each user id
-            ret_stps = ret_stps.assign(serial_num=(ret_stps["user_id"]).astype("category").cat.codes)
             if print_progress:
                 tqdm.pandas(desc="User location generation")
                 ret_stps = ret_stps.groupby("user_id", as_index=False).progress_apply(
@@ -94,7 +92,6 @@ def generate_locations(
                     geo_col=geo_col,
                     distance_metric=distance_metric,
                     db=db,
-                    num_total_stps=len(ret_stps),
                 )
             else:
                 ret_stps = ret_stps.groupby("user_id", as_index=False).apply(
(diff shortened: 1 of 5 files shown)


trackintel/preprocessing/staypoints.py
```diff
@@ -169,3 +178,28 @@ def generate_locations(
...
+        p = np.array([[radians(q.y), radians(q.x)] for q in (user_staypoints[geo_col])])
+    else:
+        p = np.array([[q.x, q.y] for q in (user_staypoints[geo_col])])
+    labels = db.fit_predict(p)
+
+    # enforce unique lables across all users without changing noise labels
+    labels[labels != -1] = labels[labels != -1] + location_id_counter
+
+    # add staypoint - location matching to original staypoints
+    user_staypoints.loc[user_staypoints.index, "location_id"] = labels
```
hong2223, line 202: `user_staypoints.loc[user_staypoints.index, "location_id"]` is basically `user_staypoints["location_id"]`?

trackintel/preprocessing/staypoints.py
```diff
@@ -169,3 +178,28 @@ def generate_locations(
...
     return ret_stps, ret_loc
+
+
+def _generate_locations_per_user(df, location_id_counter, distance_metric, db, geo_col):
+    """function called after groupby: should only contain records of one user; see generate_locations() function for parameter meaning."""
+
+    user_staypoints = df
+
+    # ensuring unique labels: we assume that every pandas group must have a unique index for the first element of the group
+    location_id_counter += user_staypoints.index[0]
```
hong2223, line 189: Does this imply that the original staypoints need to have a monotonic increasing index starting from 0? As staypoints' index is user input, this might not be always the case...

A solution could be we imitate the structure of [generate_staypoints](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/positionfixes.py) function: 
the function here `_generate_locations_per_user` returns location with location-staypoint pairs (the inverse mapping of `labels`), and we reassign the staypoint-location mapping after the groupby operation. 
This also means we need to write the section `### create locations as grouped staypoints` (Line [108-118](https://github.com/mie-lab/trackintel/blob/4cf96573272bbdc8e475063e3edc1c36b7e44f69/trackintel/preprocessing/staypoints.py#L108)) into the groupby operation

hong2223 (changes requested):
Thanks for the work!
I am not sure if the index assigning part works as expected, please check my detailed comments

trackintel/preprocessing/staypoints.py
abcnishant007: Do you mean that if the input staypoints file (.csv) is shuffled, it should not affect the tests? 
I just tested with a shuffled staypoints input file. I get a crash at [this](https://github.com/mie-lab/trackintel/blob/4cf96573272bbdc8e475063e3edc1c36b7e44f69/tests/preprocessing/test_staypoints.py#L172) assertion. Need to fix that.

But the master also has the same problem, if I try to run with a shuffled staypoints.csv file, [the same](https://github.com/mie-lab/trackintel/blob/4cf96573272bbdc8e475063e3edc1c36b7e44f69/tests/preprocessing/test_staypoints.py#L172) assertion fails when running master

trackintel/preprocessing/staypoints.py
abcnishant007: yes, modified

trackintel/preprocessing/staypoints.py
```diff
@@ -169,3 +178,28 @@ def generate_locations(
     ret_loc["user_id"] = ret_loc["user_id"].astype(ret_stps["user_id"].dtype)
 
     return ret_stps, ret_loc
+
+
+def _generate_locations_per_user(df, location_id_counter, distance_metric, db, geo_col):
+    """function called after groupby: should only contain records of one user; see generate_locations() function for parameter meaning."""
+
+    user_staypoints = df
```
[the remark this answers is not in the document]
abcnishant007: modified

trackintel/preprocessing/staypoints.py
hong2223: I think shuffled staypoints can be reordered using `start_at`; we can add this line at the beginning of the function. What I am referring to is the index of staypoints: it could be from 0 to n-1, but could also be unordered or even like "1x32sa" (all it needs to be is unique, but not necessarily ordered)

abcnishant007: Added sort by `start_at` and ensured unique labels for a randomized staypoint csv with string indices as suggested ([randomised_stps_string_indices.csv](https://github.com/mie-lab/trackintel/files/6576064/index_randomised.csv)).

abcnishant007: 7901528 should address the residual concern for user_ids being strings. I have modified it to reserve new blocks of integers for every user (block size: total number of staypoints). Later in the main function, I ensure that the location ids are continuous and the _holes_ induced due to reserving entire blocks have been removed. (test file: 
[with_strings_for_user_id.csv](https://github.com/mie-lab/trackintel/files/6604470/with_strings_for_user_id.csv))

trackintel/preprocessing/staypoints.py
```diff
@@ -72,27 +84,34 @@ def generate_locations(
...
-                user_staypoints = ret_stps[ret_stps["user_id"] == user_id_this]
 
-                if distance_metric == "haversine":
-                    # the input is converted to list of (lat, lon) tuples in radians unit
-                    p = np.array([[radians(g.y), radians(g.x)] for g in user_staypoints.geometry])
-                else:
-                    p = np.array([[g.x, g.y] for g in user_staypoints.geometry])
-                labels = db.fit_predict(p)
+            # we create a new column: 'serial_num' where we assign serial numbers to each user id
+            ret_stps = ret_stps.assign(serial_num=(ret_stps["user_id"]).astype("category").cat.codes)
```
hong2223, line 89: I would suggest to change this line to `ret_stps["serial_num"] = ret_stps["user_id"].astype("category").cat.codes` for readability

trackintel/preprocessing/staypoints.py
```diff
@@ -72,27 +84,34 @@ def generate_locations(
...
 
-                # enforce unique lables across all users without changing noise labels
-                max_label = np.max(labels)
-                labels[labels != -1] = labels[labels != -1] + location_id_counter
-                if max_label > -1:
-                    location_id_counter = location_id_counter + max_label + 1
+            # we remove the holes in the staypoint ids to ensure sequential location ids
+            noise_labels = ret_stps["location_id"] == -1
+            ret_stps = ret_stps.assign(location_id=(ret_stps["location_id"]).astype("category").cat.codes)
+            ret_stps.loc[noise_labels, "location_id"] = -1
```
hong2223, line 113: I would suggest change line 112-113 into 
`ret_stps.loc[~noise_labels, "location_id"] = (ret_stps.loc[~noise_labels,"location_id"].astype("category").cat.codes)`
Otherwise the sequential location id will start from 1 (0 is the original -1 noise label, but reassigned to -1 in line 113)

trackintel/preprocessing/staypoints.py
```diff
@@ -72,27 +84,34 @@ def generate_locations(
...
+            else:
+                ret_stps = ret_stps.groupby("user_id", as_index=False).apply(
+                    _generate_locations_per_user,
+                    geo_col=geo_col,
+                    distance_metric=distance_metric,
+                    db=db,
+                    num_total_stps=len(ret_stps),
+                )
+
+            ret_stps = ret_stps.drop(columns={"serial_num"})
```
hong2223, line 108: For MOBIS dataset, the returned location id is `[-2146247740 -2146247739 -2146247738 ...  2147416528  2147416529  2147416530]`, this means an integer overflow. We can of course change the offset into int64, but I think this is not elegant and we will again face this issue when dealing with real big data (MOBIS "only" contains 4k users).

I could not think of a way to `reduce` offset while still ensure uniqueness. Otherwise, I can try to imitate the way we did for the other functions and add to this pull request.

hong2223 (changes requested):
Thanks for the change. For the SBB dataset it works but for MOBIS I observe an integer overflow for the `offset`. I think we cannot afford to use this kind of offset (see my comments below)

trackintel/preprocessing/staypoints.py
henrymartin1: Ah nice, good that you have found it now. Overflow errors are often very hard to detect later on. 
I had a look and maybe I have found a solution to this problem. 
The solution has two steps:
1. Delete [line 207](https://github.com/mie-lab/trackintel/pull/232/files#diff-e5f965436547ae93edf1ce55e8085c31f842c8a131e70edf674692b546205b9eR207) in `_generate_locations_per_user`
2. Add the following code after [line 108](https://github.com/mie-lab/trackintel/pull/232/files#diff-e5f965436547ae93edf1ce55e8085c31f842c8a131e70edf674692b546205b9eR108)
```python 
# sort so that the last location id of a user = max(location id)
ret_stps = ret_stps.sort_values(['user_id', 'location_id'])
# identify start positions of new user_ids
start_of_user_id = ret_stps['user_id'] != ret_stps['user_id'].shift(1)
# calculate the offset (= last location id of the previous user)
# multiplication is to mask all positions where no new user starts and addition is to have a +1 when a 
# new user starts
loc_id_offset = ret_stps['location_id'].shift(1) * start_of_user_id + start_of_user_id
# fill first nan with 0 and create the cumulative sum
loc_id_offset = loc_id_offset.fillna(0).cumsum()
ret_stps['location_id2'] = ret_stps['location_id'] + loc_id_offset
```
@abcnishant007 let me know if you can work with this.

trackintel/preprocessing/staypoints.py
abcnishant007: This line has been removed in the new version

trackintel/preprocessing/staypoints.py
abcnishant007: This line has been removed in the new version

trackintel/preprocessing/staypoints.py
abcnishant007: Thanks @henrymartin1, this works perfect. I have applied these changes, needed to apply a sort by "user_id" if handling shuffled rows. Apart from that, this works as is. Tested for 
- geolife with 100K lines (first 100K lines extracted from the geolife combined data sent by @hong2223 )
- "with_strings_for_user_id.csv" file as discussed previously in this chain

trackintel/preprocessing/staypoints.py
```diff
@@ -72,27 +84,40 @@ def generate_locations(
...
-                # Slice staypoints array by user. This is not a copy!
-                user_staypoints = ret_stps[ret_stps["user_id"] == user_id_this]
 
-                if distance_metric == "haversine":
-                    # the input is converted to list of (lat, lon) tuples in radians unit
-                    p = np.array([[radians(g.y), radians(g.x)] for g in user_staypoints.geometry])
-                else:
-                    p = np.array([[g.x, g.y] for g in user_staypoints.geometry])
-                labels = db.fit_predict(p)
+            # we create a new column: 'serial_num' where we assign serial numbers to each user id
```
hong2223, line 88: delete this

trackintel/preprocessing/staypoints.py
```diff
@@ -72,27 +84,40 @@ def generate_locations(
...
-                labels[labels != -1] = labels[labels != -1] + location_id_counter
-                if max_label > -1:
-                    location_id_counter = location_id_counter + max_label + 1
+            # calculate the offset (= last location id of the previous user)
+            # multiplication is to mask all positions where no new user starts and addition is to have a +1 when a
+            # new user starts
+            loc_id_offset = ret_stps["location_id"].shift(1) * start_of_user_id + start_of_user_id
+
+            # fill first nan with 0 and create the cumulative sum
+            loc_id_offset = loc_id_offset.fillna(0).cumsum()
```
hong2223, line 117: This solution works perfectly for solution without noise label (i.e., setting `num_samples=1`). I think we did not deal with the noise label where location_id == -1. Therefore calculating cumsum will create identical `location_id` when switching users (see image below). We need to mask location_id = -1 records. 

You can reproduce the error by setting num_samples to a bigger value, e.g., num_samples=5.

image

trackintel/preprocessing/staypoints.py
```diff
@@ -169,3 +194,20 @@ def generate_locations(
     ret_loc["user_id"] = ret_loc["user_id"].astype(ret_stps["user_id"].dtype)
 
     return ret_stps, ret_loc
+
+
+def _generate_locations_per_user(user_staypoints, distance_metric, db, geo_col):
+    """function called after groupby: should only contain records of one user; see generate_locations() function for parameter meaning."""
```
hong2223, line 200: This line is too long. We limit the line limit to 120, [here ](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#others) Black will not format comments

hong2223 (changes requested):
The solution works perfectly for num_samples=1, and the performance gain is really big! 
We still need to consider cases where num_samples>2 that contain -1 noise location_ids. I think we also need a new test function to cover this.

trackintel/preprocessing/staypoints.py
abcnishant007: @hong2223 , thanks for looking into this. I am having trouble reproducing this. I think I can exactly reproduce it if I get your test script:
1) can you share the test script and any new input file you used? 
2) can you confirm that you don't have this problem with master?

When I tried to change all `num_sampes` to 5, several of the test functions in `test_staypoints.py` crashes with the mater as well.

abcnishant007:
> The solution works perfectly for num_samples=1, and the performance gain is really big!
> We still need to consider cases where num_samples>2 that contain -1 noise location_ids. I think we also need a new test function to cover this.

that is really good to know! 
any rough estimates of the improvement?

hong2223:
> > The solution works perfectly for num_samples=1, and the performance gain is really big!
> > We still need to consider cases where num_samples>2 that contain -1 noise location_ids. I think we also need a new test function to cover this.
> 
> that is really good to know!
> any rough estimates of the improvement?

See #180

trackintel/preprocessing/staypoints.py
abcnishant007: After the latest fix, tested with this file [geolife_stps_10k](https://github.com/mie-lab/trackintel/files/6669303/geolife_stps_10k.csv) file. 

The fix is same as the comment [above](https://github.com/mie-lab/trackintel/pull/232#discussion_r651086632) by you. 
Also added a new [test](https://github.com/abcnishant007/trackintel/blob/0799e9e4e0a1711e062247e356b1c107f292965e/tests/preprocessing/test_staypoints.py#L140) to handle this:

trackintel/preprocessing/staypoints.py
abcnishant007: done

trackintel/preprocessing/staypoints.py
abcnishant007: done

hong2223:
Thanks for the fix. I slightly changed the test structure and added some docstrings to our old test files. 
I also added the line `ret_stps.sort_values(["user_id", "started_at"], inplace=True)` after your `pd.concat()` to ensure `-1` records are placed together with the other records. I will merge this once the tests passed

hong2223 (approved):

abcnishant007:
@hong2223 thanks for the change! 
+1 to the sorting at the end!

PR #234, opened 2021-06-07
CLN: A variety of smaller consistensy issues

### LINKED ISSUE #228: Some minor points I noticed while using trackintel
1. In the `read_geolife` doc it says that returning labels is optional while it always returns labels. [link](https://trackintel.readthedocs.io/en/latest/modules/io.html#trackintel.io.dataset_reader.read_geolife)
2. The printing of progress in the geolife dataset reader should be optional
3. It would be great if `geolife_add_modes_to_triplegs` would have an example available
4. I think the explanation of `gap_threshold ` in `generate_staypoints` is not very clear. We should add a section under notes that briefly explains that positionfixes with a time distance that is greater than `gap_threshold` are excluded from a staypoint.
5. I think the default value of `gap_threshold ` in `generate_staypoints` should be something useful (e.g., 15 minutes as suggested in the notes) or nan/None/Null/inf/False to make clear that it is not used by default.
6. I think the documentation style of writing dark blue on light blue for the function definition has very bad contrast e.g., [here](https://trackintel.readthedocs.io/en/latest/modules/preprocessing.html#trackintel.preprocessing.positionfixes.generate_triplegs)
7. The clarification `if ‘distance_metric’ is ‘haversine’ or ‘euclidean’, the unit is in meters.` (from `generate_locations`) should be copied to `generate_staypoints`.
8. I think that using preprocessing methods via the accessors is somehow poorly documented. If using the accessors is the standard way it might be useful to mention them in the docstring of the method so that users will know that they can use a specific method via the accessor. It is sometimes not clear to me (as a maintainer^^) which methods are available via the accessor. Having the accessor written out in the doc string would also help because I could then simply copy the code from the documentation.
9. I think the default value of 5 minutes as a time threshold for [create_activity_flag](https://trackintel.readthedocs.io/en/latest/modules/analysis.html#trackintel.analysis.label.create_activity_flag) should be increase to 15
10. The example in [trip generation](https://trackintel.readthedocs.io/en/latest/modules/preprocessing.html#trackintel.preprocessing.triplegs.generate_trips) should show the usage of the accessor
11. The documentation of the [geolife dataset reader](https://trackintel.readthedocs.io/en/latest/modules/io.html#trackintel.io.dataset_reader.read_geolife) should mention that we support reading of mode labels.
12. I think the category `Label` in the [documentation ](https://trackintel.readthedocs.io/en/latest/modules/analysis.html#label) should rather be called `Labeling`.
13. `geolife_add_modes_to_triplegs` is not in documentation ❗ 
14. The example in [generate_trips](https://trackintel.readthedocs.io/en/latest/modules/preprocessing.html#trackintel.preprocessing.triplegs.generate_trips) should show the full path of the `generate_trips` function. E.g., `trackintel.preprocessing.triplegs.generate_trips`
15. `generate_trips` is shown under preprocessing/triplegs but it is not available via a tripleg accessor. Maybe it should be moved to analysis?

henrymartin1:
fixes  #228

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* linguist-vendored
+*.py linguist-vendored=false
--- a/docs/modules/io.rst
+++ b/docs/modules/io.rst
@@ -87,4 +87,10 @@ Predefined dataset readers
 ==========================
 We also provide functionality to parse well-known datasets directly into the trackintel framework.
 
+Geolife
+-----------
+We support easy parsing of the Geolife dataset including available mode labels.
+
 .. autofunction:: trackintel.io.dataset_reader.read_geolife
+
+.. autofunction:: trackintel.io.dataset_reader.geolife_add_modes_to_triplegs
--- a/tests/analysis/test_label.py
+++ b/tests/analysis/test_label.py
@@ -1,6 +1,7 @@
 import os
-import pandas as pd
+
 import numpy as np
+import pandas as pd
 import pytest
 
 import trackintel as ti
@@ -18,7 +19,7 @@ class TestCreate_activity_flag:
         activity_true = stps_test["activity"].copy()
         stps_test["activity"] = False
 
-        stps_test = stps_test.as_staypoints.create_activity_flag()
+        stps_test = stps_test.as_staypoints.create_activity_flag(method="time_threshold", time_threshold=5.0)
 
         pd.testing.assert_series_equal(stps_test["activity"], activity_true)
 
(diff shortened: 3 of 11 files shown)


henrymartin1:
I tried to simplify the review by creating a commit for every point from #228 (I just noticed that the issue number in the commit names is wrong ;-) )

hong2223 (commented):
Nice fixes! I only want to add some descriptions in the geolife reading function. I will add them myself and merge this PR once the tests passed.

PR #236, opened 2021-06-07
BUG: fixed name for geometry column

### LINKED ISSUE #230: generate triplegs uses fixed name for geometry column
[here](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/positionfixes.py#L357)
Should be replaced with tpls.geometry.name

henrymartin1:
fixes #230 
I can not really wait for someone to grab it...

### DIFF
--- a/trackintel/preprocessing/positionfixes.py
+++ b/trackintel/preprocessing/positionfixes.py
@@ -1,10 +1,9 @@
+import datetime
+
 import geopandas as gpd
 import numpy as np
 import pandas as pd
 from shapely.geometry import LineString, Point
-from sklearn.cluster import DBSCAN
-
-import datetime
 from tqdm import tqdm
 
 from trackintel.geogr.distances import haversine_dist
@@ -354,7 +353,7 @@ def generate_triplegs(pfs_input, stps_input, method="between_staypoints", gap_th
         posfix_grouper = pfs.groupby("tripleg_id")
 
         tpls = posfix_grouper.agg(
-            {"user_id": ["mean"], "tracked_at": [min, max], "geom": list}
+            {"user_id": ["mean"], "tracked_at": [min, max], pfs.geometry.name: list}
         )  # could add a "number of pfs": can be any column "count"
 
         # prepare dataframe: Rename columns; read/set geometry/crs;


abcnishant007:
@hong2223 I see my previous review comment above was ignored; was it because it is already taken care of the other pull request #240 ? 
If yes, all good and I will delete it and this one. Otherwise, I need to look deeper and fix any hole in my understanding of the problem.

hong2223:
> @hong2223 I see my previous review comment above was ignored; was it because it is already taken care of the other pull request #240 ?
> If yes, all good and I will delete it and this one. Otherwise, I need to look deeper and fix any hole in my understanding of the problem.

I am sorry, I did not see any review comment. It is unlikely that #240 solves the problem, as #240 and #236 are only changing the `geom` name. If you spot a problem, could you add an issue or document it somewhere so that we can look at it?

abcnishant007:
I see the problem. My comment is not visible if I open this tab using an incognito window! :(

abcnishant007:
Reproduced the comment below: 
Just below this line, there are other instances of using "geom" in place of geometry.name  for example [here](https://github.com/mie-lab/trackintel/blob/8a005014e395416f145042dc87a03808ff5741d2/trackintel/preprocessing/positionfixes.py#L363) and [here](https://github.com/mie-lab/trackintel/blob/8a005014e395416f145042dc87a03808ff5741d2/trackintel/preprocessing/positionfixes.py#L364). Why did we just change this one instance?

Screenshot 2021-06-10 at 3 59 45 PM

henrymartin1:
> 
> 
> Reproduced the comment below:
> Just below this line, there are other instances of using "geom" in place of geometry.name for example [here](https://github.com/mie-lab/trackintel/blob/8a005014e395416f145042dc87a03808ff5741d2/trackintel/preprocessing/positionfixes.py#L363) and [here](https://github.com/mie-lab/trackintel/blob/8a005014e395416f145042dc87a03808ff5741d2/trackintel/preprocessing/positionfixes.py#L364). Why did we just change this one instance?
> 
> Screenshot 2021-06-10 at 3 59 45 PM

These instances of "geom" are generated by us. They are a column of "tpls" which is generated by us in the `generate_triplegs` function, therefore we know that the geometry column has the name "geom". In the other instances it was data provided by the user

abcnishant007:
okay, got it! now it makes sense! thanks for the clarifying

PR #237, opened 2021-06-08
ENH: Added links to source codes

### LINKED ISSUE #226: Use a documentation template with link to source
I think it would be really great if there would be 
- a link to the source code in the documentation as it is done in [scikit learn ](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html#sklearn.neighbors.NearestNeighbors.radius_neighbors)
I did not look into it yet but theoretically we should be able to do this by only changing the sphinx template ;-)
- better contrast at some locations e.g. [here](https://trackintel.readthedocs.io/en/latest/modules/io.html#csv-file-import)

abcnishant007:
closes #226

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -2,9 +2,16 @@
   {% block footer %} {{ super() }}
 
   <style>
-    .section dl:nth-of-type(1) dt:nth-of-type(1){
-      color:#FF0000;
-    }
+
+html.writer-html4 .rst-content dl:not(.docutils)>dt,
+html.writer-html5 .rst-content dl[class]:not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple)>dt {
+	background: rgba(181, 213, 219, 0.54);
+	color: #004465;
+}
+.rst-content .viewcode-back,
+.rst-content .viewcode-link {
+	color: #0801fc;
+}
 
   </style>
 {% endblock %}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,0 +1,10 @@
+{% extends "!layout.html" %}
+  {% block footer %} {{ super() }}
+
+  <style>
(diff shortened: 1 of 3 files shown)


henrymartin1:
Is this still a draft on purpose?

abcnishant007:
@henrymartin1 yes this is draft on purpose, basically the color change part I did not manage to get what I wanted. I changed the colors of the background and the font in function signature, but the color of interleaving commas did not change. Maybe there is a simpler way to do that, hence waiting to figure out. 

The links to source codes was is exactly a one line change as you said :)

abcnishant007:
I added a `_template` folder inside docs and created a `layout.html` inside so that we can overwrite the theme colors. But, when I use the `dl` and `dt` according to the source (see image below), it somehow updates the second value instead of the first. Am I am missing something trivial here @hong2223? 

Screenshot 2021-06-13 at 12 52 24 AM

abcnishant007:
Finally, solved this using a weird workflow as described below, for future reference! 
1. Get the source html using inspect element by opening in google-chrome, 
2. Download the exact CSS being used finally (finally implying base css from the theme, any other overrides done by a previous developer etc..)
3. Copy paste it a [website](https://www.freeformatter.com/css-beautifier.html) to beautify it because the downloaded css consists of everything in just a single line
4. Get a screenshot of the page and identify colors using an online [hex color selector ](https://imagecolorpicker.com/en)
5. Use the hex color key to search for the exact element name from the CSS file
6. Use the selected element in layout.html to override the colors from the theme

abcnishant007:
The output looks as shown below. Please feel free to suggest any color change as we need to change just a single line in the `docs/_templates/layout.html`

Screenshot 2021-06-27 at 8 20 01 PM

henrymartin1 (approved):
Great work! I can not describe how much easier the link to the source makes my life ;-)

PR #246, opened 2021-06-10
BLD: delete imageio dependency

### LINKED ISSUE #225: Why is imageio in environment.yml?
I don't think that this is currently in use. Is it a dependency for something?

hong2223:
closes #225


henrymartin1 (approved):
looks good 👍

PR #252, opened 2021-06-17
ENH: Fix init.py and label module

### LINKED ISSUE #242: Remove wildcard imports from __init__.py files
We should remove Wildcard imports from the \_\_init\_\_.py files. [Example from trackintel](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/__init__.py)

Reasons:

[Official python documentation](https://docs.python.org/3/tutorial/modules.html)
>This imports all names except those beginning with an underscore (_). In most cases Python programmers do not use this facility since it introduces an unknown set of names into the interpreter, possibly hiding some things you have already defined.

>Note that in general the practice of importing * from a module or package is frowned upon, since it often causes poorly readable code. However, it is okay to use it to save typing in interactive sessions.

[PEP8](https://www.python.org/dev/peps/pep-0008/#imports)
>Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools. There is one defensible use case for a wildcard import, which is to republish an internal interface as part of a public API (for example, overwriting a pure Python implementation of an interface with the definitions from an optional accelerator module and exactly which definitions will be overwritten isn't known in advance).

[geopandas](https://geopandas.readthedocs.io/en/latest/community/contributing.html#contributing-to-geopandas)
> Imports should be grouped with standard library imports first, 3rd-party libraries next, and GeoPandas imports third. Within each grouping, imports should be alphabetized. *Always use absolute imports when possible*, and explicit relative imports for local imports when necessary in tests.

[geopandas \_\_init\_\_.py example](https://github.com/geopandas/geopandas/blob/master/geopandas/tools/__init__.py)

hong2223:
closes #242

- the imports of read functions (e.g., `from trackintel.io.file import read_positionfixes_csv`) in `trackintel/__init__.py` ensures we can directly call `ti.read_positionfixes_csv()` instead of going into the IO module (e.g., `ti.io.file.read_positionfixes_csv()`). I am doing this because many of our test functions use this, but of course, we can delete this shortcut
- the `__all__` ensures that if someone calls, e.g., `from ti.analysis import *`, still only the function defined in `__all__` would be imported.

Also, the module `label `is now renamed to `labelling`


docs/modules/analysis.rst
```diff
@@ -4,12 +4,12 @@ Analysis
 The analysis module provides a variety of functions to analyze the mobility
 datasets computed by trackintel. 
 
-Label
-=====
+Labelling
+=========
```
abcnishant007, line 8: good one! extending === 
that is some serious attention to detail! 🔎

trackintel/analysis/__init__.py
```diff
@@ -1,3 +1,22 @@
-from .tracking_quality import *
-from .label import *
-from .modal_split import *
+from .tracking_quality import temporal_tracking_quality
```
abcnishant007, line 1: a generic question: 
why do we need the . before every `import` path?

trackintel/io/__init__.py
```diff
@@ -29,3 +29,4 @@
 from .from_geopandas import read_trips_gpd
 
 from .dataset_reader import read_geolife
+from .dataset_reader import geolife_add_modes_to_triplegs
```
abcnishant007, line 32: I don't see this `geolife_add_modes_to_triplegs` being used anywhere in this list of changes. Wondering if it serves any purpose?

abcnishant007 (changes requested):
Mostly minor; Summary:
(1) About . in import 
(2) One potentially unused module being imported

trackintel/analysis/__init__.py
hong2223: maybe see the discussion [here](https://stackoverflow.com/questions/14132789/relative-imports-for-the-billionth-time). 
`.` is the symbol for relative import relative to the current init.py file

trackintel/io/__init__.py
hong2223: if we do not include a function in `init.py`, the function will not be able to call outside this module. e.g., using `from trackintel.io.dataset_reader import geolife_add_modes_to_triplegs` will raise an error.
So if we want a function to be called by the user ("outside" function), we need to add it to the corresponding `init.py` file

abcnishant007:
@hong2223 my concerns have been answered/explained. 
Should I merge this? Or feel free to merge from your end.

PR #254, opened 2021-06-18
ENH: gap_threshold for generate_stps

### LINKED ISSUE #245: Enhancements for generate_staypoints() function
Two points related to generate_staypoints() function
> I think the explanation of gap_threshold in generate_staypoints is not very clear. We should add a section under notes that briefly explains that positionfixes with a time distance that is greater than gap_threshold are excluded from a staypoint.

> I think the default value of gap_threshold in generate_staypoints should be something useful (e.g., 15 minutes as suggested in the notes) or nan/None/Null/inf/False to make clear that it is not used by default.

hong2223:
closes #245

`gap_threshold` now default to 15 minutes. If signal is missing larger than `gap_threshold`, no staypoint will be generated. An example is given below. The signal between 469 and 470 is missing for several hours; originally this will generate a staypoint with a long duration, but setting `gap_threshold=15` will ignore this staypoint.
image

Some of our old test functions only worked when setting `gap_threshold` large. 
We have a test function `test_temporal` to test the different combinations of `time_threshold` and `gap_threshold`

### DIFF
--- a/tests/preprocessing/test_positionfixes.py
+++ b/tests/preprocessing/test_positionfixes.py
@@ -110,28 +110,26 @@ class TestGenerate_staypoints:
 
         # the duration should be not longer than time_threshold
         time_threshold_ls = [3, 5, 10]
-        for time_threshold in time_threshold_ls:
-            _, stps = pfs_input.as_positionfixes.generate_staypoints(
-                method="sliding", dist_threshold=50, time_threshold=time_threshold
-            )
-
-            duration_stps_min = (stps["finished_at"] - stps["started_at"]).dt.total_seconds() / 60
-            # all durations should be longer than the time_threshold
-            assert (duration_stps_min > time_threshold).all()
 
         # the missing time should not exceed gap_threshold
-        gap_threshold_ls = [10, 15, 20]
-        for gap_threshold in gap_threshold_ls:
-            pfs, _ = pfs_input.as_positionfixes.generate_staypoints(
-                method="sliding", dist_threshold=50, time_threshold=time_threshold, gap_threshold=gap_threshold
-            )
-            # get the difference between pfs tracking time, and assign back to the previous pfs
-            pfs["diff"] = ((pfs["tracked_at"] - pfs["tracked_at"].shift(1)).dt.total_seconds() / 60).shift(-1)
-            # get the last pf of stps and check the gap size
-            pfs.dropna(subset=["staypoint_id"], inplace=True)
-            pfs.drop_duplicates(subset=["staypoint_id"], keep="last", inplace=True)
-            # all last pfs should be shorter than the gap_threshold
-            assert (pfs["diff"] < gap_threshold).all()
+        gap_threshold_ls = [5, 10, 15, 20]
+        for time_threshold in time_threshold_ls:
+            for gap_threshold in gap_threshold_ls:
+                pfs, stps = pfs_input.as_positionfixes.generate_staypoints(
+                    method="sliding", dist_threshold=50, time_threshold=time_threshold, gap_threshold=gap_threshold
+                )
+
+                # all durations should be longer than the time_threshold
+                duration_stps_min = (stps["finished_at"] - stps["started_at"]).dt.total_seconds() / 60
+                assert (duration_stps_min > time_threshold).all()
+
+                # all last pfs should be shorter than the gap_threshold
+                # get the difference between pfs tracking time, and assign back to the previous pfs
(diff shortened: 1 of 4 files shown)


tests/preprocessing/test_positionfixes.py
```diff
@@ -110,28 +110,26 @@ def test_temporal(self):
...
+                    method="sliding", dist_threshold=50, time_threshold=time_threshold, gap_threshold=gap_threshold
+                )
+
+                # all durations should be longer than the time_threshold
+                duration_stps_min = (stps["finished_at"] - stps["started_at"]).dt.total_seconds() / 60
+                assert (duration_stps_min > time_threshold).all()
+
+                # all last pfs should be shorter than the gap_threshold
+                # get the difference between pfs tracking time, and assign back to the previous pfs
+                pfs["diff"] = ((pfs["tracked_at"] - pfs["tracked_at"].shift(1)).dt.total_seconds() / 60).shift(-1)
```
henrymartin1, line 128: Is this really correct? You are getting the difference between the future next two positionfixes instead of between the current and the upcoming positionfix.
image

henrymartin1 (changes requested):
Looks good, I have just a question that you should look at before we merge it

trackintel/preprocessing/positionfixes.py
```diff
@@ -15,7 +15,7 @@ def generate_staypoints(
     distance_metric="haversine",
     dist_threshold=100,
     time_threshold=5.0,
-    gap_threshold=1e6,
+    gap_threshold=15.0,
```
abcnishant007, line 18: why such a drastic change of value? does it mean previous was wrong?

trackintel/preprocessing/positionfixes.py
```diff
@@ -413,7 +414,7 @@ def _generate_staypoints_sliding_user(
             # the duration of gap in the last two pfs
             gap_t = (pfs[curr]["tracked_at"] - pfs[curr - 1]["tracked_at"]).total_seconds()
 
-            # we want the spt to have long duration, but the gap of missing signal should not be too long
+            # we want the spt to have long duration, but the gap of two consecutive pfs should not be too long
```
abcnishant007, line 417: should we stick to the convention I have seen in other places for comments? like writing "positionfixes" in place of pfs, "staypoints" in place of stps etc..?

just wondering.. :)

abcnishant007 (changes requested):
Mostly minor. 
(1) about drastic change of gap_threshold
(2) comment related

I think @henrymartin1 's [concern](https://github.com/mie-lab/trackintel/pull/254/files#r658328195) is crucial

tests/preprocessing/test_positionfixes.py
hong2223: I did not understand your comment... if the last row is the diff, then I think it should be correct
`pfs["tracked_at"] - pfs["tracked_at"].shift(1)` gets the difference between `pfs[i+1]` and `pfs[i]`. Index is `pfs[i+1]`.
`pfs["tracked_at"] - pfs["tracked_at"].shift(1)).dt.total_seconds() / 60` transforms seconds to minutes
the last `.shift(-1)` transforms the index from `pfs[i+1]` to `pfs[i]` so that the diff is assigned to the beginning pfs

trackintel/preprocessing/positionfixes.py
hong2223: see #245. Previously the function was not "gap aware". This is more of a definition change than a bug.

trackintel/preprocessing/positionfixes.py
hong2223: I am not sure if we have this conversion, maybe need to agree on one
But here the comments have been changed

abcnishant007:
@hong2223 my concerns have been addressed/answered. 
How should I convery this? Does a comment here (such as this comment) suffice?

hong2223:
> @hong2223 my concerns have been addressed/answered.
> How should I convery this? Does a comment here (such as this comment) suffice?

The status of this PR is still "requested changes" from you, maybe you can try to approve the change somewhere?

abcnishant007 (approved):

abcnishant007:
> > @hong2223 my concerns have been addressed/answered.
> > How should I convery this? Does a comment here (such as this comment) suffice?
> 
> The status of this PR is still "requested changes" from you, maybe you can try to approve the change somewhere?

done!

tests/preprocessing/test_positionfixes.py
henrymartin1: err yes.... I should stop reviewing code at night ;-)

henrymartin1 (approved):
Nice work! Sorry for the delay

PR #255, opened 2021-06-18
ENH: correct dataframe check

### LINKED ISSUE #215: Enable trackintel datatype asserts via accessor
We often use the accessor in an assert statement 
e.g., 
```Python
assert spts.as_staypoints
```
However, I think at the moment this is not really correct. I think the input to an ```assert``` should be boolean and either be True or False. The accessors call the ```__init__``` method (e.g., [for staypoints](https://github.com/mie-lab/trackintel/blob/master/trackintel/model/staypoints.py#L39)) which calls ```self._validate()```. At the moment our ```_validate``` method does not return True or False, it simply fails internally (which kind of works in the end).
In general I don't think that this is a super serious problem but the following would not give the expected results:
```Python
assert spts.as_staypoints, "This assert failed, therefore I show my custom assert message"
```
In that case the custom assert message would not be shown as an internal assertion error is raised instead. 

To fix it, we could either stop writing 
```Python
assert spts.as_staypoints
```
but instead write
```Python
spts.as_staypoints
```
which would lead the same results but is less readable. Or we would add a wrapper function that returns a boolean. 
E.g.

```Python
assert spts.as_staypoints.is_valid
```` 

I think a great advantage of the first option is that it explains the user why the assert fails (as we have very descriptive internal assert messages)

### LINKED ISSUE #248: Wrong abbrevation in read_staypoints_postgis
`read_staypoints_postgis` uses `pfs` instead of `spts` (or `stps`) 

https://github.com/mie-lab/trackintel/blob/master/trackintel/io/postgis.py#L175

### PR DESCRIPTION
**Pull request by hong2223**
closes #248 
closes #215

### DIFF
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -46,7 +46,8 @@ def location_identifier(spts, method="FREQ", pre_filter=True, **pre_filter_kwarg
     --------
     >>> ti.analysis.location_identifier(spts, pre_filter=True, method="FREQ")
     """
-    assert spts.as_staypoints
+    # check the correctness
+    spts.as_staypoints
     spts = spts.copy()
     if "location_id" not in spts.columns:
         raise KeyError(
@@ -119,7 +120,8 @@ def pre_filter_locations(
     >> mask = ti.analysis.pre_filter_locations(spts)
     >> spts = spts[mask]
     """
-    assert spts.as_staypoints
+    # check the correctness
+    spts.as_staypoints
     spts = spts.copy()
     if isinstance(thresh_loc_time, str):
         thresh_loc_time = pd.to_timedelta(thresh_loc_time)
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -46,8 +46,9 @@ def location_identifier(spts, method="FREQ", pre_filter=True, **pre_filter_kwarg
     --------
     >>> ti.analysis.location_identifier(spts, pre_filter=True, method="FREQ")
     """
-    # check the correctness
+    # assert validity of staypoints
     spts.as_staypoints
+
     spts = spts.copy()
     if "location_id" not in spts.columns:
         raise KeyError(
@@ -120,8 +121,9 @@ def pre_filter_locations(
     >> mask = ti.analysis.pre_filter_locations(spts)
     >> spts = spts[mask]
     """
-    # check the correctness
+    # assert validity of staypoints
     spts.as_staypoints
+
     spts = spts.copy()
     if isinstance(thresh_loc_time, str):
         thresh_loc_time = pd.to_timedelta(thresh_loc_time)
--- a/trackintel/io/file.py
+++ b/trackintel/io/file.py
@@ -93,7 +93,8 @@ def read_positionfixes_csv(*args, columns=None, tz=None, index_col=object(), crs
     pfs = gpd.GeoDataFrame(df, geometry="geom")
     if crs:
         pfs.set_crs(crs, inplace=True)
-    assert pfs.as_positionfixes
+    # check the correctness
+    pfs.as_positionfixes
     return pfs
 
 
@@ -195,7 +196,9 @@ def read_triplegs_csv(*args, columns=None, tz=None, index_col=object(), crs=None
     tpls = gpd.GeoDataFrame(df, geometry="geom")
     if crs:
         tpls.set_crs(crs, inplace=True)
-    assert tpls.as_triplegs
+
+    # check the correctness
+    tpls.as_triplegs
     return tpls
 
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: trackintel/analysis/location_identification.py [hunk 1]
```diff
@@ -46,7 +46,8 @@ def location_identifier(spts, method="FREQ", pre_filter=True, **pre_filter_kwarg
     --------
     >>> ti.analysis.location_identifier(spts, pre_filter=True, method="FREQ")
     """
-    assert spts.as_staypoints
+    # check the correctness
```
henrymartin1, line 49: Hm... I think this comment is maybe not specific enough. In the ones below you wrote "check the correctness of the staypoints", I like that more because the reader immediately knows what is going on.
I think "correctness" is OK but I am not a big fan. I would prefer using "valid" 
Maybe "check if staypoints are valid" or "assert validity of staypoints"
(but as I said, I am OK with "correctness as long as it is changed to "correctness of staypoints" as you did below.

trackintel/io/from_geopandas.py
```diff
@@ -332,6 +337,6 @@ def read_tours_gpd(
     #     if not pd.api.types.is_datetime64tz_dtype(trs[col]):
     #         trs[col] = localize_timestamp(dt_series=trs[col], pytz_tzinfo=tz, col_name=col)
 
-    # assert trs.as_tours
+    # trs.as_tours
```
henrymartin1, line 340: wrong comment

trackintel/io/postgis.py
```diff
@@ -35,7 +35,8 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
         pfs = gpd.GeoDataFrame.from_postgis("SELECT * FROM %s" % table_name, conn, geom_col=geom_col, *args, **kwargs)
     finally:
         conn.close()
-    assert pfs.as_positionfixes
+    # check the correctness of the positionfix
```
henrymartin1, line 38: I would probably go for plural

trackintel/io/postgis.py
```diff
@@ -106,7 +107,8 @@ def read_triplegs_postgis(conn_string, table_name, geom_col="geom", *args, **kwa
         )
     finally:
         conn.close()
-    assert pfs.as_triplegs
+    # check the correctness of the tripleg
```
henrymartin1, line 110: I would probably go for plural

henrymartin1 (changes requested):
I have some minor (and a bit picky) comments. See the comments in the code for details.

- replace with more specific comments
- change singular to plural

trackintel/io/from_geopandas.py
```diff
@@ -271,7 +275,8 @@ def read_locations_gpd(gdf, user_id="user_id", center="center", mapper={}):
     locs = gdf.rename(columns=columns)
     locs = locs.set_geometry("center")
 
-    assert locs.as_locations
+    # check the correctness
```
abcnishant007, line 278: Same as henry's comment: check the correctnes of ......

trackintel/io/file.py
```diff
@@ -376,7 +381,9 @@ def read_locations_csv(*args, columns=None, index_col=object(), crs=None, **kwar
     locs = gpd.GeoDataFrame(df, geometry="center")
     if crs:
         locs.set_crs(crs, inplace=True)
-    assert locs.as_locations
+
+    # check the correctness
```
abcnishant007, line 385: Same as henry's comment: check the correctnes of ......

trackintel/io/file.py
```diff
@@ -195,7 +196,9 @@ def read_triplegs_csv(*args, columns=None, tz=None, index_col=object(), crs=None
     tpls = gpd.GeoDataFrame(df, geometry="geom")
     if crs:
         tpls.set_crs(crs, inplace=True)
-    assert tpls.as_triplegs
+
+    # check the correctness
```
abcnishant007, line 200: Same as henry's comment: check the correctnes of ......

trackintel/io/file.py
```diff
@@ -93,7 +93,8 @@ def read_positionfixes_csv(*args, columns=None, tz=None, index_col=object(), crs
     pfs = gpd.GeoDataFrame(df, geometry="geom")
     if crs:
         pfs.set_crs(crs, inplace=True)
-    assert pfs.as_positionfixes
+    # check the correctness
```
abcnishant007, line 96: Same as henry's comment: check the correctnes of ......

trackintel/analysis/location_identification.py
```diff
@@ -119,7 +120,8 @@ def pre_filter_locations(
     >> mask = ti.analysis.pre_filter_locations(spts)
     >> spts = spts[mask]
     """
-    assert spts.as_staypoints
+    # check the correctness
```
abcnishant007, line 123: Same as henry's comment: check the correctnes of ......

abcnishant007 (changes requested):
mostly minor comments
Similar to what @henrymartin1 highlighted, I marked all other locations as well

abcnishant007:
@hong2223 All concerns have been addressed. Please feel free to merge this.

Issue #243, opened 2021-06-10
all write_*_postgis function do not use the schema


henrymartin1:
E.g. [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/io/postgis.py#L147)

In general I think that these wrapper functions should allow setting all variables from geopandas' [to_postgis](https://geopandas.org/docs/reference/api/geopandas.GeoDataFrame.to_postgis.html#geopandas.GeoDataFrame.to_postgis) function.

bifbof:
In PR #264 I added the schema argument to each `to_postgis` function. 
Should we add the others in that PR as well? 
The missing arguments (index and dtype) except index_label are used anyway, we may need to make our reading_*_postgis functions more flexible to accommodate this change.

PR #264, opened 2021-07-07
ENH: Update read_*_postgis functions.

### LINKED ISSUE #211: Update for read postgis functions
There are two issues with the read postgis functions:

Problems
- SQL injections: 
The current implementation of the read postgis functions are very inflexible and are vulnerable to sql injections. e.g., `gpd.GeoDataFrame.from_postgis("SELECT * FROM %s" % table_name, conn, geom_col=geom_col, *args, **kwargs)`
- Indices are not stable at the moment. E.g., the behavior what to do with the index when reading from or writing to the database is not defined.

Solution:
- Use the `from_geopandas` functions as a wrapper to transform the result of `geopandas.from_postgis` into the trackintel format.
- Don't make assumptions on the SQL format (e.g., only take an SQL string as input)

### LINKED ISSUE #227: write_locations_postgis throws AttributeError for multiple geometries
Our locations can have two geometry columns (center + extent). The geopandas.to_postgis function does only support having a single geometry column and throws the following error when having locations with two geometry columns: `AttributeError: 'GeometryDtype' object has no attribute 'base'`
For this case, we might have to send the data to postgis manually as WKB via the pandas sql functions. It might be a good solution to fall back on the [initial code](https://github.com/mie-lab/trackintel/blob/f9b97e1a1b04726277798d0c9cb91e2bec8f762f/trackintel/io/postgis.py#L309) for that.

Other minor adjustments for the postgis IO functions:

- The default values for `if_exists` should not be "replace" in the `to_postgis` functions. (only true in accessors)
- It would be nice if the postgis io functions would accept connection strings, engines or connection objects instead of only connection strings.
- the examples of the postgis io functions should use the trackintel abbrevations e.g., `spts.as_staypoints.to_postgis(conn_string, table_name)` instead of `df.as_staypoints.to_postgis(conn_string, table_name)`
- The accessors for staypoints and triplegs are only accepting two arguments (conn string and table name)

bifbof:
closes #227 
closes #211
The following changes in this pull request.
- Decorator to handle not only connection strings, but also engines/connection objects
- WKT transformation for locations extent
- Updating accessor arguments (adding missing ones and `if_exists` to "fail".
- new argument `sql` instead of fix string.
- handing building GeoDataFrame to `read_*_gpd` functions

### DIFF
--- a/tests/io/test_postgis_gpd.py
+++ b/tests/io/test_postgis_gpd.py
@@ -7,7 +7,7 @@ import pandas as pd
 import pytest
 from geopandas.testing import assert_geodataframe_equal
 from pandas.testing import assert_frame_equal
-from shapely.geometry import LineString, Point
+from shapely.geometry import LineString, Point, Polygon
 import trackintel as ti
 
 
@@ -410,6 +410,28 @@ class TestLocations:
             locs.as_locations.to_postgis(conn_string, table)
             locs_db = ti.io.read_locations_postgis(conn_string, table, geom_col)
             assert_geodataframe_equal(locs, locs_db)
+        finally:
+            del_table(conn, table)
+
+    def test_write_extent(self, example_locations, conn_postgis):
+        """Test if extent geometry is handled correctly."""
+        conn_string, conn = conn_postgis
+        table = "locations"
+        coords = [[8.45, 47.6], [8.45, 47.4], [8.55, 47.4], [8.55, 47.6], [8.45, 47.6]]
+        extent = Polygon(coords)
+        example_locations["extent"] = extent
+        example_locations["extent"] = gpd.GeoSeries(example_locations["extent"])
+        try:
+            example_locations.as_locations.to_postgis(conn_string, table)
+            columns_db, dtypes = get_table_schema(conn, table)
+            columns = example_locations.columns.tolist() + [example_locations.index.name]
+            assert len(columns_db) == len(columns)
+            assert set(columns_db) == set(columns)
+            srid = _get_srid(example_locations)
+            geom_schema_center = f"geometry(Point,{srid})"
+            geom_schema_extent = f"geometry(Polygon,{srid})"
+            assert geom_schema_center in dtypes
+            assert geom_schema_extent in dtypes
         finally:
             del_table(conn, table)
 
--- a/tests/io/test_postgis_gpd.py
+++ b/tests/io/test_postgis_gpd.py
@@ -502,7 +502,7 @@ class Test_Handle_Con_String:
 
     def test_conn(self, conn_postgis):
         """Test handeling of connection input"""
-        _ , conn = conn_postgis
+        _, conn = conn_postgis
 
         @ti.io.postgis._handle_con_string
         def wrapped(con):
--- a/tests/io/test_postgis_gpd.py
+++ b/tests/io/test_postgis_gpd.py
@@ -58,11 +58,11 @@ def example_positionfixes():
     t3 = pd.Timestamp("1971-01-02 07:00:00", tz="utc")
 
     list_dict = [
-        {"user_id": 0, "tracked_at": t1, "geometry": p1},
-        {"user_id": 0, "tracked_at": t2, "geometry": p2},
-        {"user_id": 1, "tracked_at": t3, "geometry": p3},
+        {"user_id": 0, "tracked_at": t1, "geom": p1},
+        {"user_id": 0, "tracked_at": t2, "geom": p2},
+        {"user_id": 1, "tracked_at": t3, "geom": p3},
     ]
-    pfs = gpd.GeoDataFrame(data=list_dict, geometry="geometry", crs="EPSG:4326")
+    pfs = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
     pfs.index.name = "id"
     assert pfs.as_positionfixes
     return pfs
@@ -81,11 +81,11 @@ def example_staypoints():
     one_hour = datetime.timedelta(hours=1)
 
     list_dict = [
-        {"user_id": 0, "started_at": t1, "finished_at": t2, "geometry": p1},
-        {"user_id": 0, "started_at": t2, "finished_at": t3, "geometry": p2},
-        {"user_id": 1, "started_at": t3, "finished_at": t3 + one_hour, "geometry": p3},
+        {"user_id": 0, "started_at": t1, "finished_at": t2, "geom": p1},
+        {"user_id": 0, "started_at": t2, "finished_at": t3, "geom": p2},
+        {"user_id": 1, "started_at": t3, "finished_at": t3 + one_hour, "geom": p3},
     ]
-    stps = gpd.GeoDataFrame(data=list_dict, geometry="geometry", crs="EPSG:4326")
+    stps = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
     stps.index.name = "id"
     assert stps.as_staypoints
     return stps
@@ -105,12 +105,12 @@ def example_triplegs():
     one_hour = datetime.timedelta(hours=1)
 
     list_dict = [
-        {"id": 0, "user_id": 0, "started_at": t1, "finished_at": t2, "geometry": g1},
-        {"id": 1, "user_id": 0, "started_at": t2, "finished_at": t3, "geometry": g2},
-        {"id": 2, "user_id": 1, "started_at": t3, "finished_at": t3 + one_hour, "geometry": g3},
+        {"id": 0, "user_id": 0, "started_at": t1, "finished_at": t2, "geom": g1},
+        {"id": 1, "user_id": 0, "started_at": t2, "finished_at": t3, "geom": g2},
+        {"id": 2, "user_id": 1, "started_at": t3, "finished_at": t3 + one_hour, "geom": g3},
     ]
 
-    tpls = gpd.GeoDataFrame(data=list_dict, geometry="geometry", crs="EPSG:4326")
+    tpls = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
     tpls.set_index("id", inplace=True)
 
     assert tpls.as_triplegs
@@ -247,12 +247,14 @@ class TestPositionfixes:
         pfs = example_positionfixes.copy()
         conn_string, conn = conn_postgis
         table = "positionfixes"
+        sql = f"SELECT * FROM {table}"
         geom_col = pfs.geometry.name
 
         try:
             pfs.as_positionfixes.to_postgis(conn_string, table)
-            pfs_db = ti.io.read_positionfixes_postgis(conn_string, table, geom_col)
+            pfs_db = ti.io.read_positionfixes_postgis(sql, conn_string, geom_col)
             pfs_db = pfs_db.set_index("id")
+            print(pfs_db)
             assert_geodataframe_equal(pfs, pfs_db)
         finally:
             del_table(conn, table)
@@ -262,11 +264,12 @@ class TestPositionfixes:
         pfs = example_positionfixes.copy()
         conn_string, conn = conn_postgis
         table = "positionfixes"
+        sql = f"SELECT * FROM {table}"
         geom_col = pfs.geometry.name
         pfs.crs = None
         try:
             pfs.as_positionfixes.to_postgis(conn_string, table)
-            pfs_db = ti.io.read_positionfixes_postgis(conn_string, table, geom_col)
+            pfs_db = ti.io.read_positionfixes_postgis(sql, conn_string, geom_col)
             pfs_db = pfs_db.set_index("id")
             assert_geodataframe_equal(pfs, pfs_db)
         finally:
@@ -296,11 +299,12 @@ class TestStaypoints:
         spts = example_staypoints.copy()
         conn_string, conn = conn_postgis
         table = "staypoints"
+        sql = f"SELECT * FROM {table}"
         geom_col = spts.geometry.name
 
         try:
             spts.as_staypoints.to_postgis(conn_string, table)
-            spts_db = ti.io.read_staypoints_postgis(conn_string, table, geom_col)
+            spts_db = ti.io.read_staypoints_postgis(sql, conn_string, geom_col)
             assert_geodataframe_equal(spts, spts_db)
         finally:
             del_table(conn, table)
@@ -310,11 +314,12 @@ class TestStaypoints:
         spts = example_staypoints.copy()
         conn_string, conn = conn_postgis
         table = "staypoints"
+        sql = f"SELECT * FROM {table}"
         geom_col = example_staypoints.geometry.name
         spts.crs = None
         try:
             spts.as_staypoints.to_postgis(conn_string, table)
-            spts_db = ti.io.read_staypoints_postgis(conn_string, table, geom_col)
+            spts_db = ti.io.read_staypoints_postgis(sql, conn_string, geom_col)
             assert_geodataframe_equal(spts, spts_db)
         finally:
             del_table(conn, table)
@@ -343,11 +348,12 @@ class TestTriplegs:
         tpls = example_triplegs.copy()
         conn_string, conn = conn_postgis
         table = "triplegs"
+        sql = f"SELECT * FROM {table}"
         geom_col = tpls.geometry.name
 
         try:
             tpls.as_triplegs.to_postgis(conn_string, table)
-            tpls_db = ti.io.read_triplegs_postgis(conn_string, table, geom_col)
+            tpls_db = ti.io.read_triplegs_postgis(sql, conn_string, geom_col)
             assert_geodataframe_equal(tpls, tpls_db)
         finally:
             del_table(conn, table)
@@ -357,12 +363,13 @@ class TestTriplegs:
         tpls = example_triplegs.copy()
         conn_string, conn = conn_postgis
         table = "triplegs"
+        sql = f"SELECT * FROM {table}"
         geom_col = tpls.geometry.name
         tpls.crs = None
 
         try:
             tpls.as_triplegs.to_postgis(conn_string, table)
-            tpls_db = ti.io.read_triplegs_postgis(conn_string, table, geom_col)
+            tpls_db = ti.io.read_triplegs_postgis(sql, conn_string, geom_col)
             assert_geodataframe_equal(tpls, tpls_db)
         finally:
             del_table(conn, table)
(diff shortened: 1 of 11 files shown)


bifbof:
There are a lot of warnings, apparently pandas has redesigned its SQL module and geopandas has yet to adapt.

trackintel/io/postgis.py
```diff
@@ -66,56 +94,52 @@ def write_positionfixes_postgis(
...
+        How to behave if the table already exists.
+
+        - fail: Raise a ValueError.
+        - replace: Drop the table before inserting new values.
+        - append: Insert new values to the existing table.
 
     Examples
     --------
-    >>> df.as_positionfixes.to_postgis(conn_string, table_name)
+    >>> pfs.as_positionfixes.to_postgis(conn_string, table_name)
```
hong2223, line 105: would be nice to include also examples of calling this function directly, e.g., 
same for other write functions
```suggestion
    >>> pfs.as_positionfixes.to_postgis(conn_string, table_name)
    >>> ti.io.postgis.write_positionfixes_postgis(pfs, conn_string, table_name)
```

trackintel/io/postgis.py
```diff
@@ -210,11 +230,15 @@ def write_staypoints_postgis(staypoints, conn_string, table_name, schema=None, s
...
+        How to behave if the table already exists.
+
+        - fail: Raise a ValueError.
+        - replace: Drop the table before inserting new values.
+        - append: Insert new values to the existing table.
 
     Examples
     --------
-    >>> df.as_staypoints.to_postgis(conn_string, table_name)
+    >>> spts.as_staypoints.to_postgis(conn_string, table_name)
```
hong2223, line 241:
```suggestion
    >>> spts.as_staypoints.to_postgis(conn_string, table_name)
    >>> ti.io.postgis.write_staypoints_postgis(spts, conn_string, table_name)
```

trackintel/io/postgis.py
```diff
@@ -287,50 +303,67 @@ def write_locations_postgis(locations, conn_string, table_name, schema=None, sql
...
+        How to behave if the table already exists.
+
+        - fail: Raise a ValueError.
+        - replace: Drop the table before inserting new values.
+        - append: Insert new values to the existing table.
 
     Examples
     --------
-    >>> df.as_locations.to_postgis(conn_string, table_name)
+    >>> locs.as_locations.to_postgis(conn_string, table_name)
```
hong2223, line 314:
```suggestion
    >>> locs.as_locations.to_postgis(conn_string, table_name)
    >>> ti.io.postgis.write_locations_postgis(locs, conn_string, table_name)
```

tests/io/test_postgis_gpd.py
```diff
@@ -246,12 +247,14 @@ def test_read(self, example_positionfixes, conn_postgis):
...
         table = "positionfixes"
+        sql = f"SELECT * FROM {table}"
         geom_col = pfs.geometry.name
 
         try:
             pfs.as_positionfixes.to_postgis(conn_string, table)
-            pfs_db = ti.io.read_positionfixes_postgis(conn_string, table, geom_col)
+            pfs_db = ti.io.read_positionfixes_postgis(sql, conn_string, geom_col)
             pfs_db = pfs_db.set_index("id")
+            print(pfs_db)
```
hong2223, line 257: Do we need this printing?

tests/io/test_postgis_gpd.py
```diff
@@ -452,3 +485,30 @@ def test_srid(self, example_positionfixes):
...
+        conn_string, conn = conn_postgis
+
+        @ti.io.postgis._handle_con_string
+        def wrapped(con):
+            assert con is conn
+            return
+
+        wrapped(conn)
+
+    pass
```
hong2223, line 514: delete

tests/io/test_postgis_gpd.py
```diff
@@ -452,3 +485,30 @@ def test_srid(self, example_positionfixes):
...
+        assert wrapped(conn_string).closed
+
+    def test_conn(self, conn_postgis):
+        """Test handeling of connection input"""
+        conn_string, conn = conn_postgis
+
+        @ti.io.postgis._handle_con_string
+        def wrapped(con):
+            assert con is conn
+            return
```
hong2223, line 510: do we need a return here?

tests/io/test_postgis_gpd.py
```diff
@@ -452,3 +485,30 @@ def test_srid(self, example_positionfixes):
         srid = 3857
         gdf.set_crs(f"epsg:{srid}", inplace=True)
         assert _get_srid(gdf) == srid
+
+
+class Test_Handle_Con_String:
+    def test_conn_string(self, conn_postgis):
+        """Test if decorator opens a connection with connection string and closes it."""
+        conn_string, conn = conn_postgis
```
hong2223, line 493:
```suggestion
        conn_string, _ = conn_postgis
```

tests/io/test_postgis_gpd.py
```diff
@@ -452,3 +485,30 @@ def test_srid(self, example_positionfixes):
...
+        def wrapped(con):
+            assert isinstance(con, sqlalchemy.engine.Connection)
+            assert not con.closed
+            return con
+
+        assert wrapped(conn_string).closed
+
+    def test_conn(self, conn_postgis):
+        """Test handeling of connection input"""
+        conn_string, conn = conn_postgis
```
hong2223, line 505:
```suggestion
        _ , conn = conn_postgis
```

trackintel/io/postgis.py
```diff
@@ -22,28 +56,23 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-        pfs = gpd.GeoDataFrame.from_postgis("SELECT * FROM %s" % table_name, conn, geom_col=geom_col, *args, **kwargs)
-    finally:
-        conn.close()
 
-    # assert validity of positionfixes
-    pfs.as_positionfixes
-    return pfs
+    Examples
+    --------
+    >>> pfs = ti.io.read_postifionfixes("SELECT * FROM postionfixes", con, geom_col="geom")
```
hong2223, line 68:
```suggestion
    >>> pfs = ti.io.postgis.read_postifionfixes("SELECT * FROM postionfixes", con, geom_col="geom")
```

trackintel/io/postgis.py
```diff
@@ -66,56 +94,52 @@ def write_positionfixes_postgis(
...
     geom_col : str, default 'geom'
         The geometry column of the table.
 
+    *args
+        Further arguments as available in GeoPanda's GeoDataFrame.from_postgis().
+
+    **kwargs
+        Further keyword arguments as available in GeoPanda's GeoDataFrame.from_postgis().
+
     Returns
```
hong2223, line 131: Is it possible to add **Examples** in all _read_ functions?

hong2223 (approved):
Thanks for the change. One comment (the TODO comment) was left out in the first round, maybe you can have a look? All the others are fine. 

I would request @henrymartin1 for a review since he posted the two issues.

bifbof:
> Thanks for the change. One comment (the TODO comment) was left out in the first round, maybe you can have a look? All the others are fine.
> 
> I would request @henrymartin1 for a review since he posted the two issues.

So I looked into it, the saving works but the loading does not so far. The read_*_gpd function has no argument to set the index by column but at least the old index exists as additional column. Me may have to add the remaining arguments for the to_postgis functions (see #243) and then also update the read_*_gpd function.

trackintel/io/postgis.py
```diff
@@ -139,55 +167,59 @@ def write_triplegs_postgis(
...
+    >>> tpls.as_triplegs.to_postgis(conn_string, table_name)
+    >>> ti.io.postgis.write_triplegs_postgis(tpls, conn_string, table_name)
     """
-    engine = create_engine(conn_string)
-    conn = engine.connect()
-    try:
-        triplegs.to_postgis(table_name, conn, if_exists=if_exists, index=True, chunksize=sql_chunksize)
-    finally:
-        conn.close()
+    triplegs.to_postgis(table_name, con, if_exists=if_exists, schema=schema, index=True, chunksize=sql_chunksize)
```
henrymartin1, line 181: Maybe we can add the missing arguments from the original geopandas [function](https://geopandas.org/docs/reference/api/geopandas.GeoDataFrame.to_postgis.html) as optional input just in case. 
This would concern these two arguments: 
index_label=None, dtype=None

Our default name for the index is the index name (which I think is very good)

trackintel/io/postgis.py
```diff
@@ -210,11 +230,15 @@ def write_staypoints_postgis(staypoints, conn_string, table_name, schema=None, s
...
+        - replace: Drop the table before inserting new values.
+        - append: Insert new values to the existing table.
 
     Examples
     --------
-    >>> df.as_staypoints.to_postgis(conn_string, table_name)
+    >>> spts.as_staypoints.to_postgis(conn_string, table_name)
     """
 
     # todo: Think about a concept for the indices. At the moment, an index
```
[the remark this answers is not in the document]
henrymartin1: I think yes. 
So the default should be that if we download something that is not in trackintel format, that it is changed into trackintel format (e.g., the case that there is no index in the database).
But if we download/upload/download something that is in trackintel format, that the table is not changed.

It is a bit tricky to verify this without a test

henrymartin1 (changes requested):
Really great work! 
I would only propose a minor change so that the `dtype` argument can be passed along to the geopandas functions. 

Also I think it would be good to add two tests to ensure that the open todo is solved by this PR. 
An idea for a test would be the following:
- Stability of trackintel conform tables:
   - There is a dataframe that is trackintel conform in the database. If we download it (->x), upload it and download it again (->y) than x == y should be true.
   - It probably would be better to do the comparison in postgresql/postgis after uploading x again but such a comparison seems a bit harder/weird to do in postgresql ([example1](https://dba.stackexchange.com/questions/72641/checking-whether-two-tables-have-identical-content-in-postgresql/72656), [example2](https://dba.stackexchange.com/questions/206992/check-if-two-tables-have-the-same-structure-are-compatible)

trackintel/io/postgis.py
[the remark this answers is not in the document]
bifbof: So I looked into it, the saving works but the loading does not so far. The read_*_gpd function has no argument to set the index by column but at least the old index exists as additional column. Me may have to add the remaining arguments for the to_postgis functions (see #243) and then also update the read_*_gpd function.

bifbof:
With the last two commits, I added all the parameters of `to_postgis` from geopandas to all the write*postgis functions and adjusted our order of arguments accordingly. Since all functions use exactly the same function in the same way, I took the liberty to simplify the handling with the docstrings for these functions a bit.

I haven't done the index preservation test yet. For that you would have to change all read*postgis functions and the related read*gpd functions. As this PR is already big, I would move that to another PR.

henrymartin1:
> 
> 
> With the last two commits, I added all the parameters of `to_postgis` from geopandas to all the write*postgis functions and adjusted our order of arguments accordingly. Since all functions use exactly the same function in the same way, I took the liberty to simplify the handling with the docstrings for these functions a bit.
> 
> I haven't done the index preservation test yet. For that you would have to change all read*postgis functions and the related read*gpd functions. As this PR is already big, I would move that to another PR.

Sounds good. Could you please open an issue with a description of the problem for this?

bifbof:
> > With the last two commits, I added all the parameters of `to_postgis` from geopandas to all the write*postgis functions and adjusted our order of arguments accordingly. Since all functions use exactly the same function in the same way, I took the liberty to simplify the handling with the docstrings for these functions a bit.
> > I haven't done the index preservation test yet. For that you would have to change all read_postgis functions and the related read_gpd functions. As this PR is already big, I would move that to another PR.
> 
> Sounds good. Could you please open an issue with a description of the problem for this?

Already did. :)

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-        conn.close()
 
-    # assert validity of positionfixes
-    pfs.as_positionfixes
-    return pfs
+    Examples
+    --------
+    >>> pfs = ti.io.read_postifionfixes_postgis("SELECT * FROM postionfixes", con, geom_col="geom")
+    """
+    pfs = gpd.GeoDataFrame.from_postgis(sql, con, geom_col, *args, **kwargs)
```
henrymartin1, line 70: This is OK

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
 
 
+@_handle_con_string
 def write_positionfixes_postgis(
-    positionfixes, conn_string, table_name, schema=None, sql_chunksize=None, if_exists="fail"
+    positionfixes, name, con, schema=None, if_exists="fail", index=True, index_label=None, chunksize=None, dtype=None
 ):
-    """Stores positionfixes to PostGIS. Usually, this is directly called on a positionfixes
-    DataFrame (see example below).
+    positionfixes.to_postgis(name, con, schema, if_exists, index, index_label, chunksize, dtype)
```
henrymartin1, line 78: Here we should use keyword arguments to tunnel these many arguments to prevent future errors.

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-    conn = engine.connect()
-    try:
-        positionfixes.to_postgis(table_name, conn, if_exists=if_exists, index=True, chunksize=sql_chunksize)
-    finally:
-        conn.close()
-
-
-def read_triplegs_postgis(conn_string, table_name, geom_col="geom", *args, **kwargs):
+@_handle_con_string
+def read_triplegs_postgis(sql, con, geom_col="geom", *args, **kwargs):
```
henrymartin1, line 82: It is not allowed to have arguments after an keyword argument

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-        conn.close()
 
-    # assert validity of triplegs
-    pfs.as_triplegs
-    return pfs
+    Examples
+    --------
+    >>> tpls = ti.io.read_triplegs_postgis("SELECT * FROM triplegs", con, geom_col="geom")
+    """
+    tpls = gpd.GeoDataFrame.from_postgis(sql, con, geom_col=geom_col, index_col="id", *args, **kwargs)
```
henrymartin1, line 111: It is not allowed to use positional arguments (*args) after a keyword argument (index_col="id")

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-    triplegs : GeoDataFrame
-        The triplegs to store to the database.
-
-    conn_string : str
-        A connection string to connect to a database, e.g.,
-        ``postgresql://username:password@host:socket/database``.
-
-    table_name : str
-        The name of the table to write to.
+    triplegs.to_postgis(name, con, schema, if_exists, index, index_label, chunksize, dtype)
```
henrymartin1, line 119: I think we should use keyword arguments to pass on so many arguments

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-    conn = engine.connect()
-    try:
-        triplegs.to_postgis(table_name, conn, if_exists=if_exists, index=True, chunksize=sql_chunksize)
-    finally:
-        conn.close()
-
-
-def read_staypoints_postgis(conn_string, table_name, geom_col="geom", *args, **kwargs):
+@_handle_con_string
+def read_staypoints_postgis(sql, con, geom_col="geom", *args, **kwargs):
```
henrymartin1, line 123: It is not allowed to use positional arguments (*args) after a keyword argument

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-    try:
-        staypoints.to_postgis(table_name, conn, if_exists=if_exists, index=True, chunksize=sql_chunksize)
-    finally:
-        conn.close()
+
+@_handle_con_string
+def write_staypoints_postgis(
+    staypoints, name, con, schema=None, if_exists="fail", index=True, index_label=None, chunksize=None, dtype=None
+):
+    staypoints.to_postgis(name, con, schema, if_exists, index, index_label, chunksize, dtype)
```
henrymartin1, line 163: I would use kwds

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
 
 
-def read_locations_postgis(conn_string, table_name, geom_col="geom", *args, **kwargs):
+@_handle_con_string
+def read_locations_postgis(sql, con, geom_col="geom", *args, **kwargs):
```
henrymartin1, line 167: It is not allowed to use positional arguments (*args) after a keyword argument

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
     """
-    engine = create_engine(conn_string)
-    conn = engine.connect()
-    try:
-        locs = gpd.GeoDataFrame.from_postgis(
-            "SELECT * FROM %s" % table_name, conn, geom_col=geom_col, index_col="id", *args, **kwargs
-        )
-    finally:
-        conn.close()
+    locs = gpd.GeoDataFrame.from_postgis(sql, con, geom_col=geom_col, index_col="id", *args, **kwargs)
```
henrymartin1, line 196: It is not allowed to use positional arguments (*args) after a keyword argument

trackintel/io/postgis.py
```diff
@@ -22,348 +56,259 @@ def read_positionfixes_postgis(conn_string, table_name, geom_col="geom", *args,
...
-    engine = create_engine(conn_string)
-    conn = engine.connect()
-    try:
-        locations.to_postgis(table_name, conn, if_exists=if_exists, index=True, chunksize=sql_chunksize)
-    finally:
-        conn.close()
+    >>> trips = ti.io.read_trips_postgis("SELECT * FROM trips", con, geom_col="geom")
 
+    """
+    trips = pd.read_sql(sql, con, index_col="id", *args, **kwargs)
```
henrymartin1, line 254: It is not allowed to use positional arguments (*args) after a keyword argument

henrymartin1 (changes requested):
Hey nice work, fancy idea with the doc strings ;-) 
I would leave them now like they are but I think in general it is a bit weird to not have docstrings directly in the code. I think a better alternative would be to link the geopandas documentation for further information.

I like the minimalistic style for the reading function to simply pass everything to the geopandas function but I would change it to keyword assignments to exclude positional errors. 
I wanted to fix this myself but then I found a different error where we assign keyword arguments and then pass `*args` which I think is not allowed. I have marked these location in the code

henrymartin1 (approved):
Great work, thanks for the PR!

PR #265, opened 2021-07-07
Access geometry via geom_col attribute without renaming

### LINKED ISSUE #262: Allow for flexible geometry names in `read_*_gpd` functions.
For an old model of trackintel all geometry columns had to have the same name "geom". Later we dropped this prerequisite and now we access the name directly via the geometry attribute (see [contributing.md](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#adressing-geometry-columns)). This old model still persists in the `read_*_gpd` functions.
The proposed change would be to not rename the geometry column anymore and just add it directly.

Like:
```diff
- columns = {tracked_at: "tracked_at", user_id: "user_id", geom: "geom"}
+ columns = {tracked_at: "tracked_at", user_id: "user_id"}
  columns.update(mapper)
  pfs = gdf.rename(columns=columns)
- pfs = pfs.set_geometry("geom")
+ pfs = pfs.set_geometry(geom)
```

There are may some side effects that must be handled.
Additionally may rename argument to `geom_col` to be more consistent with the rest of the io module.

NinaWie:
closes #262

The geometry is now not renamed anymore and instead accessed via the attribute. The attribute was furthermore renamed from `geom` to `geom_col`. This requires a **change in the tutorial notebook**. 
Also, the tests fail because the generated table (e.g. `pfs_from_gpd`) does not correspond to the ground truth table loaded from a csv file (e.g. `tpls_from_csv`). To avoid a change in the test data, I renamed the column of the ground truth table in the test file.

### DIFF
--- a/tests/io/test_from_geopandas.py
+++ b/tests/io/test_from_geopandas.py
@@ -17,7 +17,7 @@ class TestFromGeopandas:
 
         pfs_file = os.path.join("tests", "data", "positionfixes.csv")
         pfs_from_csv = ti.read_positionfixes_csv(pfs_file, sep=";", tz="utc", index_col="id")
-        pfs_from_csv = pfs_from_csv.rename(columns={"geom":"geometry"})
+        pfs_from_csv = pfs_from_csv.rename(columns={"geom": "geometry"})
 
         pd.testing.assert_frame_equal(pfs_from_gpd, pfs_from_csv, check_exact=False)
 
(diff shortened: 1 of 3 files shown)


tests/io/test_from_geopandas.py
```diff
@@ -13,21 +13,23 @@ def test_read_positionfixes_gpd(self):
         """Test if the results of reading from gpd and csv agrees."""
         gdf = gpd.read_file(os.path.join("tests", "data", "positionfixes.geojson"))
         gdf.set_index("id", inplace=True)
-        pfs_from_gpd = ti.io.from_geopandas.read_positionfixes_gpd(gdf, user_id="User", geom="geometry", tz="utc")
+        pfs_from_gpd = ti.io.from_geopandas.read_positionfixes_gpd(gdf, user_id="User", geom_col="geometry", tz="utc")
```
NinaWie, line 16: In `read_positionfixes.gpd` there is an optional parameter `columns`. I thought I could use this parameter to rename the column to `geometry`, but it looks like all columns can be renamed except for the `geometry` column. Therefore, I had to rename the column in the test function in order to make `assert_frame_equal` work.

hong2223 (approved):
First pull request for `trackintel`! Nice fix, I think we can merge this now

hong2223:
We try to follow the standard pandas commit message (with prefixes) for arranging a bit our commits and pull requests, [here](https://pandas.pydata.org/pandas-docs/stable/development/contributing.html#committing-your-code). I would now rename this PR to `CLN: Access geometry via geom_col attribute without renaming`

PR #220, opened 2021-05-21
ENH: Cleaner and faster trip generation

henrymartin1:
fixes #219

Main idea of the new implementation is that we assign a trip_ids to a combined dataframe of triplegs and staypoints so that we are able to create trips using a groupby statement that is based on the assigned trip id.

The implementation itself has several main steps:

0. Merge triplegs and staypoints into a single dataframe and sort them by [user_id, started_at]
1. identify start of new trips (activities, user change, temporal gaps) 
2. assign trip ids to all triplegs/staypoints that belong to a trip
3. delete activities as they are not part of trips
4. group by trip_id to create trips
5. add activities again to trip dataframe to assign origin_staypoint_id and destination_staypoint_id
6. delete activities again
7. Do some id stuff

Some tests are already passing but I have added the remaining todos in the code

PR #271, opened 2021-07-12
CLN: trips as geodataframe with start and end coordinates

### LINKED ISSUE #57: start and endpoint for trip model
I think trips in trackintel should have the start and end point as geometry. Even though from the perspective of creating trips from staypoints this is redundant but in real-life use cases it might appear that we only analyze trips without having access to staypoints.

- Add columns geom_start and geom_end as required (and with a fixed name I guess...)
- Add column primary_mode as an optional column

NinaWie:
Closes #57 
The aim was to add start and end point of trips to the trips data frame. This however required major changes to the tests, since so far the trips data frames are assumed to be dataframes in trackintel, and not geodataframes. An alternative to changing the test data would be given by keeping the compatibility with data frames, e.g. adding the following line in read and write functions:
```
if "geom" in df.columns:
    df["geom"] = df["geom"].apply(wkt.dumps)
```

Also, problems occur if start or end activity of a trip are not given, because MultiPoint cannot contain null points or empty points. I thus suggest to set the start coordinates of the trip to the start of its first tripleg (and same for the destination coordinates, to the last coordinates on the last tripleg) as an approximation. Happy for other suggestions though.


trackintel/io/file.py
```diff
@@ -472,6 +472,10 @@ def read_trips_csv(*args, columns=None, tz=None, index_col=object(), **kwargs):
         if not pd.api.types.is_datetime64tz_dtype(trips[col]):
             trips[col] = _localize_timestamp(dt_series=trips[col], pytz_tzinfo=tz, col_name=col)
 
+    # convert to geodataframe
+    trips["geom"] = trips["geom"].apply(wkt.loads)
```
NinaWie, line 476: Here one would have to add an if clause to keep the compatibility with dataframes (compared to geodataframes)

tests/data/trips_mod_columns.csv
```diff
@@ -1,5 +1,5 @@
-id;user_id;started_at;finished_at;orig_stp;dest_stp
-0;1;2015-11-27T08:00:00Z;2015-11-27T08:15:00Z;2;5
-1;1;2015-11-27T08:20:22Z;2015-11-27T08:35:22Z;2;5
-2;1;2015-11-27T12:39:28Z;2015-11-27T12:55:28Z;2;5
-3;1;2015-11-27T13:57:20Z;2015-11-27T14:13:20Z;2;5
\ No newline at end of file
+id;user_id;started_at;finished_at;orig_stp;dest_stp;geom
```
NinaWie, line 1: I left the geom column here in order to test whether the dataframes are equal (except for this column) - Tested in `test_file.py`. Apart from this one and the `geolife_long/trips.csv`, I changed the test data back to dataframes, since trips tables now support either dataframes or geodataframes.

trackintel/model/trips.py
```diff
@@ -25,10 +28,11 @@ class TripsAccessor(object):
 
     The following assumptions are implemented
     
-        - All movement before the first and after the last activity is omitted
```
NinaWie, line 28: I removed this because I think it is wrong, or misleading. As far as I could see, the triplegs before or after the first or last activity are also considered as trips, and are not ommitted. Please correct me if I'm wrong there

henrymartin1:
Results from our last meeting: 
- Geometry should be optional
- Multipoint with the first point of the first tripleg (or the last point of the last tripleg) can be used if the start/end point is missing because multipoint does not allow for missing geometries

tests/data/geolife_long/trips.csv
```diff
@@ -1,15 +1,15 @@
...
-6,1,2008-10-23 10:32:53+00:00,2008-10-23 11:10:19+00:00,10.0,11.0
-7,1,2008-10-23 11:49:08+00:00,2008-10-23 11:49:48+00:00,11.0,12.0
-8,1,2008-10-23 23:43:46+00:00,2008-10-24 00:23:03+00:00,12.0,14.0
-9,1,2008-10-24 01:45:41+00:00,2008-10-24 02:02:46+00:00,14.0,15.0
-10,1,2008-10-24 02:28:19+00:00,2008-10-24 02:31:32+00:00,15.0,16.0
-11,1,2008-10-24 03:16:35+00:00,2008-10-24 04:12:50+00:00,16.0,20.0
-12,1,2008-10-24 05:28:05+00:00,2008-10-24 05:39:50+00:00,20.0,21.0
-13,1,2008-10-24 06:08:42+00:00,2008-10-24 06:35:50+00:00,21.0,
+id,user_id,started_at,finished_at,origin_staypoint_id,destination_staypoint_id,geom
+0,0,2008-10-23 02:53:04+00:00,2008-10-23 03:05:00+00:00,,0.0,"MULTIPOINT (116.3184169999999966 39.9847019999999986, 116.2987250000000046 39.9840190000000035)"
```
henrymartin1, line 2: If the geometry is optional there should be some tests without geometry to see if that works correctly

tests/data/trips_mod_columns.csv
henrymartin1: Ah ok. So to have some test data with and some without geometry?

tests/io/test_file.py
```diff
@@ -280,12 +280,13 @@ def test_from_to_csv(self):
...
         mod_trips = ti.read_trips_csv(mod_file, columns=column_mapping, sep=";", index_col="id")
-        assert mod_trips.equals(trips)
+        mod_trips_wo_geom = pd.DataFrame(mod_trips.drop(columns=["geom"]))
+        assert mod_trips_wo_geom.equals(trips)
 
         trips["started_at"] = trips["started_at"].apply(lambda d: d.isoformat().replace("+00:00", "Z"))
         trips["finished_at"] = trips["finished_at"].apply(lambda d: d.isoformat().replace("+00:00", "Z"))
         columns = ["user_id", "started_at", "finished_at", "origin_staypoint_id", "destination_staypoint_id"]
-        trips.as_trips.to_csv(tmp_file, sep=";", columns=columns)
+        ti.io.write_trips_csv(trips, tmp_file, sep=";", columns=columns)
```
henrymartin1, line 289: We should talk about accessors ;-)

tests/preprocessing/test_triplegs.py
```diff
@@ -65,7 +65,21 @@ def test_generate_trips(self):
...
+        # test if coordinates of start and destination are correct
+        for i, row in trips.iterrows():
+            if not pd.isna(row["origin_staypoint_id"]):
+                start_point_trips = row["geom"][0]  # get origin Point in generated trips
+                corresponding_stp = stps.loc[row["origin_staypoint_id"], "geom"]
+                # compare to the Point in the staypoints
+                assert corresponding_stp == start_point_trips
+
+            if not pd.isna(row["destination_staypoint_id"]):
+                dest_point_trips = row["geom"][1]  # get origin Point in generated trips
```
henrymartin1, line 79: Change comment to destination

tests/preprocessing/test_triplegs.py
```diff
@@ -65,7 +65,21 @@ def test_generate_trips(self):
         # generate trips and a joint staypoint/triplegs dataframe
         stps, tpls, trips = ti.preprocessing.triplegs.generate_trips(stps, tpls, gap_threshold=15)
         # test if generated trips are equal
-        pd.testing.assert_frame_equal(trips_loaded, trips)
+        assert_geodataframe_equal(trips_loaded, trips)
+
+        # test if coordinates of start and destination are correct
```
henrymartin1, line 70: I think this should be its own test.

trackintel/io/file.py
```diff
@@ -424,6 +424,7 @@ def read_trips_csv(*args, columns=None, tz=None, index_col=object(), **kwargs):
         The column names to rename in the format {'old_name':'trackintel_standard_name'}.
         The required columns for this function include: "user_id", "started_at",
         "finished_at", "origin_staypoint_id" and "destination_staypoint_id".
+        An optional column is "geom" (geometry column with start and destination location)
```
henrymartin1, line 427: Here we should describe that the type is "multipoint"

trackintel/io/file.py
```diff
@@ -434,8 +435,8 @@ def read_trips_csv(*args, columns=None, tz=None, index_col=object(), **kwargs):
 
     Returns
     -------
-    trips : DataFrame (as trackintel trips)
-        A DataFrame containing the trips.
+    trips : GeoDataFrame / DataFrame (as trackintel trips)
```
henrymartin1, line 438: I think the return type should be unique

trackintel/io/file.py
henrymartin1: I would be good if you could update the example under "examples" in the docstring below

trackintel/io/file.py
```diff
@@ -472,6 +473,11 @@ def read_trips_csv(*args, columns=None, tz=None, index_col=object(), **kwargs):
         if not pd.api.types.is_datetime64tz_dtype(trips[col]):
             trips[col] = _localize_timestamp(dt_series=trips[col], pytz_tzinfo=tz, col_name=col)
 
+    # convert to geodataframe
+    if "geom" in trips.columns:
+        trips["geom"] = trips["geom"].apply(wkt.loads)
+        trips = gpd.GeoDataFrame(trips, geometry="geom")
```
henrymartin1, line 479: if we want a unique return type, then it should always be transformed into a geodataframe

trackintel/model/trips.py
```diff
@@ -12,8 +12,11 @@ class TripsAccessor(object):
...
     ['user_id', 'started_at', 'finished_at', 'origin_staypoint_id', 'destination_staypoint_id']
+    Optional column:
+    ['geom'] 
+    Trips is treated as a GeoDataFrame if a geometry column is provided (containing start and destination points)
 
-    The 'index' of the GeoDataFrame will be treated as unique identifier of the `Trips`
+    The 'index' of the (Geo)DataFrame will be treated as unique identifier of the `Trips`
 
     For several usecases, the following additional columns are required:
     ['context', 'origin_activity', 'destination_activity', 'modes', 'primary_mode', 'tour_id']
```
henrymartin1, line 22: I think the geometry column could be added here. Additionally I would add the sentence "Trips have an optional geometry of type multipoint which describes the start and the end point of the trip" or something similar.

This docstring here is very important because it serves as the main documentation for trips.

trackintel/model/trips.py
henrymartin1: No, I think you are right. It was different in the past, thanks for the update

trackintel/model/trips.py
```diff
@@ -25,10 +28,11 @@ class TripsAccessor(object):
 
     The following assumptions are implemented
     
-        - All movement before the first and after the last activity is omitted
         - If we do not record a person for more than `gap_threshold` minutes, we assume that the person performed an \
             activity in the recording gap and split the trip at the gap.
-        - Trips that start/end in a recording gap can have an unknown origin/destination.
+        - Trips that start/end in a recording gap can have an unknown origin/destination staypoint.
+        - If the origin (or destination) staypoint is unkown (and a geometry column exists), the origin/destination
```
henrymartin1, line 34: typo in "unkown"

trackintel/model/trips.py
henrymartin1: The sentence I proposed could also be added to "notes"

trackintel/preprocessing/triplegs.py
```diff
@@ -229,6 +232,10 @@ def _generate_trips_user(df, gap_threshold):
         if in_trip is True:
             # during trip generation/recording
 
+            # if the origin activity is unknown, fill origin coordinates with coordinates of first tripleg
+            if origin_activity["geom"] is None:
```
henrymartin1, line 236: Oh shit... this will be horrible to merge with our updated trip generation [PR220](https://github.com/mie-lab/trackintel/pull/220)
The best thing to do might be to wait until we merged it... Unfortunatly it is somewhat stuck with a bug we are working on...

trackintel/model/trips.py
```diff
@@ -61,6 +65,15 @@ def _validate(obj):
             obj["finished_at"]
         ), "dtype of finished_at is {} but has to be datetime64 and timezone aware".format(obj["finished_at"].dtype)
 
+        # Check geometry if Trips is a GeoDataFrame
+        if "geom" in obj.columns:
```
henrymartin1, line 69: nice!

trackintel/model/trips.py
```diff
@@ -61,6 +65,15 @@ def _validate(obj):
...
+            # check geometry
+            assert obj.geometry.is_valid.all(), (
+                "Not all geometries are valid. Try x[~ x.geometry.is_valid] " "where x is you GeoDataFrame"
+            )
+            if obj.geometry.iloc[0].geom_type != "MultiPoint":
+                raise AttributeError("The geometry must be a MultiPoint (only first checked).")
```
henrymartin1, line 75: Can you add a test for that special case? Can be a small one with a dummy example

henrymartin1 (changes requested):
Great work! It is honestly a complicated issue because it touches on the core of the data model. 
Special credit for reading through the trip generation! I hope you still like trackintel ;-) 

Here are a few additional comments: 
- Is it possible to have it always as a Geodataframe even though it sometimes is without geometries? I think that would be cleaner than if it always changes types
- We should have a test that checks if the behavior for taking the first point of a tripleg or the last point of a staypoint is correct. 

The generate_trips function is under review at the moment and will receive a major update (#220) we should discuss on Wednesday what we merge first.

henrymartin1:
Reminder: In the end we will have to update the documentation

NinaWie:
I changed the structure according to your comment @henrymartin1 such that the trips is always a GeoDataFrame, no matter if it contains a geometry column or not. However, I get an error and I'm confused about this test: https://github.com/mie-lab/trackintel/blob/e0c0cdd0d8472ba7b113b3819d062ea8abcd8168/tests/io/test_from_geopandas.py#L64
In all other `from_geopandas` tests, the reading from a .geojson file is compared to reading form a csv file. In contrast, in the trips tests the csv file is basically compared to a csv. Since one of them is a GeoDataFrame now, the test fails. 

I could now either 1) change the test such that the first one is also a gdf, which however makes the test quite useless (because we load the same csv file in both cases and then convert it to a gdf), or 2) generate a new geojson file from the `trips.csv` (which does not have a geometry column) or `geolife_long/trips.csv` (which has a geometry column). 
Let me know what you think.

tests/data/geolife_long/trips.csv
NinaWie: Yes, these tests exist in `io/test_file.py`

trackintel/io/file.py
NinaWie: Changed this now

trackintel/model/trips.py
NinaWie: Also changed with the last commits

trackintel/model/trips.py
NinaWie: I added this test in `tests/model/test_trips.py`, see below

trackintel/preprocessing/triplegs.py
NinaWie: Yes let's discuss this on Wednesday

tests/io/test_from_geopandas.py
```diff
@@ -69,6 +69,7 @@ def test_read_trips_gpd(self):
 
         trips_file = os.path.join("tests", "data", "trips.csv")
         trips_from_csv = ti.read_trips_csv(trips_file, sep=";", tz="utc", index_col="id")
+        trips_from_csv = pd.DataFrame(trips_from_csv)
```
NinaWie, line 72: I had to add this line to make the test work (since `read_trips_csv` now always returns a GeoDataFrame), but as I commented, the test is rather useless now. Maybe we should add a geojson for `trips.csv`.

tests/data/trips_mod_columns.csv
NinaWie: Yes, as said above, in `test_file.py` some tests read csv files with geometry and others without, and the two are compared as well for consistency

tests/preprocessing/test_triplegs.py
```diff
@@ -162,6 +206,7 @@ def test_compare_to_old_trip_function(self):
         # generate trips and a joint staypoint/triplegs dataframe
         stps, tpls, trips = generate_trips(stps, tpls, gap_threshold=15)
         stps_, tpls_, trips_ = _generate_trips_old(stps, tpls, gap_threshold=15)
+        trips.drop(columns=["geom"], inplace=True)
```
NinaWie, line 209: Here, we could also change the `_old_generate_trips` function to include the geometry, but it seems a bit overkill since the function is deprecated anyway. However, it would then of course be tested, whether iterating for each user leads also to the same geometry column as the function implemented now (at the moment I am sure that it yields the same result, since the test data was generated with my previous iterative version). Let me know if you think we should modify the `_old_generate_trips`.

trackintel/preprocessing/triplegs.py
```diff
@@ -108,11 +111,8 @@ def generate_trips(spts, tpls, gap_threshold=15):
     spts["type"] = "staypoint"
 
     # create table with relevant information from triplegs and staypoints.
-    spts_tpls = pd.concat(
-        [
-            spts[["started_at", "finished_at", "user_id", "type", "activity"]],
-            tpls[["started_at", "finished_at", "user_id", "type"]],
-        ]
+    spts_tpls = spts[["started_at", "finished_at", "user_id", "type", "activity", "geom"]].append(
+        tpls[["started_at", "finished_at", "user_id", "type", "geom"]]
```
NinaWie, line 115: concat does not work for geometries

trackintel/preprocessing/triplegs.py
```diff
@@ -205,6 +209,21 @@ def _seperate_ids(row):
...
 
+    # fill NaN start and end points with start of first tripleg (or end of last tripleg) on the trip
+    def geom_by_tripleg(trips, tpls, is_origin):
+        point_idx = 0 if is_origin else -1
+        col = "origin_staypoint_id" if is_origin else "destination_staypoint_id"
+        nan_geom_cols = trips[pd.isna(trips[col])].copy()
+        nan_geom_cols.loc[:, "tpls_id"] = nan_geom_cols.tpls.map(lambda x: x[point_idx])
+        nan_geom_cols = nan_geom_cols.join(tpls, on="tpls_id", how="left", lsuffix="_l", rsuffix="_r")
+        nan_geom_cols["geom_point"] = nan_geom_cols.geom_r.apply(lambda x: x.boundary[point_idx])
+        return nan_geom_cols["geom_point"]
```
NinaWie, line 220: I did use the triplegs id now to join the two tables, in order to fill in the points where the staypoint is NaN. Maybe there is a smarter way to do this.

NinaWie (commented):
I wondered whether we should make the `geom` column optional in `generate_trip`, i.e. as a kw argument. This would avoid the computational effort of getting the geometry from stps and tpls if the user does not need it. Let me know if you think we should change that.

tests/io/test_from_geopandas.py
henrymartin1: I am not so sure if I understand what you mean.
Does that mean, that `read_trips_gpd` does not return a geodataframe?

henrymartin1 (changes requested):
Great work! Especially because this is quite a complex PR that affects so many different areas of the library. It seems like you know trackintel quite well now ;-)
I have added some comments in the file and I have two general comments:
- Be aware that any input geodataframe can have arbitrary geometry column names (e.g., spts/tpls in the generate_trips function.
- Please add a bullet point in the notes section of the docstring of the generate_trips function that says something like "trips optionally have their start and end point as geometry of type multipoint". You should also describe in the docstring that we take the first point of the first tripleg in case of missing origin/destination and that these trips can still be found because they miss the orgin/destination id. (Maybe its rather a small subsection than a bulletpoint ;-) )
- I think it would be a good idea to make the calculation of the geometry optional to have a faster version of the trip_generation if needed (if it is not too much work). I think the default should be that the geometry is generated

trackintel/model/trips.py
```diff
@@ -65,6 +67,15 @@ def _validate(obj):
             obj["finished_at"]
         ), "dtype of finished_at is {} but has to be datetime64 and timezone aware".format(obj["finished_at"].dtype)
 
+        # Check geometry if Trips is a GeoDataFrame
+        if "geom" in obj.columns:
```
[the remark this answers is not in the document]
NinaWie: At the moment, checking `obj.geometyr is None` only leads to an error `AttributeError: No geometry data set yet`. Furthermore, it seems that geopandas will at some point be updated to return a dataframe if no geometry column exists (https://github.com/geopandas/geopandas/issues/1911). This is problematic for us, since the code now is designed in a way that the geometry column is optional, but trips is still always a GeoDataFrame. Since this is already tricky, and will probably become impossible with new geopandas versions, I'd maybe suggest to go back to the option we previously considered, where trips can be either a GeoDataframe or a DataFrame? Let me know what you think.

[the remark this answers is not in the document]
henrymartin1: Ah, I see. Yes in this case it is better to go for your initial idea. In the case it is a geodataframe we still need to address the geometry column dynamically.

tests/preprocessing/test_triplegs.py
```diff
@@ -65,8 +66,72 @@ def test_generate_trips(self):
...
+        )
+        stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
+        pfs, tpls = pfs.as_positionfixes.generate_triplegs(stps)
+
+        # generate trips dataframe with geometry
+        _, _, trips = generate_trips(stps, tpls, gap_threshold=15)
+        trips = pd.DataFrame(trips.drop(["geom"], axis=1))
+
+        # generate trips without geometry
+        _, _, trips_wo_geom = generate_trips(stps, tpls, gap_threshold=15, add_geometry=False)
```
NinaWie, line 90: I added this test to compare the result with and without computing the geometry

tests/preprocessing/test_triplegs.py
```diff
@@ -287,7 +353,7 @@ def test_only_staypoints_in_trip(self):
         spts_tpls = pd.DataFrame(spts_tpls)
         spts = spts_tpls[spts_tpls["type"] == "staypoint"]
         tpls = spts_tpls[spts_tpls["type"] == "tripleg"]
-        spts_, tpls_, trips = generate_trips(spts, tpls)
+        spts_, tpls_, trips = generate_trips(spts, tpls, add_geometry=False)
```
NinaWie, line 356: The optional `add_geometry` actually simplified some tests; here for example we don't need to drop the geometry afterwards.

trackintel/io/file.py
```diff
@@ -485,13 +497,16 @@ def write_trips_csv(trips, filename, *args, **kwargs):
...
+    trips : (Geo)DataFrame (as trackintel trips)
         The trips to store to the CSV file.
 
     filename : str
         The file to write to.
     """
     df = trips.copy()
+    if isinstance(df, GeoDataFrame):
+        geom_col_name = df.geometry.name
+        df[geom_col_name] = df[geom_col_name].apply(wkt.dumps)
```
NinaWie, line 509: I am now using `isinstance` in all cases to check whether the geometry exists or not. We have to be careful with GeoDataFrames that do not contain a geometry in further PRs

trackintel/preprocessing/triplegs.py
```diff
@@ -74,7 +82,12 @@ def generate_trips(spts, tpls, gap_threshold=15):
         - If we do not record a person for more than `gap_threshold` minutes,
           we assume that the person performed an activity in the recording gap and split the trip at the gap.
         - Trips that start/end in a recording gap can have an unknown origin/destination
-        - There are no trips without a (recored) tripleg
+        - There are no trips without a (recorded) tripleg
+        - Trips optionally have their start and end point as geometry of type MultiPoint, if `add_geometry==True`
+        - If the origin (or destination) staypoint is unknown, and `add_geometry==True`, the origin (and destination)
+            geometry is set as the first coordinate of the first tripleg (or the last coordinate of the last tripleg),
+            respectively. Trips with missing values can still be identified via col `origin_staypoint_id`.
```
NinaWie, line 89: Added this to the notes in response to your comment

trackintel/preprocessing/triplegs.py
```diff
@@ -218,6 +239,26 @@ def _seperate_ids(row):
         axis=1,
     )
 
+    # fill the missing points and convert to multipoint
+    if add_geometry:
+        # fill coordinates for origin staypoints that are NaN
+        origin_nan_rows = trips[pd.isna(trips["origin_staypoint_id"])].copy()
+        trips.loc[pd.isna(trips["origin_staypoint_id"]), "origin_geom"] = origin_nan_rows.tpls.map(
+            # from tpls table, get the first point of the first tripleg for the trip
+            lambda x: tpls.loc[x[0], tpls.geometry.name].boundary[0]
```
NinaWie, line 248: I liked your solution better, and I think it was only unreadable because of the `trip_idx` variable etc that I had earlier. Now, I got rid of the nested function and simply repeat the same step for the distination.

NinaWie (commented):
Should be much cleaner now, thanks for your comments! Let me know what else is left to do

tests/io/test_from_geopandas.py
NinaWie: Never mind, this changed again with the last commit :)

trackintel/model/trips.py
```diff
@@ -28,11 +31,11 @@ class TripsAccessor(object):
...
 
     The following assumptions are implemented
-
-        - All movement before the first and after the last activity is omitted
         - If we do not record a person for more than `gap_threshold` minutes, we assume that the person performed an \
             activity in the recording gap and split the trip at the gap.
-        - Trips that start/end in a recording gap can have an unknown origin/destination.
+        - Trips that start/end in a recording gap can have an unknown origin/destination staypoint.
+        - If the origin (or destination) staypoint is unknown (and a geometry column exists), the origin/destination
+            geometry is set as the first coordinate of the first tripleg (or the last coordinate of the last tripleg)
```
henrymartin1, line 38: [Doc]
"geometry" needs to start exactly below "if" 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks

henrymartin1 (changes requested):
Hey Nina, 
great work! I think everything concerning content is done. 🎉 I had several small minor comments concerning the clearity of the comments and the documentation. Usually I would simply add them myself and then merge but I had one question in ` tests/model/test_trips.py ` where I was not sure if it was a typo or not .
For the other comments I have added what I would have changed as comments, you can either include this directly or I can do it before I merge it.


[Documentation]
At the end of an important pull request it is often a good idea to compile the documentation and check if everything is set correctly. [Here](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#documentation) is a bad description on how to do it. After you installed sphinx I think you also need to install "sphinx_rtd_theme" but I am not sure...
It is not really necessary that you do that now everything works fine if you fix the number of spaces in the docstring of `TripsAccessor`.

hong2223 (approved):

NinaWie:
Thanks for the review; I did the changes to the documentation as you suggested :) I also compiled it locally and - after removing some more spaces - it looked fine. Let me know if there is anything else to do

henrymartin1 (approved):
Nice work, I am very happy to merge this now. Congratulations and thank you for your first major contribution to the framework!

PR #272, opened 2021-07-12
CLN: osx specifics in gitignore

NinaWie:
Would be nice if we could add these things to the gitignore, for avoiding the OSX DS_Store files.


hong2223 (approved):
👍

PR #274, opened 2021-07-14
ENH: Copy docstring for accessor methods.

### LINKED ISSUE #209: Add documentation to accessor methods.
The accessor functions with the pattern `as_triplegs.xyz` do not have a complete documentation, but rather refer to the wrapped functions in a different modules. This is inconvenient for working in an editor, as one cannot look up the functionality of the various parameters in the docstring. Instead one has to look up the referenced docstring seperatly.

Two ways to change that would be:
1. Just copy the docstring over at every change. 
2. Use a decorator to copy the docstring directly.

The first way is rather error-prone as one has to remember to change both docstrings at any change. Here it would be good to write a test that checks the equality of the `__doc__` attribute of the functions. The second way can be unclear in the documentation as it replaces silently the existing docstring. Also other problem arises in terms of circular imports etc. .

bifbof:
closes #209 

Added the function `copy_docstring` decorator that uses `functools.update_wrapper` to copy the docstring from the method to the accessor.
To copy the docstring from another function, the other function must be defined at creation time, so I added many imports for all functions. It needs to be verified that it works on other machines before we merge it, because I'm not sure if relative or absolute imports are used on my machine (and if that would matter).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,12 @@ target/
 # Jupyter Notebook
 .ipynb_checkpoints
 
+# VS Code
+*.code-workspace
+
+# OSX specific
+.DS_Store
+
 # pyenv
 .python-version
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 [![Documentation Status](https://readthedocs.org/projects/trackintel/badge/?version=latest)](https://trackintel.readthedocs.io/en/latest/?badge=latest)
 [![codecov.io](https://codecov.io/gh/mie-lab/trackintel/coverage.svg?branch=master)](https://codecov.io/gh/mie-lab/trackintel)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
-   
+[![Open in Visual Studio Code](https://open.vscode.dev/badges/open-in-vscode.svg)](https://open.vscode.dev/mie-lab/trackintel)
+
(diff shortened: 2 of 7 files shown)


hong2223:
I can confirm that the solution works on my machine. Thanks!

Two minor points: 
1. We have a warning while generating: `D:\trackintel\docs\modules\model.rst:2: WARNING: Duplicate explicit target name: "here".`, the solution would be to add one underline `_` after Line 90 in `model.rst`. Solution from [here](https://stackoverflow.com/questions/5464627/how-to-have-same-text-in-two-links-with-restructured-text) (fixed in af0c752).
2. The new docstring for each Accessor is quite long, I propose to add a hyperlink before each Accessor function, e.g., 
```
UsersAccessor
-----------
```
in line 59. The compiled version looks like 
image

Let me know what you think about the second point

henrymartin1 (approved):
Really nice solution!
I ran the tests on my local machine + our continuous integration pipeline covers Windows, Macos and Ubuntu. 
I then installed the version from your fork and tried to access docstrings but that did not work out for me. However, I am not using [this feature](https://www.jetbrains.com/help/pycharm/inline-documentation.html#ws_viewing_inline_documentation) regularly and I am not sure how it behaves. I can get the documentation for functions that I import directly but not for accessors. I am actually not getting any documentation for the accessors.

If it works for you, than I would be fine with merging it.

bifbof:
> Let me know what you think about the second point

I think it looks great and is easier to navigate.

> I then installed the version from your fork and tried to access docstrings but that did not work out for me.

I tested it with IPython and PyCharm and trackintel install via pip. IPython worked fine, but PyCharm didn't show 
 the docstring even before the change (at least on my laptop), so I would say that it is ok to merge it.

henrymartin1:
> I tested it with IPython and PyCharm and installed trackintel via pip. IPython worked fine, but PyCharm didn't show
> the docstring even before the change (at least on my laptop), so I would say that it is ok to merge it.
 fine by me! @hong2223 do you want to have a look at the change you proposed before the merge?

hong2223 (approved):

hong2223:
I checked and this looks very nice on my local machine. I will now merge this PR, thanks for the effort!

PR #275, opened 2021-07-14
BUG: Handle less entries than labels for location identifier.

### LINKED ISSUE #244: Location identification fails for user with a single significant location
There are edge cases where the function `pre_filter_locations` can result in filtering all but 1 location of a user. This leads to the following error in `_freq_assign` in [line 230](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L230):
`ValueError: could not broadcast input array from shape (2,) into shape (1,)`
because it tries to assign 2 labels to 1 location.

Unfortunately, the filter is relatively complex and therefore I failed creating an independent script using artificial data but I have some test data available for download [here](https://1drv.ms/u/s!Ar5qZtQfLW_Mmq95kMzADK9ZhmWQdg?e=OhpPju) that can be used to reproduce the error.

```python
import pandas as pd
import geopandas as gpd
import trackintel as ti


path_to_testdata = r"D:\Geolife\geolife_all_spts.geojson"
spts = gpd.read_file(path_to_testdata)
spts["started_at"] = pd.to_datetime(spts["started_at"], utc=True)
spts["finished_at"] = pd.to_datetime(spts["finished_at"], utc=True)

print("extract locations")
spts, _ = spts.as_staypoints.generate_locations(
    method="dbscan", epsilon=50, num_samples=1, distance_metric="haversine", agg_level="user"
)
spts = ti.analysis.location_identifier(spts, method="FREQ", pre_filter=True)
```

bifbof:
closes #244


tests/analysis/test_location_identification.py
```diff
@@ -358,6 +366,39 @@ def test_leisure_weighting(self):
...
+
+    def test_only_one_leisure_location(self):
+        t_leis = pd.Timestamp("2021-07-14 01:00:00", tz="utc")
+        h = pd.Timedelta("1h")
+        p = Point(0.0, 0.0)  # not used
+        list_dict = [{"user_id": 0, "location_id": 0, "started_at": t_leis, "finished_at": t_leis + h, "g": p}]
+        spts = gpd.GeoDataFrame(data=list_dict, geometry="g")
+        spts.index.name = "id"
+        result = osna_method(spts)
+        spts["activity_label"] = "home"
```
henrymartin1, line 399: Should this be leisure?
As far as I understood The [osna function](https://github.com/mie-lab/trackintel/blob/72643c60c8f99b85435f6f924df2484ab4255b35/trackintel/analysis/location_identification.py#L330) the outcome should actually be leisure, shouldn't it?

henrymartin1: [here](https://github.com/mie-lab/trackintel/blob/72643c60c8f99b85435f6f924df2484ab4255b35/trackintel/analysis/location_identification.py#L285) the mean time is calculated but in the case here this should be 01:30:00, which should be classified as leisure. Maybe you can find what is wrong...

henrymartin1 (approved):
Good work and thanks for the fix! I found a small issue in the tests that you should comment on.

henrymartin1 (changes requested):
see comments

tests/analysis/test_location_identification.py
bifbof: No, the time is in the timeframe "leisure" but as in the paper suggested everything in leisure should attribute to a "home" label (althougth less weighted than "rest").

henrymartin1: Ah ok, thanks for the clarification.

PR #276, opened 2021-07-15
BUG: Index error in `pre_filter_locations`

### LINKED ISSUE #247: Index error in pre_filter_locations
The function `pre_filter_locations` used in `location_identification` sometimes throws the following warning: 
```
UserWarning: Boolean Series key will be reindexed to match DataFrame index.
```
The warning originates from [line 64](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L64) in `location_identification` because `spts` and the boolean series `f` have different length when `spts[f]` is called. 

This error can occur when the staypoints have a non-continuous (e.g., I deleted some users so there were gaps in the index) index. (my index was also unordered but I am not sure if that is important.
If this is the case, then `user_filter` in [line 138](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L135) uses the real index (unordered with gaps) but `loc_filter` in [line 156](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L156) gets a new index that starts from 0 (note the reset index)

Both filters now have different indices but are combined in [line 158](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L158) using `total_filter = user_filter & loc_filter`.
Here it is important to note that the `&` operator in pandas is index-aware. In this case the result is a longer series with the intersection of both indices. See the example below:

```python
a = pd.Series(data=[True, False, True], index=[1, 2, 3], name="val")
b = pd.Series(data=[True, False, True], index=[2,1, 5], name="val")
a&b
```
results in:
```
1    False
2    False
3    False
5    False
Name: val, dtype: bool
```

I have generated the warning using the following code:
```python
import pandas as pd
import geopandas as gpd
import trackintel as ti
import pickle
import warnings
warnings.simplefilter('error')

spts = pickle.load( open( r"geolife_all_stps_locid.p", "rb" ) )
spts = spts.sort_values("user_id")
spts = spts[~spts['user_id'].isin([8, 48, 90, 98, 120, 123, 131, 133, 136, 137, 138, 139, 143, 155])]


spts = ti.analysis.location_identifier(spts, method="FREQ", pre_filter=True)
```
Where `geolife_all_stps_locid.p` are the staypoints with location id of all geolife users. It can be downloaded [here](https://1drv.ms/u/s!Ar5qZtQfLW_Mmq96of601Ws2KLFZew?e=XdiYkY).

bifbof:
closes #247 
Thanks Henry for the thorough bug report.
Fix that both bool Series use same index.

### DIFF
--- a/tests/analysis/test_location_identification.py
+++ b/tests/analysis/test_location_identification.py
@@ -6,6 +6,7 @@ import pandas as pd
 import pytest
 import trackintel as ti
 from geopandas.testing import assert_geodataframe_equal
+from pandas.testing import assert_index_equal
 from shapely.geometry import Point
 from trackintel.analysis.location_identification import (
     _freq_assign,
@@ -105,6 +106,13 @@ class TestPre_Filter:
         default_kwargs["agg_level"] = "unknown"
         with pytest.raises(ValueError):
             pre_filter_locations(example_staypoints, **default_kwargs)
+
+    def test_non_continous_index(self, example_staypoints, default_kwargs):
+        """Test if function works with non-continous index."""
+        # issue-#247
+        example_staypoints.index = [0, 999, 1, 15]
+        f = pre_filter_locations(example_staypoints, **default_kwargs)
+        assert_index_equal(f.index, example_staypoints.index)
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


henrymartin1 (approved):
Nice! Very clean 👍

PR #281, opened 2021-07-20
WIP: ENH: Tour generation

### LINKED ISSUE #263: Implement a generate tour function
Tours represent an integral part of the trackintel data model. However they are not supported yet. A core functionality that we should support is to be able to create tours from existing trips.

A tour is defined as the collection of all consecutive **trips** that are used to return to the starting location. Tours can be nested and a tour that starts and ends at home is called a home-based tour or a journey.
Here is an image that can hopefully provide some clarification.
grafik

NinaWie:
Closes #263
Building up on PR #271 where trips dataframes include a geometry column.

In addition to the `generate_tours` function, I have also added functions to read and write tour dataframes from csvs.
The `generate_tours` function is now designed in the following way:
* **If the the locations are available,** the locations are used to connect trips to a tour. Specifically, if the start point of one trip shares the location of the end point of the previous trip, this is pursued further as a tour candidate.
* **If no locations are available** (only the trips df is required as input and no other preprocessing is necessary), the trips are connected to tours if their start and end points are less than a threshold `max_dist` apart. 

Other parameters can be set, including the maximum time period that a tour can take, and whether gaps (missing trips inbetween) are allowed on the tour. 

Open questions:
* How to determine whether a tour is a journey? I can't find activity labels in the test data tables.
* Is there a maximum time between two consecutive trips (!= overall time of the tour)?
* If the two options for the algorithm seem reasonable (w / wo locations), there is a required column for trackintel tours which might only contain NaNs, namely the `origin_destination_location_id` column. Probably we should either change the algorithm, or make this column optional.


henrymartin1 (changes requested):
I think it will be very hard to review if all the commits of the trip data model pull request are also marked as changes. This happend to me several times in the past but I never found out how to get rid of all the commits that are already merged into the master branch. The only way I could do it was to create a new branch from master and manually transfer the changes. If you were strict with your commits, you can cherry-pick them rather than copying the code by hand.

PR #282, opened 2021-07-21
BUG: fixed 279 + tests

### LINKED ISSUE #279: generate_locations is broken
There is a huge problem in the `generate_locations` method at the moment. The problem comes from [line 122](https://github.com/mie-lab/trackintel/blame/master/trackintel/preprocessing/staypoints.py#L122) because `ret_stps_non_noise_labels` and `ret_stps` have disjunct indices (see line 105 and 106](https://github.com/mie-lab/trackintel/blame/master/trackintel/preprocessing/staypoints.py#L106)). Therefore the sum in line 122 basically appends `ret_stps["location_id"]` to `ret_stps_non_noise_labels`

Here is code to reproduce the problem:
```python

import pandas as pd
import geopandas as gpd
import datetime
import trackintel as ti


def example_staypoints():
    """Staypoints to load into the database."""
    p1 = Point(8.5067847, 47.4)
    p2 = Point(8.5067847, 47.5)
    p3 = Point(8.5067847, 47.6)

    t2 = pd.Timestamp("1971-01-01 05:00:00", tz="utc")
    t3 = pd.Timestamp("1971-01-02 07:00:00", tz="utc")
    one_hour = datetime.timedelta(hours=1)

    list_dict = [
        {"user_id": 0, "started_at": t1, "finished_at": t2, "geom": p1},
        {"user_id": 0, "started_at": t2, "finished_at": t3, "geom": p2},
        {"user_id": 1, "started_at": t3, "finished_at": t3 + one_hour, "geom": p3},
    ]
    stps = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
    stps.index.name = "id"
    assert stps.as_staypoints
    return stps

sp = example_staypoints()
sp2, locs = sp.as_staypoints.generate_locations(method='dbscan', epsilon=30, num_samples=2,
                                                   distance_metric='haversine', agg_level='user')
pd.set_option('display.max_rows', 10)
print(sp)
print(sp2)
```

`sp`
id        user_id ...     geom                     
0         0  ...  POINT (8.50678 47.40000)
1         0  ...  POINT (8.50678 47.50000)
2         1  ...  POINT (8.50678 47.60000)

`sp2`
    
id user_id                started_at  ...                      geom location_id
0       0.0 1971-01-01 00:00:00+00:00  ...  POINT (8.50678 47.40000)        <NA>
1       0.0 1971-01-01 05:00:00+00:00  ...  POINT (8.50678 47.50000)        <NA>
2       1.0 1971-01-02 07:00:00+00:00  ...  POINT (8.50678 47.60000)        <NA>
0       NaN                       NaT  ...                      None        <NA>
1       NaN                       NaT  ...                      None        <NA>
2       NaN                       NaT  ...                      None        <NA>
[6 rows x 5 columns]

henrymartin1:
fixes #279

### DIFF
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -119,7 +119,7 @@ def generate_locations(
             # fill first nan with 0 and create the cumulative sum
             loc_id_offset = loc_id_offset.fillna(0).cumsum()
 
-            ret_stps_non_noise_labels["location_id"] = ret_stps["location_id"] + loc_id_offset
+            ret_stps_non_noise_labels["location_id"] = ret_stps_non_noise_labels["location_id"] + loc_id_offset
             ret_stps = gpd.GeoDataFrame(pd.concat([ret_stps_non_noise_labels, ret_stps_noise_labels]), geometry=geo_col)
             ret_stps.sort_values(["user_id", "started_at"], inplace=True)
 
--- a/tests/preprocessing/test_staypoints.py
+++ b/tests/preprocessing/test_staypoints.py
@@ -1,13 +1,57 @@
+import datetime
 import os
 
-import numpy as np
 import geopandas as gpd
+import numpy as np
 import pandas as pd
+import pytest
 from shapely.geometry import Point
 from sklearn.cluster import DBSCAN
 
 import trackintel as ti
 from trackintel.geogr.distances import calculate_distance_matrix
+
+
+@pytest.fixture
+def example_staypoints():
+    """Staypoints for location generation.
+    Staypoints have non-continous ids and should result in noise and several locations per user.
+    The following staypoint ids should form a location (1, 15), (5,6), (80, 3)
+    The following staypoint ids should be noise (2, 7)
+
+    for agg_level="dataset"
+    The following staypoint ids should form a location (1, 15), (5,6, 80, 3),
+    The following staypoint ids should be noise (2, 7)
+    """
+    p1 = Point(8.5067847, 47.4)
+    p2 = Point(8.5067847, 47.5)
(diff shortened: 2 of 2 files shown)


trackintel/preprocessing/staypoints.py
```diff
@@ -119,7 +119,7 @@ def generate_locations(
             # fill first nan with 0 and create the cumulative sum
             loc_id_offset = loc_id_offset.fillna(0).cumsum()
 
-            ret_stps_non_noise_labels["location_id"] = ret_stps["location_id"] + loc_id_offset
+            ret_stps_non_noise_labels["location_id"] = ret_stps_non_noise_labels["location_id"] + loc_id_offset
```
abcnishant007, line 122: This seems to be a clear bug and pretty much careless! sorry for introducing it! 
I am just trying to understand if we need a new test for this or a minor tweak in one of the other tests checks for the stability of indices?

abcnishant007: Apart from the need for the new tests, the change is fine. I will approve it once you confirm that the separate test was needed @henrymartin1

abcnishant007 (commented):
@henrymartin1 this is ready to merge. Please take a look at my comment and confirm that `the new tests` were intended.

trackintel/preprocessing/staypoints.py
henrymartin1: Don't worry about the error, happens to all of us, that's why the review is so important. And even then we introduce bugs and errors ;-)
Regarding your commment, I don't really understand it. Do want to say that we don't really need the new tests (I mean they are there yet, so 🤷‍♂️) or that we should have more tests?

PR #283, opened 2021-07-21
BUG: Fix tripleg generation

### LINKED ISSUE #280: BUG: index selection in generate_triplegs
generate_triplegs occasionally throws the error `LineStrings must have at least 2 coordinate tuples` even though we filtered out the invalid triplegs in [line 363](https://github.com/mie-lab/trackintel/blob/8f53d0cc6495e53df06554a76adbd66a57dd8e2d/trackintel/preprocessing/positionfixes.py#L363).

This is because in all the `cond_*` functions we deal with records based on their index in the array but not with the df.index values. However, In [line 365](https://github.com/mie-lab/trackintel/blob/8f53d0cc6495e53df06554a76adbd66a57dd8e2d/trackintel/preprocessing/positionfixes.py#L365), we regard `cond_to_remove` as `pfs.index` values, and this will delete the wrong pfs if the `pfs.index` is not continues or not starting from 0. 

The proposed fix would be to replace the `pfs.loc[pfs.index.isin(cond_to_remove), "tripleg_id"] = -1` into `pfs.loc[pfs.index[cond_to_remove], "tripleg_id"] = -1` in Line 365.

A new test for this will follow.

hong2223:
closes #280

### DIFF
--- a/tests/preprocessing/test_positionfixes.py
+++ b/tests/preprocessing/test_positionfixes.py
@@ -49,14 +49,13 @@ def example_positionfixes_isolated():
     """
     Positionfixes with isolated positionfixes.
 
-    User1 have the same geometry but different timestamps.
-    User2 have different geometry and timestamps.
-    User3 have only one isolated positionfixes.
+    User1 has the same geometry but different timestamps.
+    User2 has different geometry and timestamps.
+    User3 has only one isolated positionfix.
     """
     p1 = Point(8.5067847, 47.4)
     p2 = Point(8.5067847, 47.5)
     p3 = Point(8.5067847, 47.6)
-    p4 = Point(8.5067847, 47.6)
 
     t1 = pd.Timestamp("1971-01-01 00:00:00", tz="utc")
     t2 = pd.Timestamp("1971-01-02 01:01:00", tz="utc")
@@ -71,10 +70,10 @@ def example_positionfixes_isolated():
         {"user_id": 1, "tracked_at": t1, "geometry": p1, "staypoint_id": 2},
         {"user_id": 1, "tracked_at": t2, "geometry": p2, "staypoint_id": pd.NA},
         {"user_id": 1, "tracked_at": t3, "geometry": p3, "staypoint_id": pd.NA},
-        {"user_id": 1, "tracked_at": t4, "geometry": p4, "staypoint_id": 3},
+        {"user_id": 1, "tracked_at": t4, "geometry": p3, "staypoint_id": 3},
         {"user_id": 2, "tracked_at": t1, "geometry": p1, "staypoint_id": 4},
         {"user_id": 2, "tracked_at": t2, "geometry": p2, "staypoint_id": pd.NA},
-        {"user_id": 2, "tracked_at": t4, "geometry": p4, "staypoint_id": 5},
+        {"user_id": 2, "tracked_at": t4, "geometry": p3, "staypoint_id": 5},
     ]
     pfs = gpd.GeoDataFrame(data=list_dict, geometry="geometry", crs="EPSG:4326")
     pfs.index.name = "id"
@@ -244,7 +243,7 @@ class TestGenerate_triplegs:
         # regenerate noncontinuous index
         pfs.index = range(0, pfs.shape[0] * 2, 2)
         pfs.index.name = "id"
-        # unorder index
+        # shuffle index
         pfs = pfs.sample(frac=1)
 
         # a warning shall raise due to the duplicated positionfixes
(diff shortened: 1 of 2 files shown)


tests/preprocessing/test_positionfixes.py
```diff
@@ -45,8 +45,14 @@ def example_positionfixes():
 
 
 @pytest.fixture
-def isolated_positionfixes():
-    """Positionfixes with two isolated positionfixes that have the same geometry but different timestamps."""
+def example_positionfixes_isolated():
+    """
+    Positionfixes with isolated positionfixes.
+
+    User1 have the same geometry but different timestamps.
```
henrymartin1, line 52: have -> has

tests/preprocessing/test_positionfixes.py
```diff
@@ -45,8 +45,14 @@ def example_positionfixes():
...
+    User2 have different geometry and timestamps.
+    User3 have only one isolated positionfixes.
```
henrymartin1, line 54: positionfix singular if it is only 1

tests/preprocessing/test_positionfixes.py
```diff
@@ -45,8 +45,14 @@ def example_positionfixes():
...
+    """
```
henrymartin1, line 55: p3 and p4 below are the same. If this is intentional I would rather only use p3 below

tests/preprocessing/test_positionfixes.py
```diff
@@ -228,9 +237,31 @@ def test_temporal(self):
...
 
-    def test_invalid_isolates(self, isolated_positionfixes):
+    def test_noncontinuous_unordered_index(self, example_positionfixes_isolated):
+        """The unordered and noncontinuous index of pfs shall not affect the generate_triplegs() result."""
+        pfs = example_positionfixes_isolated
+
+        # regenerate noncontinuous index
+        pfs.index = range(0, pfs.shape[0] * 2, 2)
+        pfs.index.name = "id"
+        # unorder index
```
henrymartin1, line 247: I would say "shuffle"

tests/preprocessing/test_positionfixes.py
```diff
@@ -228,9 +237,31 @@ def test_temporal(self):
 class TestGenerate_triplegs:
     """Tests for generate_triplegs() method."""
 
-    def test_invalid_isolates(self, isolated_positionfixes):
+    def test_noncontinuous_unordered_index(self, example_positionfixes_isolated):
+        """The unordered and noncontinuous index of pfs shall not affect the generate_triplegs() result."""
```
henrymartin1, line 241: I think it is "unordered" but "non-continuous"

tests/preprocessing/test_positionfixes.py
```diff
@@ -228,9 +237,31 @@ def test_temporal(self):
...
+        warn_string = "The positionfixes with ids .* lead to invalid tripleg geometries."
+        with pytest.warns(UserWarning, match=warn_string):
+            pfs, tpls = pfs.as_positionfixes.generate_triplegs()
+
+        # the index shall be reordered
+        assert np.all(pfs.index == range(0, pfs.shape[0] * 2, 2))
+
+        # only user 1 has generated a tripleg.
+        # user 0 and 2 tripleg has invalid geometry (two identical points and only one point respectively)
+        assert (tpls["user_id"].unique()[0] == 1) and (len(tpls) == 1)
```
henrymartin1, line 260: I think `(tpls["user_id"].unique()[0] == 1)` is quite confusing. How about `(tpls["user_id"].iloc[0] == 1)`

henrymartin1 (changes requested):
Looks good to me! I only have minor spelling stuff

tests/preprocessing/test_positionfixes.py
hong2223: I am not an expert in English, but in the dictionary it says [noncontinuous](https://www.collinsdictionary.com/dictionary/english/noncontinuous)

tests/preprocessing/test_positionfixes.py
hong2223: strange legacy issue. Already changed

tests/preprocessing/test_positionfixes.py
hong2223: the former implies there is only one `user_id` whereas the latter only takes the first record. It is true they are the same under the condition `(len(tpls) == 1)`.

How about `(tpls["user_id"].unique() == [1])`?

henrymartin1: `tpls["user_id"].unique() == [1]` works for me.

henrymartin1 (approved):
Thanks @hong2223 for the quick fix!

PR #284, opened 2021-07-22
ENH: test data guidelines

henrymartin1:


henrymartin1:
I was thinking about this and this is only really a problem if we compare the result of a pipeline to a predefined set of results (e.g., like we do [here](https://github.com/mie-lab/trackintel/blob/27d80e3320fa8046da56d9ebb4c3e6fd4a798c36/tests/preprocessing/test_triplegs.py#L56).
In [this test](https://github.com/mie-lab/trackintel/blob/5042eaeb387a20da2d055b0d8356a6e2770833fe/tests/preprocessing/test_triplegs.py#L74) for checking the coordinates in the trip generation using the result of a preprocessing pipeline is no problem because it is compared to itself. So this should still be true no matter what happens to the staypoint or tripleg generation.

Issue #98, opened 2021-02-04
tests (and test data) missing for (read) tours


svenruf:
function tours_from_gpd is not tested (#93) 
function read_tours_csv is not available. is this correct or should it be added?

hong2223:
I think this is because tours is not yet fully supported. I'll add function signatures to remind us later.

henrymartin1:
> 
> 
> I think this is because tours is not yet fully supported. I'll add function signatures to remind us later.

I think adding function signatures is right thing to do. Our future selfs will thank you for that :D

PR #287, opened 2021-07-26
ENH: Tour generation

### LINKED ISSUE #263: Implement a generate tour function

NinaWie:
Closes #263

In addition to the `generate_tours` function, I have also added functions to read and write tour dataframes from csvs.
The `generate_tours` function is now designed in the following way:
* **If the locations are available,** the locations are used to connect trips to a tour. Specifically, if the start point of one trip shares the location of the end point of the previous trip, this is pursued further as a tour candidate.
* **If no locations are available** (only the trips df is required as input and no other preprocessing is necessary), the trips are connected to tours if their start and end points are less than a threshold `max_dist` apart. 

Other parameters can be set, including the maximum time period that a tour can take, and whether gaps (missing trips inbetween) are allowed on the tour. 

Open questions:
* How to determine whether a tour is a journey? I can't find activity labels in the test data tables.
* Is there a maximum time between two consecutive trips (!= overall time of the tour)?
* If the two options for the algorithm seem reasonable (w / wo locations), there is a required column for trackintel tours which might only contain NaNs, namely the `origin_destination_location_id` column. Probably we should either change the algorithm, or make this column optional.


henrymartin1:
also closes #98

tests/preprocessing/test_trips.py
```diff
@@ -0,0 +1,87 @@
...
+    def test_accessor(self):
+        """Test if the accessor leads to the same results as the explicit function."""
+        # create trips from geolife (based on positionfixes)
+        pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife_long"))
+        pfs, stps = pfs.as_positionfixes.generate_staypoints(
+            method="sliding", dist_threshold=25, time_threshold=5, gap_threshold=1e6
+        )
+        stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
+        pfs, tpls = pfs.as_positionfixes.generate_triplegs(stps)
+        stps, tpls, trips = ti.preprocessing.triplegs.generate_trips(stps, tpls, gap_threshold=15)
```
henrymartin1, line 52: If you have the same preprocessing/data loading steps several times, you increase readability by creating a [fixture](https://docs.pytest.org/en/6.2.x/fixture.html)

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+    --------
+    p1, p2: shapely Point objects
+    max_dist: int
+
+    Returns
+    ------
+    dist_below_thresh: bool
+        indicating whether p1 and p2 are less than max_dist apart
+    """
+    geod = Geod(ellps="CPM")  # other guy used WGS84
```
henrymartin1, line 268: Ah, I don't know if "CPM" is a good choice as ellipse (I actually don't know ;-) ) but distance calculation should be coordinate system aware. Even though we use WGS84 most of the time we also want to support projected (=local) coordinate systems. It is OK to use the pyproj functions for WGS84 coordinate systems but for planar coordinate systems it is easier to use the euclidean distance [provided by shapely](https://shapely.readthedocs.io/en/stable/manual.html#object.distance). Best would be if you use our helper function [check_gdf_crs] for that but I just noticed an [issue](https://github.com/mie-lab/trackintel/issues/288) with this function. We also have a vectorized version of the [haversine distance ](https://github.cohttps://github.com/mie-lab/trackintel/blob/563473b885a91858a8463da9db66353d5adb1089/trackintel/geogr/distances.py#L186/mie-lab/trackintel/blob/master/trackintel/geogr/point_distances.py).
We don't need to support non-planar coordinate systems other than WGS84.

henrymartin1 (commented):
Really cool implementation. Again quite a complex PR that touches many different files. I have finished going through the function once and added many minor comments. 
There are three bigger points that will need some additional work/discussion: 
- I am not sure if it is a good idea to count trips to check with `max_gap_size ` (if I understood this right).
- I think that the test data is currently insufficient because it only covers exotic corner cases
- The implementation of `_check_max_dist` should be independent of the used coordinate system.

[Test data]
At the moment it is hard to really check the core functionality of your algorithm given the test data. I think the geolife_long testdata is not really well suited to test the generate tour algorithm. I used the following code to join the location ids to user_trip_df while debugging: (the code looks very complex because I wanted to avoid reindexing but had to account for missing values...)
```python
user_trip_df.loc[~user_trip_df['origin_staypoint_id'].isna(), 'origin_location_id'] = stps_w_locs.loc[user_trip_df.loc[~user_trip_df['origin_staypoint_id'].isna(), 'origin_staypoint_id'], 'location_id'].values
user_trip_df.loc[~user_trip_df['destination_staypoint_id'].isna(), 'destination_location_id'] = stps_w_locs.loc[user_trip_df.loc[~user_trip_df['destination_staypoint_id'].isna(), 'destination_staypoint_id'], 'location_id'].values
```

Here are the origin_location/destination ids for user_trip_df for user 1 and 2 (every row is a trip). In this view we can directly identify the tours but we can also see that they are very weird because they are both generated by self-loops. 

user_id = 0
grafik
user_id = 1
grafik

I think the best would be to create a new (independent) test data set that covers regular cases in tour generation as well as corner cases. I think a good example were we already did this was for testing the gap detection in the trip generation ([this test](https://github.com/mie-lab/trackintel/blob/563473b885a91858a8463da9db66353d5adb1089/tests/preprocessing/test_triplegs.py#L250). 
For this we created [triplegs](https://github.com/mie-lab/trackintel/blob/master/tests/data/trips/triplegs_gaps.csv) and [staypoints](https://github.com/mie-lab/trackintel/blob/master/tests/data/trips/staypoints_gaps.csv) by hand in an excel table in a way that they cover all the different gap types. We then also created the desired [outcome](https://github.com/mie-lab/trackintel/blob/master/tests/data/trips/triplegs_gaps.csv). To do that we exportet a small portion of the geolife dataset (e.g., geolife_long), opened it in excel and then modified timestamps etc.). I think this manual creation of test cases is the only way to get a reliable coverage of the function without adding a significant portion of the geolife dataset to the framework (bloats the package as it is installed together with the source code). 
It is usually nicer to define the test data directly in the test file as it was done [here](https://github.com/mie-lab/trackintel/blob/563473b885a91858a8463da9db66353d5adb1089/tests/preprocessing/test_positionfixes.py#L48), however the tour generation might be too complex for this (e.g., it might be easier than creating it in excel or similar)

tests/data/geolife_long/tours.csv
```diff
@@ -0,0 +1,3 @@
+id,user_id,started_at,finished_at,origin_staypoint_id,destination_staypoint_id,origin_destination_location_id,journey
+0,1,2008-10-23 10:32:53+00:00,2008-10-24 00:23:03+00:00,10.0,14.0,4,
```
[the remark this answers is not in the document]
NinaWie: As discussed, I moved journey to the optional columns and did not fill it for now (since the label is not provided atm)

tests/io/test_file.py
```diff
@@ -315,4 +315,25 @@ def test_set_index(self):
...
     """Test for 'read_tours_csv' and 'write_tours_csv' functions."""
 
-    pass
+    def test_from_to_csv(self):
+        """Test basic reading and writing functions."""
+        orig_file = os.path.join("tests", "data", "geolife_long", "tours.csv")
+        tmp_file = os.path.join("tests", "data", "tours_test.csv")
+        tours = ti.read_tours_csv(orig_file, index_col="id")
+
+        columns = [
```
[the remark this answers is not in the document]
NinaWie: You're right it's not necessary, I removed it now

trackintel/model/tours.py
```diff
@@ -54,7 +55,7 @@ def to_csv(self, filename, *args, **kwargs):
 
         See :func:`trackintel.io.file.write_tours_csv`.
         """
-        raise NotImplementedError
+        ti.io.file.write_tours_csv(self._obj, filename, *args, **kwargs)
```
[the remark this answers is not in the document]
NinaWie: I added that in the file tests

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+    max_dist=100,
+    min_trips_on_tour=2,
+    max_time=timedelta(days=1),
+    max_gap_size=0,
+    geom_col="geom",
+    print_progress=False,
+):
+    """
+    Generate trackintel-tours from trips
+    - Tours are defined as a collection of trips in a certain time frame that start and end at the same point
```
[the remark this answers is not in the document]
NinaWie: Done

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+
+    # collect tours
+    tours = []
+    # Iterate over trips
+    for i, row in user_trip_df.iterrows():
+        trip_id = row.name  # trip id
+        end_time = row["finished_at"]
+        # print("current candidates", start_candidates)
+
+        # check if there is a gap between the previous and current trip
```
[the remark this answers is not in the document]
NinaWie: Done

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+        if len(start_candidates) > 0:
+            # compare end of last to start of new
+            if stps_w_locs is not None:
+                end_start_at_same_loc = _check_same_loc(
+                    user_trip_df.loc[start_candidates[-1], "destination_staypoint_id"],  # dest. stp of previous trip
```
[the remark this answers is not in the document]
NinaWie: Not at this point. At this point I check whether I there is a spatial gap between the last trip in our list, and the current trip. In other words, if two consecutive trips don't share the location, there is a spatial gap. Only further below in the algorithm I check whether it concludes a tour (with the destination of the current trip). I updateded the comment to make it more clear.

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+                break
+
+        # remove points because they are out of the time window
+        start_candidates = start_candidates[new_list_start:]
+
+    tours_df = pd.DataFrame(tours)
+    return tours_df
+
+
+def _check_same_loc(stp1, stp2, stps_w_locs):
```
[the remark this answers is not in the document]
NinaWie: Added

trackintel/preprocessing/trips.py
NinaWie: Okay I think I should read a bit about coordinate systems, because I don't know much about it tbh :D I changed it to use the trackintel `haversine_dist` function now, is that okay?

NinaWie (commented):
I hope I regarded all your comments and didn't miss anything! I have also changed the tests a lot, and deleted some argumetns to the algorithm as we discussed
Let me know what you think

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+        Maximum time that a tour is allowed to take
+
+    max_gap_size: int, default 0
+        It is possible to allow gaps to occur on the tour, which might be useful to deal with missing data.
+        For example, home-work, supermarket-home would still be detected as a tour when max_gap_size >= 1, although
+        the work-supermarket trip is missing.
+
+    gap_threshold : float, default 15 (minutes)
+        Maximum allowed temporal gap size in minutes. If tracking data is missing for more than
+        `gap_threshold` minutes, then a new trip begins after the gap.
```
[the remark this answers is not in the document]
NinaWie: Sorry this argument was not even used, it's was an error because I copied another docstring. Removed it now

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,322 @@
...
+    """
+    # Two options: either the location IDs for staypoints on the trips are provided, or a maximum distance threshold
+    # between end and start of trips is used
+    if stps_w_locs is not None:
+        assert (
+            "location_id" in stps_w_locs.columns
+        ), "Staypoints with location ID is required, otherwise tours are generated wo location from maximum distance"
+
+    # rename geom column if necessary
+    assert geom_col in trips_inp.columns, "Trips table must be a GeoDataFrame and requires a geomentry column"
```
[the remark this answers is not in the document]
NinaWie: Yes, changed that as well

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,331 @@
...
+
+                nr_gaps = np.sum(np.isnan(np.array(start_candidates[j:])))
+
+                if nr_gaps > max_nr_gaps:
+                    # No tour found, too many gaps inbetween
+                    continue
+                # print("Tour found!", tour_candidate.head())
+
+                # remove trips that were on the tour
+                start_candidates = start_candidates[:j]
```
henrymartin1, line 228: I am not sure if it is safe to change the list you are iterating on during the loop. I would rather not do this. 

I am not sure if you are potentially losing subtours by cutting of the start. Consider this example: 
Home ---trip1--->Work---trip2--->Lunch---trip3--->Work---trip4--->Home

So if you going trip by trip, the first tour are are detecting is trip2 + trip3. Then you would delete trip2 and trip3 of `start_candidates` and the next tour would be only trip1+trip4. I am actually not entirely sure which solution we should prefer but it might be that the big tour should contain all trips performed from home to home.

trackintel/preprocessing/trips.py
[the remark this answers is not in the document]
henrymartin1: 👍

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,331 @@
...
+    --------
+    p1, p2: shapely Point objects
+    max_dist: int
+
+    Returns
+    ------
+    dist_below_thresh: bool
+        indicating whether p1 and p2 are less than max_dist apart
+    """
+    dist = ti.geogr.point_distances.haversine_dist(p1.x, p1.y, p2.x, p2.y)
```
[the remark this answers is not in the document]
NinaWie: Okay thanks a lot for the explanation!
I changed it such that it checks the crs, similarly to what you do here https://github.com/mie-lab/trackintel/blob/0a4945230980359d11fd5481439243cb639d2c99/trackintel/analysis/labelling.py#L128

NinaWie (commented):
I'm not sure what you mean by your comment 
> Add a test with non-continuous unordered indices for trips when using tour generation. This can be achieved by simply replacing the index

The test data I generated already has unordered indices, doesn't it? Now I added a test though that tests the index stability - hope that's what you meant :)

tests/io/test_file.py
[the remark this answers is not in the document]
henrymartin1: 👍 Feel free to resolve conversations if you agree with the comment and have addressed it. You can of course leave it open if you are not sure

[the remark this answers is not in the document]
NinaWie: Ah sure, I resolved a few, and left some where you can maybe double check whether they were addressed sufficiently :)

NinaWie:
Renaming of the variables should be done, let me know if I missed any `stps`. So it's ready for review.

tests/preprocessing/test_trips.py
```diff
@@ -0,0 +1,297 @@
...
+    trips.as_trips
+    return trips, sp_locs
+
+
+@pytest.fixture
+def example_nested_tour(example_trip_data):
+    """Helper function to create a nested trip"""
+    trips, _ = example_trip_data
+
+    # construct trips that lies between trips 6 and 15 and forms a tour on its own
```
henrymartin1, line 125: # construct trips that lie between trips 6 and 15 and form a tour on their own

tests/preprocessing/test_trips.py
```diff
@@ -0,0 +1,297 @@
...
+            time_diff = row["finished_at"] - row["started_at"]
+            assert time_diff > datetime.timedelta(0) and time_diff < max_time
+        # check that all times are taken correctly from the trips table
+        for tour_id, tour_df in trips_out.groupby("tour_id"):
+            gt_start = tour_df.iloc[0]["started_at"]
+            gt_end = tour_df.iloc[-1]["finished_at"]
+            assert gt_start == tours.loc[tour_id, "started_at"]
+            assert gt_end == tours.loc[tour_id, "finished_at"]
+
+    def test_tour_geom(self, example_trip_data):
```
henrymartin1, line 185: nice ;-) This test saves me a bug report :D

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,387 @@
...
+    tours.index = tours.index.astype("int64")
+    trips["tour_id"] = trips["tour_id"].astype("Int64")
+
+    # cleaning
+    # tours.drop(columns=["trips"], inplace=True)
+
+    return trips, tours
+
+
+def _generate_tours_user(
```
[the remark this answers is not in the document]
NinaWie: I added a docstring, most things could be copied anyway from above

NinaWie (commented):
Thanks for the comments! I did the changes and maybe Ye can now check :)

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+            "location_id" in staypoints.columns
+        ), "Staypoints with location ID is required, otherwise tours are generated without location using max_dist"
+        geom_col = None  # not used
+        crs_is_projected = False  # not used
+    else:
+        # if no location is given, we need the trips table to have a geometry column
+        assert isinstance(trips_inp, gpd.geodataframe.GeoDataFrame), "Trips table must be a GeoDataFrame"
+        geom_col = trips_inp.geometry.name
+        # get crs
+        crs_is_projected = ti.geogr.distances.check_gdf_crs(trips_inp.copy())
```
henrymartin1, line 80: Reminder: This function will change its name soon: [PR](https://github.com/mie-lab/trackintel/pull/316)

henrymartin1 (commented):
Reminder: `check_gdf_crs` will change its name soon: [PR](https://github.com/mie-lab/trackintel/pull/316)

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+    # No tours found
+    if len(tours) == 0:
+        return trips, tours
+
+    # index management
+    tours["id"] = np.arange(len(tours))
+    tours.set_index("id", inplace=True)
+
+    # assign tour id to trips
+    tour2trip_map = tours[["trips"]].to_dict()["trips"]
```
hong2223, line 123: Is it possible to do the id assignment using `explode`? like [here ](https://github.com/mie-lab/trackintel/blob/8372d269c9ab2489c3a4ef7a8bf4bb04625127bb/trackintel/preprocessing/triplegs.py#L244). I find that significantly increases readability. Again, we don't need to fix this in this PR :)

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+    # temp_grouped = temp.groupby("id").agg({"tour_id": list}) # other version: provide list of tour ids
+    trips = trips.join(temp, how="left")
+
+    ## dtype consistency
+    # trips id (generated by this function) should be int64
+    tours.index = tours.index.astype("int64")
+    trips["tour_id"] = trips["tour_id"].astype("Int64")
+
+    # cleaning
+    # tours.drop(columns=["trips"], inplace=True)
```
hong2223, line 143: If we don't need this, we can delete this line?

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+        The trips have to follow the standard definition for trips DataFrames
+
+    staypoints : GeoDataFrame (as trackintel staypoints, preprocessed to contain location IDs), default None
+        The staypoints have to follow the standard definition for staypoints DataFrames. The location ID column
+        is necessary to connect trips via locations to a tour. If None, trips will be connected based only on a
+        distance threshold `max_dist`.
+
+    max_dist: float, default 100 (meters)
+        Maximum distance between the end point of one trip and the start point of the next trip on a tour.
+        Note: If `max_nr_gaps > 0` (see below), a tour can contain gaps
```
hong2223, line 172: Maybe see notes in `generate_tours`

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+
+    max_time: Timedelta, default 1 day
+        Maximum time that a tour is allowed to take
+
+    max_nr_gaps: int, default 0
+        Maximum number of spatial gaps on the tour. Use with caution - see notes below.
```
hong2223, line 178: Maybe see notes in `generate_tours`

trackintel/preprocessing/trips.py
hong2223: Do we need to copy here?

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+        geom_col = None  # not used
+        crs_is_projected = False  # not used
+    else:
+        # if no location is given, we need the trips table to have a geometry column
+        assert isinstance(trips_inp, gpd.geodataframe.GeoDataFrame), "Trips table must be a GeoDataFrame"
+        geom_col = trips_inp.geometry.name
+        # get crs
+        crs_is_projected = ti.geogr.distances.check_gdf_crs(trips_inp.copy())
+
+    trips = trips_inp.copy()
```
hong2223, line 83: According to our new convention, the long and short name of trips are both `trips`. That mean this line will become `trips = trips.copy()`, I am not sure if there will be some problems if we wrote this...

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+        "origin_staypoint_id": first_trip["origin_staypoint_id"],
+        "destination_staypoint_id": last_trip["destination_staypoint_id"],
+        "trips": list(temp_tour_stack.index),
+        "location_id": start_loc,
+    }
+
+    return tour_dict_entry
+
+
+def get_trips_grouped(trips, tours):
```
hong2223, line 397: I could not find the usage of this function in our code, so this is a function exposed to the users and maybe not a helper function for us? Then consider moving this function to the beginning of the file. 

Somewhile ago we though it might be good to have a standard for placing functions, [here](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#order-of-functions-in-code)

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,433 @@
...
+
+def _generate_tours_user(
+    user_trip_df,
+    staypoints=None,
+    max_dist=100,
+    max_nr_gaps=0,
+    max_time=timedelta(days=1),
+    geom_col="geom",
+    crs_is_projected=False,
+):
```
hong2223, line 156: I can grasp the basic idea, but my head won't work recognizing the corner cases - -. I'll believe you and Henry until I started using this :).

What could also help is drawing a flowchart for the algorithm, like our old generate_trip function :). 

https://github.com/mie-lab/trackintel/blob/master/docs/_static/tripalgorithm.png

hong2223 (approved):
Nice and complicated. Finally, we have the tour generation function and can test the result of other papers! I have some comments and notes that we don't necessarily need to address in this PR. 
Approved!

hong2223:
I think we need to change the function name for `check_gdf_crs` to let the tests pass.

trackintel/io/from_geopandas.py
```diff
@@ -285,7 +285,7 @@ def read_tours_gpd(
     user_id="user_id",
     started_at="started_at",
     finished_at="finished_at",
-    origin_destination_location_id="origin_destination_location_id",
+    location_id="location_id",
     journey="journey",
     tz=None,
     mapper={},
```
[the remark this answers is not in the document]
NinaWie: Yes, I wanted to leave that for another PR

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+import geopandas as gpd
+import pandas as pd
+import numpy as np
+from tqdm import tqdm
+
+import trackintel as ti
+
+
+def generate_tours(
+    trips_inp,
```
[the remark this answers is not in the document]
NinaWie: I did some renaming, maybe you can check whether the variable names are better now

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+    staypoints=None,
+    max_dist=100,
```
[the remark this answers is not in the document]
NinaWie: Yes good point, I'll try to clarify it in the documentation

trackintel/preprocessing/trips.py
NinaWie: Thanks, I deleted it. No need to copy

trackintel/preprocessing/trips.py
NinaWie: See new commit, I renamed both the `trips` as well as the `trips_inp` variable now

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+
+    kwargs = {
+        "max_dist": max_dist,
+        "max_nr_gaps": max_nr_gaps,
+        "max_time": max_time,
+        "staypoints": staypoints,
+        "geom_col": geom_col,
+        "crs_is_projected": crs_is_projected,
+    }
+    if print_progress:
```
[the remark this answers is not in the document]
NinaWie: Yes I think we should do it in another PR, to finish this one at some point

trackintel/preprocessing/trips.py
NinaWie: We can do that for the trackintel paper ;) I'll put it on my todo list

trackintel/preprocessing/trips.py
NinaWie: Done

trackintel/preprocessing/trips.py
NinaWie: Yes, it's a helper function I wrote because of the problem that we have a n-to-n relation ship between trips and tours (one trip can be part of several tours). I moved it to the top of the file now.

NinaWie (commented):
Thanks for the comments! I did the cleanups and smaller changes you suggested, and we can decide tomorrow whether we tackle the others in this PR

trackintel/preprocessing/trips.py
```diff
@@ -0,0 +1,425 @@
...
+        )
+    else:
+        tours = (
+            trips.groupby(["user_id"], group_keys=False, as_index=False)
+            .apply(_generate_tours_user, **kwargs)
+            .reset_index(drop=True)
+        )
+
+    # No tours found
+    if len(tours) == 0:
```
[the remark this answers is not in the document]
NinaWie: I manually added the columns because I think it makes sense to output the tours dataframe in the correct format. And I added the warning

trackintel/preprocessing/trips.py
NinaWie: Done - it really simplifies the code a lot!

hong2223:
One last thing: I think we need to add the functions into documentation, `read_tours_csv` and `generate_tours`

NinaWie:
So, I implemented the last changes to this PR in the trackintel meeting:
* Regarding the Geolife example that @henrymartin1 posted above, we decided that the column `tour_id` in the outputted trips should be a list, such that the N:N relationship between trips and tours is clear. So now it is clear in the output that a trip can be part of multiple tours (which are nested or overlapping).
* We agreed to start the convention to use String inputs for time arguments, or to allow the user to input a String or a timedelta. The reason is that with floats one would always have to check the documentation whether this specific function means minutes/hours/days. So for this PR I also changed the function to accept either a string (by default "1d") or a timedelta.

Otherwise I think we should merge this and leave other things (paralellizing the group by, implementing read_tours_gdf etc) for another time :) Maybe @henrymartin1 can comment whether you agree to the last changes and the geolife question, and then we can merge.

NinaWie:
Merging this huge PR now! I also updated the documentation.
Let me know if you find any bugs!

PR #290, opened 2021-07-28
BUG: Workaround for `write_locations_postgis` `extent` column.

### LINKED ISSUE #285: Write locations with extent to postgis throws error
Hi, 
I have tried to write locations with extent to a postgis database but I received an error. Here is an example that is very similar to the existing test:

```python
from shapely.geometry import Point, Polygon
import geopandas as gpd
from sqlalchemy import create_engine
import trackintel as ti
from db_login import DSN  # database login information
import numpy as np

engine = create_engine(
    "postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_database}".format(
        **DSN
    )
)

p1 = Point(8.5067847, 47.4)
p2 = Point(8.5067847, 47.5)
p3 = Point(8.5067847, 47.6)

list_dict = [
    {"user_id": 0, "center": p1},
    {"user_id": 0, "center": p2},
    {"user_id": 1, "center": p3},
]
locs = gpd.GeoDataFrame(data=list_dict, geometry="center", crs="EPSG:4326")
locs.index.name = "id"

coords = [[8.45, 47.6], [8.45, 47.4], [8.55, 47.4], [8.55, 47.6], [8.45, 47.6]]
extent = Polygon(coords)
locs["extent"] = extent  # broadcasting
locs["extent"] = gpd.GeoSeries(locs["extent"])  # dtype

locs.as_locations.to_postgis(name='locations', con=engine, schema="yumuv_graph_rep", if_exists='replace')
```
This throws the following error: 

```
psycopg2.errors.InvalidParameterValue: Geometry SRID (0) does not match column SRID (4326)
CONTEXT:  COPY locations, line 1, column extent: "POLYGON ((8.449999999999999 47.6, 8.449999999999999 47.4, 8.550000000000001 47.4, 8.550000000000001 ..."
```

I am honestly not really sure why as it is very similar to the existing test for locations. I could solve this by replacing [line 238](https://github.com/mie-lab/trackintel/blob/master/trackintel/io/postgis.py#L238)
`locations["extent"] = locations["extent"].apply(lambda x: WKTElement(x.wkt, srid=srid))` with the following:
`locations = gpd.GeoDataFrame(_convert_to_ewkb(locations, 'extent', srid), geometry='center', crs=srid)`

`_convert_to_ewkb` is a function from geopandas and can be imported `from geopandas.io.sql import _convert_to_ewkb`. This function is used in Geopandas to prepare the main geometry [link](https://github.com/geopandas/geopandas/blob/master/geopandas/io/sql.py#L402)

bifbof:
closes #285 

Workaround for the bug found by @henrymartin1.
And renaming of the test_postgis file because it annoyed me for quite a while. 

Edit: Never change a running CI-system apparently.


henrymartin1 (approved):
Nice fix! Even better than what I proposed in the issue 😉 I hope that this will fix the mysterious error... 
Other than that I am also annoyed by the file name. I saw that it is hard coded in the CI config file [here](https://github.com/mie-lab/trackintel/blob/82e1459736fa9938acdc2192d06bbf27ddc28061/.github/workflows/CI-tests.yml#L63) so if you change it here as well it should work. 
(It is also mentioned in [contributing.md](https://github.com/mie-lab/trackintel/blame/master/CONTRIBUTING.md#L53))

bifbof:
Cool, thanks for the advice how to change it. :)
The CI file is now changed. I have left the contributing.md as it is for now, because as before I would prefer a permanent link and for that we need a commit with this change first.

henrymartin1:
Finally a real filename ;-)

PR #291, opened 2021-07-28
ENH: Parallel version for generate functions

### LINKED ISSUE #289: Parallelization of generate() function
The current generate() function can be speeded up by parallelization. This can be achieved by passing an optional argument like `n_jobs`, and wrap the operations after `groupby` using a generic parallelized function, for example: 

````python

from joblib import Parallel, delayed

def applyParallel(dfGrouped, func, n_jobs, **kwargs):
    retLst = Parallel(n_jobs=n_jobs)(delayed(func)(group, **kwargs) for _, group in dfGrouped)
    return pd.concat(retLst)

# example to use the function
result = applyParallel(pfs.groupby("userid"),  _generate_staypoints_sliding_user, n_jobs = 1)
````

This mainly relates to `trackintel.preprocessing.positionfixes.generate_staypoints()` (currently the bottleneck function) and `trackintel.preprocessing.staypoints.generate_locations()` functions.

### PR DESCRIPTION
**Pull request by hong2223**
closes #289

The general function `applyParallel()` is defined in `trackintel.preprocessing.util`. The `print_progress` argument is passed to the function for progress monitoring. 

A simple performance comparison of `generate_staypoints()` function on the whole Geolife dataset: before 2100s, current 1000s (using 20 workers; seems that the main part takes 200s and the other 800s is occupied by the post-processing.)

No tests are currently available. I manually compared (`pd.assert_frame_equal()`) the result of before and after the place of change. But integrating this would be a very long test (basically keep the old generate function and compare it against the new one), and I am not sure if this is worth it. On the other hand, passing all of our previous tests is already a type of testing? 
One thing that maybe needs a test is the additional parameter `n_jobs`, but this is directly passed to the Parallel function. 
I am open to suggestions for the tests.

### DIFF
--- a/trackintel/preprocessing/util.py
+++ b/trackintel/preprocessing/util.py
@@ -1,3 +1,8 @@
+import pandas as pd
+from joblib import Parallel, delayed
+from tqdm import tqdm
+
+
 def calc_temp_overlap(start_1, end_1, start_2, end_2):
     """
     Calculate the portion of the first time span that overlaps with the second
@@ -59,3 +64,11 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
         overlap_ratio = temp_overlap / dur.total_seconds()
 
     return overlap_ratio
+
+
+def applyParallel(dfGrouped, func, n_jobs, print_progress, **kwargs):
+    # tqdm.pandas(desc="User staypoint generation")
+    df_ls = Parallel(n_jobs=n_jobs)(
+        delayed(func)(group, **kwargs) for _, group in tqdm(dfGrouped, disable=not print_progress)
+    )
+    return pd.concat(df_ls)
--- a/trackintel/preprocessing/util.py
+++ b/trackintel/preprocessing/util.py
@@ -68,20 +68,23 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
 
 def applyParallel(dfGrouped, func, n_jobs, print_progress, **kwargs):
     """
-    Funtion warpper to parallize funtions after .groupby().
+    Funtion warpper to parallelize funtions after .groupby().
 
     Parameters
     ----------
     dfGrouped: pd.DataFrameGroupBy
-        The groupby object after calling df.groupby("user_id").
+        The groupby object after calling df.groupby(COLUMN).
 
     func: function
         Function to apply to the dfGrouped object, i.e., dfGrouped.apply(func).
 
     n_jobs: int
-        The maximum number of concurrently running jobs.
+        The maximum number of concurrently running jobs. If -1 all CPUs are used. If 1 is given, no parallel
+        computing code is used at all, which is useful for debugging. See
+        https://joblib.readthedocs.io/en/latest/parallel.html#parallel-reference-documentation
+        for a detailed description
 
-    print_progress: boolen
+    print_progress: boolean
         If set to True print the progress of apply.
 
     **kwargs:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: trackintel/preprocessing/util.py [hunk 1]
```diff
@@ -59,3 +64,35 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
         overlap_ratio = temp_overlap / dur.total_seconds()
 
     return overlap_ratio
+
+
+def applyParallel(dfGrouped, func, n_jobs, print_progress, **kwargs):
+    """
+    Funtion warpper to parallize funtions after .groupby().
```
henrymartin1, line 71: parallelize

trackintel/preprocessing/util.py
```diff
@@ -59,3 +64,35 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
...
+
+    Parameters
+    ----------
+    dfGrouped: pd.DataFrameGroupBy
+        The groupby object after calling df.groupby("user_id").
```
henrymartin1, line 76: I think we can leave the description like this but it should work with any groupby object, shouldn't it?

trackintel/preprocessing/util.py
```diff
@@ -59,3 +64,35 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
...
+
+    func: function
+        Function to apply to the dfGrouped object, i.e., dfGrouped.apply(func).
+
+    n_jobs: int
+        The maximum number of concurrently running jobs.
```
henrymartin1, line 82: I really liked your more detailed description for `n_jobs` from above. I think especially the part about what happens if you pass `n_jobs=1`. Maybe you can add it here as well. Copied from above: 
> The maximum number of concurrently running jobs. If -1 all CPUs are used. If 1 is given, no parallel
        computing code is used at all, which is useful for debugging. See
        https://joblib.readthedocs.io/en/latest/parallel.html#parallel-reference-documentation
        for a detailed description

trackintel/preprocessing/util.py
```diff
@@ -59,3 +64,35 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
...
+
+    print_progress: boolen
```
henrymartin1, line 84: boolean

trackintel/preprocessing/util.py
```diff
@@ -59,3 +64,35 @@ def calc_temp_overlap(start_1, end_1, start_2, end_2):
...
+    **kwargs:
+        Other arguments passed to func.
+
+    Returns
+    -------
+    pd.DataFrame:
+        The result of dfGrouped.apply(func)
+    """
+    df_ls = Parallel(n_jobs=n_jobs)(
+        delayed(func)(group, **kwargs) for _, group in tqdm(dfGrouped, disable=not print_progress)
```
henrymartin1, line 96: Hm... I think the `disable=not print_progress` looks/reads a bit weird but I guess there is not much we can change...

henrymartin1 (changes requested):
Cool solution! I really like the idea with the generic utility function. My comments concern only some typos.
Regarding tests, it would be nice to have something simple. wouldn't it be possible to have 1 test per function that calculates the result once using `n_jobs=1` and then compares it to `n_jobs=2`? I think this would be enough (if our CI has two cores available).

trackintel/preprocessing/util.py
hong2223: I changed this from `df.groupby("user_id")` into `df.groupby(COLUMN)`

hong2223:
Thanks for the review, The typos are fixed (together with some typos I found in the framework), and two tests for n_jobs are added. I additionally found two warnings in `test_generate_trips_gap_detection` that are now also fixed. Ready for next round!

henrymartin1 (approved):
Trackintel finally enters the age of parallel computing! Thanks for the update @hong2223 !

PR #296, opened 2021-07-30
BUG: Only numerical data for "user_id".

### LINKED ISSUE #295: generate_trips only accepts numeric user_ids
I think it would be enough to change the aggregation function [here]( https://github.com/mie-lab/trackintel/blob/33d6205475ad8c4a859a9a5883241fe1098997b2/trackintel/preprocessing/triplegs.py#L167) for `user_id` to `first` or something similar that accepts strings.

bifbof:
Closes #295

### DIFF
--- a/trackintel/preprocessing/triplegs.py
+++ b/trackintel/preprocessing/triplegs.py
@@ -162,7 +162,7 @@ def generate_trips(spts, tpls, gap_threshold=15, add_geometry=True):
 
     trips_grouper = spts_tpls_no_act.groupby("temp_trip_id")
     trips = trips_grouper.agg(
-        {"user_id": "first", "started_at": "first", "finished_at": "last", "type": list, "spts_tpls_id": list}
+        {"user_id": "first", "started_at": min, "finished_at": max, "type": list, "spts_tpls_id": list}
     )
 
     def _seperate_ids(row):
--- a/trackintel/preprocessing/triplegs.py
+++ b/trackintel/preprocessing/triplegs.py
@@ -1,11 +1,9 @@
 import warnings
 
+import geopandas as gpd
 import numpy as np
 import pandas as pd
-from shapely import geometry
-from tqdm import tqdm
-from shapely.geometry import MultiPoint, Point
-import geopandas as gpd
+from shapely.geometry import MultiPoint
 
 
 def smoothen_triplegs(triplegs, tolerance=1.0, preserve_topology=True):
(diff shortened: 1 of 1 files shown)


trackintel/preprocessing/triplegs.py
```diff
@@ -164,7 +162,7 @@ def generate_trips(spts, tpls, gap_threshold=15, add_geometry=True):
 
     trips_grouper = spts_tpls_no_act.groupby("temp_trip_id")
     trips = trips_grouper.agg(
-        {"user_id": "mean", "started_at": min, "finished_at": max, "type": list, "spts_tpls_id": list}
+        {"user_id": "first", "started_at": "first", "finished_at": "last", "type": list, "spts_tpls_id": list}
```
henrymartin1, line 165: I am not sure about "first" and "last" for timestamps. Isn't it safer to use "min" and "max" instead of assuming that the first timestamp is also the earliest?

bifbof: Both are safe to use.
In the groupby pandas documentation they write ([link](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html)):
> Groupby preserves the order of rows within each group.

And as we already sorted beforehand we can replace an O(N) with a O(1), but I think the advantage is marginal to be honest. ^^ I just liked it more but if you find it clearer the other way I would change it back.

hong2223: I can understand the "first" in started_at (sorted as you said), but am not sure about the "last" for finished_at. 

We just sorted according to the started_at columns and it could happen that the last finished_at is not the max finished_at. This means of course there is overlapping time and something is wrong with the input data, but I have encountered this in real-world tracking datasets. Perhaps this requires the user to filter these incorrect records first, but as the generate function we cannot assume that the user has done that. 

So I would say we remain "max" for the finished_at. Maybe I am thinking too much, what do you think?

henrymartin1: Yes, I agree. I am also on the better save than sorry side. Also if I think that the solution of @bifbof is probably the smarter way to do it I became quite anxious given all the bug reports ;-)

So we agree on the following?
- "user_id": "first"
- "started_at": min
- "finished_at": max

henrymartin1 (approved):
nice

henrymartin1:
Thanks a lot for the fix @bifbof !

PR #298, opened 2021-08-02
ENH: Default to geometry for `read_*_gpd` functions.

### LINKED ISSUE #278: change default for geom_col argument in read_*_gpd to geometry.name
When we use a `read_*_gpd` (e.g., `read_staypoints_gpd`) with a valid geodataframe it throws an error if the geometry column is not named 'geom'. I think if the geometry dataframe is already valid it should default to the geometry.name argument.

bifbof:
closes #278 

Default of `read_*_gpd` functions now opt for the geometry of the GeoDataFrame if available.
Also added a function `_trackintel_model` that the functions call instead of doing everything themselves.
Added tests for this new function.
Added crs keyword to all `read_*_gpd` functions.
Added "extent" to `read_locations_gpd`.
Added geometry to `read_trips_gpd`
Set default from `mapper` from `{}` to `None` to avoid bugs with mutable datatypes.

~~Two test are still missing~~
~~1. checking if `read_trips_gpd` correctly return either DataFrame or GeoDataFrame.~~
~~2. checking if `read_locations_gpd` can handle an "extent" column.~~

I just wanted to to have a small and clean PR and now we have already 400 changed lines. :]

### DIFF
--- a/tests/io/test_file.py
+++ b/tests/io/test_file.py
@@ -13,7 +13,7 @@ class TestPositionfixes:
         """Test basic reading and writing functions."""
         orig_file = os.path.join("tests", "data", "positionfixes.csv")
         mod_file = os.path.join("tests", "data", "positionfixes_mod_columns.csv")
-        tmp_file = os.path.join("tests", "data", "positionfixes_test.csv")
+        tmp_file = os.path.join("tests", "data", "positionfixes_test_1.csv")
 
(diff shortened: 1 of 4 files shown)


hong2223:
Thanks for this detailed PR. I added some descriptions and moved `_trackintel_model` to the back of the file ([contributing.md](https://github.com/mie-lab/trackintel/blob/master/CONTRIBUTING.md#order-of-functions-in-code))

I will merge this once the tests have passed

PR #302, opened 2021-08-15
ENH: `read_*_gpd` arguments available for `read_*_postgis` functions.

### LINKED ISSUE #277: read_*_postgis functions can only be used with trackintel conform databases
In practice, most databases are not trackintel conform. However, the `read_*_postgis` can only be used with these functions. The reason is that they are passing the result from the query to the `read_*_gpd` functions but without allowing to specify the `mapper` keywords. This makes these functions difficult to use in practice. 
A simple solution would be to allow to pass parameters of the `read_*_gpd` function.

### LINKED ISSUE #301: postgis io functions not ready for new trip datamodel
postgis [reading](https://github.com/mie-lab/trackintel/blob/master/trackintel/io/postgis.py#L252) and [writing](https://github.com/mie-lab/trackintel/blob/master/trackintel/io/postgis.py#L283) functions for trips need to distinguish between dataframes and geodataframes

bifbof:
closes #277 
closes #301 

Added all `from_postgis` kwargs to `read_*_postgis` functions so that the arguments of `read_*_gpd` functions are also accessible via keywords arguments.
Also added distinction between the two trip models.

### DIFF
--- a/tests/io/test_postgis.py
+++ b/tests/io/test_postgis.py
@@ -1,14 +1,14 @@
-"""A copy of the geopandas postgis test to verify the continuous integration"""
 import datetime
 import os
 
 import geopandas as gpd
-import pandas as pd
-import pytest
 from geopandas.testing import assert_geodataframe_equal
+import pandas as pd
 from pandas.testing import assert_frame_equal
-from shapely.geometry import LineString, Point, Polygon
+import pytest
 import sqlalchemy
+from shapely.geometry import LineString, MultiPoint, Point, Polygon
+
 import trackintel as ti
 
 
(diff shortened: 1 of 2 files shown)


tests/io/test_postgis.py
```diff
@@ -470,11 +519,68 @@ def test_read(self, example_trips, conn_postgis):
...
+        finally:
+            del_table(conn, table)
+
+    def test_write_with_geometry(self, example_trips, conn_postgis):
+        """Test if write of trips with geometry creates correct schema in database."""
+        mp1 = MultiPoint([(0.0, 0.0), (1.0, 1.0)])
+        mp2 = MultiPoint([(2.0, 2.0), (3.0, 3.0)])
+        trips = gpd.GeoDataFrame(example_trips, copy=True, geometry=[mp1, mp2, mp2], crs="EPSG:4326")
+        conn_string, conn = conn_postgis
+        table = "positionfixes"
```
henrymartin1, line 554: I think the table name doesn't really matter but this should probably be "trips"

tests/io/test_postgis.py
```diff
@@ -470,11 +519,68 @@ def test_read(self, example_trips, conn_postgis):
...
+        mp2 = MultiPoint([(2.0, 2.0), (3.0, 3.0)])
+        trips = gpd.GeoDataFrame(example_trips, copy=True, geometry=[mp1, mp2, mp2], crs="EPSG:4326")
+        conn_string, conn = conn_postgis
+        table = "trips"
+        sql = f"SELECT * FROM {table}"
+        geom_col = trips.geometry.name
+
+        try:
+            trips.as_trips.to_postgis(table, conn_string)
+            locs_db = ti.io.read_trips_postgis(sql, conn_string, geom_col=geom_col, index_col="id")
```
henrymartin1, line 579: should be `trips_db`

henrymartin1 (approved):
Great contribution, really thorough work!

PR #303, opened 2021-08-15
BUG: fixed duplicate rows when adding geolife mode labels

### LINKED ISSUE #273: BUG: Duplication of certain rows in `geolife_add_modes_to_triplegs`
Running `geolife_add_modes_to_triplegs` returns a dataframe with more rows out than in (duplicates).
The `tpls_id_mode` already has an index with duplicates, so it is just passed on when joined to the `tpls` on line 257. https://github.com/mie-lab/trackintel/blob/dc9c6efe28ec0b736ee8da5a243cb86b8b83eeb2/trackintel/io/dataset_reader.py#L257

Tested with user 52 from the geolife dataset.
````python
import trackintel as ti
from trackintel.io.dataset_reader import read_geolife, geolife_add_modes_to_triplegs

pfs, mode_labels = read_geolife(r".\geolife_user52", print_progress=True)

# generate triplegs
pfs, spts = pfs.as_positionfixes.generate_staypoints(time_threshold=5.0, print_progress=True)
pfs, tpls_pre = pfs.as_positionfixes.generate_triplegs(spts)
tpls_post = geolife_add_modes_to_triplegs(tpls_pre, mode_labels)

assert len(tpls_pre) == len(tpls_post)

````

### PR DESCRIPTION
**Pull request by henrymartin1**
fixes #273

### CODE REVIEW

File: trackintel/io/dataset_reader.py [hunk 1]
```diff
@@ -253,7 +258,13 @@ def geolife_add_modes_to_triplegs(
...
         tpls["mode"] = np.nan
     else:
-        tpls_id_mode = pd.DataFrame(tpls_id_mode_list).set_index("id")
+        tpls_id_mode = pd.DataFrame(tpls_id_mode_list)
+
+        # chose label with highest overlap
+        tpls_id_mode = tpls_id_mode.sort_values(by=["id", "ratio"])
+        # keep last (df sorted ascending)
+        tpls_id_mode = tpls_id_mode.drop_duplicates(subset="id", keep="last").set_index("id")
+
```
NinaWie, line 267: I assume it's on purpose that the `ratio` column is not dropped, so it is part of the tpls dataframe that is returned? Maybe it should be renamed then, e.g. to `ratio_mode_overlap`, because it could be confusing to have a column `ratio` in the output of a function that's supposed to assign a mode.

NinaWie (commented):
Looks good to me. I had one question, feel free to resolve.
Apart from that - to be as picky as you guys ;) - a test could probably be added where you simply add an overlapping label [here](https://github.com/mie-lab/trackintel/blob/b9510e37ff4f6e39288785738b60a0dc6a56f9ec/tests/io/test_dataset_reader.py#L148)

trackintel/io/dataset_reader.py
henrymartin1: dropped

henrymartin1:
Added test and dropped column

NinaWie:
Nice! Looks good to me, merging now

PR #306, opened 2021-08-16
BUG: fixed boundary problem for linestrings with same start and end l…

### LINKED ISSUE #300: generate_trips fails for linestrings with same start and end point
`generate_trips` calls the boundary property of a linestring [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/triplegs.py#L248) and [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/triplegs.py#L254). Linestrings with the same start and end point are considered valid by shapely but do not have a boundary. E.g.

```python
print(LineString([(0, 0), (1, 1), (0, 0)]).boundary)
>>>MULTIPOINT EMPTY
```

Here is a test to reproduce the error:
```python
def test_generate_trips_modified(self):
    """Test if we can generate the example trips based on example data."""
    # load pregenerated trips
    trips_loaded = ti.read_trips_csv(os.path.join("tests", "data", "geolife_long", "trips.csv"), index_col="id")

    # create trips from geolife (based on positionfixes)
    pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife_long"))
    pfs, stps = pfs.as_positionfixes.generate_staypoints(
        method="sliding", dist_threshold=25, time_threshold=5, gap_threshold=1e6
    )
    stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
    pfs, tpls = pfs.as_positionfixes.generate_triplegs(stps)

    # generate trips and a joint staypoint/triplegs dataframe
    from shapely.geometry import LineString
    tpls.loc[0, 'geom'] = LineString([(0, 0), (1, 1), (0, 0)]) # modify first geometry
    stps, tpls, trips = generate_trips(stps, tpls, gap_threshold=15)
    trips = trips[
        ["user_id", "started_at", "finished_at", "origin_staypoint_id", "destination_staypoint_id", "geom"]
    ]
    # test if generated trips are equal
    assert_geodataframe_equal(trips_loaded, trips)
```

A fix could be to use 
```python
first = Point(line.coords[0])
last = Point(line.coords[-1])
```
instead of the boundary argument

NinaWie:
closes #300 

Simple fix for issue 300, using `coords` instead of `boundary`. Is it necessary to add an extra test for this?

### DIFF
--- a/tests/preprocessing/test_triplegs.py
+++ b/tests/preprocessing/test_triplegs.py
@@ -14,6 +14,29 @@ import trackintel as ti
 from trackintel.preprocessing.triplegs import generate_trips
 
 
+@pytest.fixture
+def example_triplegs():
+    """Generate input data for trip generation from geolife positionfixes"""
+    pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife_long"))
+    pfs, stps = pfs.as_positionfixes.generate_staypoints(method="sliding", dist_threshold=25, time_threshold=5)
+    stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
+    pfs, tpls = pfs.as_positionfixes.generate_triplegs(stps)
+    return stps, tpls
+
+
+@pytest.fixture
+def example_triplegs_higher_gap_threshold():
+    """Generate input data for trip generation, but with a higher gap threshold in stp generation"""
+    # create trips from geolife (based on positionfixes)
+    pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife_long"))
+    pfs, stps = pfs.as_positionfixes.generate_staypoints(
+        method="sliding", dist_threshold=25, time_threshold=5, gap_threshold=1e6
+    )
+    stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
+    pfs, tpls = pfs.as_positionfixes.generate_triplegs(stps)
+    return stps, tpls
+
+
 class TestSmoothen_triplegs:
     def test_smoothen_triplegs(self):
         tpls_file = os.path.join("tests", "data", "triplegs_with_too_many_points_test.csv")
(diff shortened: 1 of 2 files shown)


henrymartin1 (changes requested):
Thanks for the quick fix!
If the test is really necessary always depends. If we find a bug or corner case it is always better to add an extra test for it so that we are safe in the future. Sometimes the required effort is just too high. In this case I would say that the effort is reasonable as you can take the example from the issue.

tests/preprocessing/test_triplegs.py
```diff
@@ -356,6 +356,31 @@ def test_only_staypoints_in_trip(self):
         assert (tpls_["trip_id"] == 0).all()
         assert len(trips) == 1
 
+    def test_loop_linestring_case(self):
+        # load pregenerated trips
```
henrymartin1, line 360: docstring missing

tests/preprocessing/test_triplegs.py
```diff
@@ -356,6 +356,31 @@ def test_only_staypoints_in_trip(self):
...
+        trips_loaded = ti.read_trips_csv(os.path.join("tests", "data", "geolife_long", "trips.csv"), index_col="id")
+
+        # create trips from geolife (based on positionfixes)
+        pfs, _ = ti.io.dataset_reader.read_geolife(os.path.join("tests", "data", "geolife_long"))
+        pfs, stps = pfs.as_positionfixes.generate_staypoints(
+            method="sliding", dist_threshold=25, time_threshold=5, gap_threshold=1e6
+        )
+        stps = stps.as_staypoints.create_activity_flag(time_threshold=15)
```
henrymartin1, line 368: I think we can delete the activity flag

tests/preprocessing/test_triplegs.py
```diff
@@ -356,6 +356,31 @@ def test_only_staypoints_in_trip(self):
...
+        # add a tripleg with same start as end
+        tpls.loc[0, "geom"] = LineString([(0, 0), (1, 1), (0, 0)])  # modify first geometry
+        # generate trips and a joint staypoint/triplegs dataframe
+        stps, tpls, trips = ti.preprocessing.triplegs.generate_trips(stps, tpls, gap_threshold=15)
+        trips = trips[
+            ["user_id", "started_at", "finished_at", "origin_staypoint_id", "destination_staypoint_id", "geom"]
+        ]
+
+        # test if generated trips are equal
+        assert_geodataframe_equal(trips_loaded.loc[1:], trips.loc[1:])
```
henrymartin1, line 380: I think this is fairly redundant with the other tests, so maybe we could delete it to have this test more concise. 
The goal is rather to make sure that it has no error and that the start is (0,0).
We are actually quite lucky that the first trip has a missing location ;-)

I think it would this would make the test more concise and easier to read

henrymartin1 (changes requested):
I have added some minor comments to the test. I think we can make it a bit more minimalistic and kick some of the stuff that we test in the other tests anyways.

tests/preprocessing/test_triplegs.py
NinaWie: This gives an error ```AssertionError: staypoints need the column 'activities' to be able to generate trips```
So I think we need to keep that

NinaWie:
I cleaned up the new test as you suggested, and I actually took the opportunity to refactor the whole tripleg test class because there was a lot of redundant code, which could be removed using a pytest fixture. Let me know if there is any issue with this change.
Maybe we should keep in mind that at some point we should change the test data to simulated data, since the test data now relies on staypoint and tripleg generation and activity labeling, so it's very prone to errors (and makes the test slow).

henrymartin1 (approved):
Awesome! Thanks for test and especially for cleaning the code! Much needed and very much appreciated 😃

PR #311, opened 2021-08-18
BUG: Explode function now only called on dataframes.

### LINKED ISSUE #305: trips with geometries cause error in id assignment
The problem is that geopandas overrides the explode method of pandas (see issue [here](https://github.com/geopandas/geopandas/issues/2061)). This causes a problem [here](https://github.com/geopandas/geopandas/issues/2061) and [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/preprocessing/triplegs.py#L270) as we want to access the original pandas explode method. 
I think the simplest fix would be to call the dataframe constructor before using explode. It is not too expensive and then we don't have to wait until the geopandas issue is fixed and we are not limited to only new geopandas versions.
E.g.:
`temp = pd.DataFrame(trips).explode("tpls")`
`temp = pd.DataFrame(trips).explode("spts")`

bifbof:
Closes #305 

I solved the problem differently than we discussed. 
The GeoDataFrame is only created in this `if add_geometry` block, so I moved the block after using the exploding, that way we only use the DataFrame variant.
Creating a test for this is a bit difficult because we need to specify the geopandas version. So I just leave it without one.

### DIFF
--- a/trackintel/preprocessing/triplegs.py
+++ b/trackintel/preprocessing/triplegs.py
@@ -237,6 +237,19 @@ def generate_trips(spts, tpls, gap_threshold=15, add_geometry=True):
         axis=1,
     )
 
+    # now handle the data that is aggregated in the trips
+    # assign trip_id to tpls
+    temp = trips.explode("tpls")
+    temp.index = temp["tpls"]
+    temp = temp[temp["tpls"].notna()]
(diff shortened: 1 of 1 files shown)


henrymartin1 (approved):
Ok, it took me a while to get it. I was looking about a minute and compared character by character to see what you have changed until I got that you just moved it. Very nice fix, thanks!

PR #313, opened 2021-08-20
CLN: Fix names for trackintel datatypes

### LINKED ISSUE #214: Fix official abbrevations for trackintel datatypes
In the code we are often using abbrevations to name variables we had this [discussion a while ago in a pull request](https://github.com/mie-lab/trackintel/pull/155#discussion_r594355723) but we never resolved it.

I think at the moment we commonly use: 
- triplegs: tpls
- positionfixes: pfs
- staypoints: spts vs stps

It would be nice to agree on a set of abbrevations and write it into CONTRIBUTING.md.

Personally I am in favor of spts as it reminds me of points which is often shortend as **pts**.
I don't really like tpls because to me it seems as if we would just pick random letters from the word *triplegs* but I can live with it as it seems quite consitent in the framework.

hong2223:
A huge PR that closes #214.

Only changing of variable names, no change in any functional code. We get rid of `stps` and `spts` :).
The changes are as follows:
- `stps` and `spts` all changed to `sp`.
- convention in function input for generate functions: now all functions accept the full name of datatypes as input and return the short name datatype, for example `pfs, tpls = generate_triplegs(positionfixes=pfs, staypoints=sp)`
- our naming convention is added to `contributing.md`


hong2223:
I will now merge this and check whether there will be further changes in the main branch

PR #315, opened 2021-08-20
BUG: Fix loop

### LINKED ISSUE #304: Possibility for infinite loop in tracking quality function
The [tracking_quality](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/tracking_quality.py#L7) ends up in a infinite loop ([while loop in sub structure](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/tracking_quality.py#L7)) for any entry `started_at` > `finished_at`.
I think the best would be to assert that all `started_at` <= `finished_at` in the beginning and give a reasonable error message.

### LINKED ISSUE #307: temporal_tracking_quality ends in infinite loop if start and end is at midnight
I think the problem is the comparison [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/tracking_quality.py#L274). If start and end are the same and both are at midnight they end up on different days. We could check if the first one is smaller than the second one rather than check for equality. This might be a corner case but we should avoid infinite loops at all costs.

Example code (results in infinite loop!)
```python
import datetime
import geopandas as gpd
import pandas as pd

from trackintel.analysis.tracking_quality import temporal_tracking_quality


p1 = Point(8.5067847, 47.4)
t2 = pd.Timestamp("1971-01-01 00:00:00", tz="utc")

one_hour = datetime.timedelta(hours=1)

list_dict = [
    {"user_id": 0, "started_at": t1, "finished_at": t2, "geom": p1},
]
sp = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
sp.index.name = "id"
print("start tracking quality")
quality = temporal_tracking_quality(sp, granularity="day")
print(quality)
```

hong2223:
closes #307 and closes #304

### DIFF
--- a/tests/analysis/test_tracking_quality.py
+++ b/tests/analysis/test_tracking_quality.py
@@ -1,7 +1,11 @@
 import os
+import datetime
 
 import pytest
 import pandas as pd
+import geopandas as gpd
+from shapely.geometry import Point
+
 import trackintel as ti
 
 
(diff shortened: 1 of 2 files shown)


henrymartin1 (commented):
nice fix! Nothing to add, merged on first try 😃

PR #316, opened 2021-08-20
ENH: check whether crs is planer

### LINKED ISSUE #288: Helper function check_gdf_crs assumes planar crs when not in WGS84
Our helper function [check_gdf_crs](https://github.com/mie-lab/trackintel/blob/563473b885a91858a8463da9db66353d5adb1089/trackintel/geogr/distances.py#L227) assumes that any coordinate system that is not WGS84 is planar. This might often be the case in practice but it is not correct.
We should add a check if it is really planar. E.g., using the [check_geographic_crs](https://github.com/geopandas/geopandas/blob/2394bf052a261fb0e5e6d0f8ed9c018b9ebaace7/geopandas/array.py#L338) function from gepandas.

hong2223:
closes #288

### DIFF
--- a/trackintel/geogr/distances.py
+++ b/trackintel/geogr/distances.py
@@ -212,21 +212,22 @@ def check_gdf_crs(gdf, transform=False):
     >>> check_gdf_crs(triplegs, transform=False)
     """
     if_planer = False
-    if gdf.crs is None:
-        # projection is not defined
+    if gdf.crs is None:  # projection is not defined
         if_planer = False
         if transform:
             gdf.crs = "EPSG:4326"
         else:
             warnings.warn("Your data is not projected.")
 
-    elif gdf.crs == "EPSG:4326":
-        # if projection is defined as WGS84
+    elif gdf.crs == "EPSG:4326":  # if projection is defined as WGS84
         if_planer = False
 
-    else:
-        # if projection is defined but not as WGS84
-        if_planer = True
+    else:  # if projection is defined but not as WGS84
+        if gdf.crs.is_geographic:  # if projection is a geographic crs
+            if_planer = False
+        else:
+            if_planer = True
+
         if transform:
             if_planer = False
             gdf = gdf.to_crs("EPSG:4326")
--- a/tests/geogr/test_distances.py
+++ b/tests/geogr/test_distances.py
@@ -1,13 +1,18 @@
 import os
-from math import radians
+import datetime
 
 import geopandas as gpd
+import pandas as pd
 import numpy as np
 import pytest
+
+from math import radians
 from shapely import wkt
 from shapely.geometry import LineString, MultiLineString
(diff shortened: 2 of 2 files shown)


trackintel/geogr/distances.py
```diff
@@ -212,21 +212,22 @@ def check_gdf_crs(gdf, transform=False):
     >>> check_gdf_crs(triplegs, transform=False)
     """
     if_planer = False
```
henrymartin1, line 214: I would rename this to `is_planer`

trackintel/geogr/distances.py
```diff
@@ -212,21 +212,22 @@ def check_gdf_crs(gdf, transform=False):
...
-    if gdf.crs is None:
-        # projection is not defined
+    if gdf.crs is None:  # projection is not defined
         if_planer = False
```
henrymartin1, line 216: this seems unnecessary as it was just initialized as False

trackintel/geogr/distances.py
```diff
@@ -212,21 +212,22 @@ def check_gdf_crs(gdf, transform=False):
...
-    else:
-        # if projection is defined but not as WGS84
-        if_planer = True
+    else:  # if projection is defined but not as WGS84
+        if gdf.crs.is_geographic:  # if projection is a geographic crs
+            if_planer = False
+        else:
+            if_planer = True
+
         if transform:
```
henrymartin1, line 231: I would change the indentation. At the moment it would not be transformed if gdf.crs is None but transform is true.

trackintel/geogr/distances.py
henrymartin1: ah and change all occurrences of planer to plan**a**r

henrymartin1 (changes requested):
I had some comments and I would change the docstring to:
"""
    Check if a GeoDataFrame has a planar/projected coordinate system. 

    Optionally transform a GeoDataFrame into WGS84.
    [...]"""

Actually when reading this we could change the function name to something like "check_gdf_planar" or similar.

trackintel/geogr/distances.py
hong2223: I think transform is only valid in the else statement (# if projection is defined but not as WGS84). 

If gdf.crs is None but transform is true, the function executes line 218, where the crs is defined as 84, here transform is an alias of define/set

trackintel/geogr/distances.py
hong2223: I would keep this and other `is_planar = False` statements for readability. 

Another option is to delete the initialization (Line 214), as each branch defines is_planar, but this also hurts a bit the readability

hong2223:
I changed the function name, the doctrings and the variable name, Ready for the next round!

trackintel/geogr/distances.py
henrymartin1: 👍

PR #317, opened 2021-08-20
BUG: Fix empty generation

### LINKED ISSUE #267: trackintel generate functions are not allowed to return empty dataframes
At the moment it is not possible to call our generate functions when the result would be empty. 
I think it is an OK use-case that you try to generate a trackintel object (e.g., by using generate_triplegs) without knowing if there are valid triplegs in the set of positionfixes. 

The reason for this behavior is that we usually call the accessor for validation within the generator function e.g. `assert tpls.as_triplegs` and the accessor asserts that the dataframe is not empty. E.g., `assert obj.shape[0] > 0, "Geodataframe is empty with shape: {}".format(obj.shape)`

Possible ways to fix this would be to not call the accessor for validation in the generator functions (maybe replace it with a validation function?) or change the validation in the accessor.

Here is a code example: 

```python
import pandas as pd
import geopandas as gpd
import trackintel as ti

p1 = Point(8.5067847, 47.4)
p2 = Point(8.5067847, 47.5)
p3 = Point(8.5067847, 47.6)

t2 = pd.Timestamp("1971-01-02 01:01:00", tz="utc")
t3 = pd.Timestamp("1971-01-02 01:02:00", tz="utc")
t4 = pd.Timestamp("1971-01-01 03:00:00", tz="utc")

list_dict = [
    {"user_id": 0, "tracked_at": t1, "geometry": p1, "staypoint_id": 0},
    {"user_id": 0, "tracked_at": t2, "geometry": p2, "staypoint_id": pd.NA},
    {"user_id": 1, "tracked_at": t3, "geometry": p2, "staypoint_id": pd.NA},
    {"user_id": 1, "tracked_at": t4, "geometry": p3, "staypoint_id": 1}
]
pfs = gpd.GeoDataFrame(data=list_dict, geometry="geometry", crs="EPSG:4326")
pfs.index.name = "id"
assert pfs.as_positionfixes

pfs, tpls = pfs.as_positionfixes.generate_triplegs()


```

hong2223:
closes #267.

I added the checks to ensure we check the validity only if the generated df has record in it, otherwise a warning will be raised. This is implemented for `generate_staypoints`, `generate_triplegs` and `generate_locations` functions. 

For `generate_trips` I think this check is not needed as (in the worst case), each tripleg will be one trip, and the user would not input an empty tripleg into the function


tests/preprocessing/test_positionfixes.py
```diff
@@ -110,15 +121,15 @@ def test_duplicate_pfs_warning(self, example_positionfixes):
         pfs_duplicate_all = pfs_duplicate_loc.copy()
         pfs_duplicate_all.loc[0, "tracked_at"] = pfs_duplicate_all.loc[1, "tracked_at"]
 
+        warn_string = ".* duplicates were dropped from your positionfixes."
         with pytest.warns(None) as record:
```
NinaWie, line 125: Just a general questions for understanding: pytest.warns(None) means that all warnings are catched, while pytest.warns(UserWarning) means that you only check for warnings that were created in this package?

NinaWie (approved):
Very nice! I can't find any problem, and only had one general question about pytest, so I'll merge this now 👍

tests/preprocessing/test_positionfixes.py
hong2223: Yes. pytest.warns(None) captures warnings of all types, see [here](https://docs.pytest.org/en/6.2.x/warnings.html#recwarn). pytest.warns(UserWarning) only captures the warning type `UserWarning`. I think when raising a warning, we can define the type of the warning, and `UserWarning` is the default type if not specified, see [here](https://docs.python.org/3/library/warnings.html#warnings.warn)

PR #320, opened 2021-08-21
TST: add test for duplicate_matching

hong2223:


NinaWie:
Thanks for the changes! This is ready to be merged, right? At least from my side it looks good now :)

PR #322, opened 2021-08-25
BUG: Maintain index dtype for sp.

### LINKED ISSUE #292: generate_trips changes index name of triplegs
The problem seems to be the join in [line 226](https://github.com/henrymartin1/trackintel/blob/27d80e3320fa8046da56d9ebb4c3e6fd4a798c36/trackintel/preprocessing/triplegs.py#L226).

bifbof:
closes #292 

Minor bug in generating trips, when column contains None the dtype changes.
Added a line to revert that.


hong2223:
Thanks for taking care of this. Seems good and ready to merge!

PR #323, opened 2021-08-26
ENH: embed version number 

abcnishant007:
Issue:
`ti.__version__` was not printing anything; It was returning a pointer to a function whereas `__version__.__version__` and `__version__.VERSION` were printing the version name.

Test: 
Previously we had: 
Screenshot 2021-08-26 at 9 43 17 PM

After this fix, we have:
Screenshot 2021-08-26 at 9 17 21 PM

Soluton based on [this]( https://stackoverflow.com/a/16084844/3896008 ) stackoverflow answer from users: [sorin](https://stackoverflow.com/users/99834/sorin) and [nico-schlömer](https://stackoverflow.com/users/353337/nico-schl%c3%b6mer)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ else:
 # Where the magic happens:
 setup(
     name=NAME,
-    version=about["__version__"],
+    version=__version__,
     description=DESCRIPTION,
     long_description=long_description,
     long_description_content_type="text/markdown",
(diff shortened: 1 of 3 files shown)


hong2223:
Thanks for the fix! looks nice and standard.

abcnishant007:
noticed that you had to change the PR name, sorry for that; :P
I think if I keep the first commit of the branch as the name we want for the pull request, that should ensure proper name.

PR #327, opened 2021-08-31
DOC: fix error in documentation of staypoint merging

NinaWie:
Thanks for spotting this error in the documentation!
Let me know if it appears anywhere else; I couldn't find anything at this point.


hong2223:
👍

PR #328, opened 2021-09-08
ENH: return absolute datetime in quality

### LINKED ISSUE #258: How to use temporal_tracking_quality to delete bad tracking days
Hey, 
I have a GPS tracking dataset and want to filter days that have a low tracking coverage. I noticed that the (pretty nice) function `ti.analysis.tracking_quality.temporal_tracking_quality` does exactly what I want ;-)
Except that I don't know how I can use the results for filtering. The problem is that the function does not return a date but a relative count of days since the start of tracking (I assume). Therefore I can not use the result to identify bad tracking days in my original dataset. (see the screenshot below). 
I think it would be better if the function would return a datetime index instead of a relativ day count. This is done automatically when using a pandas datetime grouper as in the modal split function [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/modal_split.py#L69). Maybe using that would be an option.

image

abcnishant007:
Closes #258 

Applied fix:
When the granularity is either `"day"` or `"week"`, the `quality` dataframe has an extra column called `"date"` or `"week_number"` respectively.
 Using the `"date"` and `"week_number"` columns, the trackintel user can estimate which specific ` "date"` or `"week_number"` were found to have poor tracking quality


trackintel/analysis/tracking_quality.py
```diff
@@ -50,7 +50,12 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
 
     The tracking quality of each user is calculated based on his or her own tracking extent.
     For granularity = ``day`` or ``week``, the quality["day"] or quality["week"] column displays the
-    time relative to the first record in the entire dataset.
+    time relative to the first record in the entire dataset. In addition to the relative values of
```
hong2223, line 53: Maybe start a new paragraph here

trackintel/analysis/tracking_quality.py
```diff
@@ -87,23 +92,27 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
...
         quality = _get_all_quality(df, raw_quality, granularity)
+        quality["date"] = quality["day"].map(dict_day_to_date)
 
     elif granularity == "week":
         # split records that span several days
         df = _split_overlaps(df, granularity="day", max_iter=max_iter)
         # get the tracked week relative to the first day
         start_date = df["started_at"].min().date()
         df["week"] = df["started_at"].apply(lambda x: (x.date() - start_date).days // 7)
+        dict_week_num = dict(zip(df["week"], df["started_at"].apply(lambda x: pd.to_datetime(x).week)))
```
hong2223, line 109: I have a warning raised here:

> FutureWarning: weekofyear and week have been deprecated, please use DatetimeIndex.isocalendar().week instead, which returns a Series.  To exactly reproduce the behavior of week and weekofyear and return an Index, you may call pd.Int64Index(idx.isocalendar().week)

Also, the week_number is in float format, I think it should be int

hong2223 (changes requested):
Thanks. I had a look at the added lines and they seem to work as expected. 

Is it possible to add a simple test for the absolute difference? For days we can just take the last tracking record's day and assert it is the same as the last absolute date in the quality. Similar ideas for the weeks.

Also, for the weeks, which one do you think would be better, the week number (now) or the start date of the week (i.e., the date of the start Monday.)?

abcnishant007:
@hong2223  Thanks for taking a look,

"Is it possible to add a simple test for the absolute difference? For days we can just take the last tracking record's day and assert it is the same as the last absolute date in the quality. Similar ideas for the weeks."
- Yes, will do that

"Also, for the weeks, which one do you think would be better, the week number (now) or the start date of the week (i.e., the date of the start Monday.)?"
- Both of them carry the same amount of information, so I have no specific preference. We can return both if that can be useful.

trackintel/analysis/tracking_quality.py
abcnishant007: Warning removed, week number changed to `int`.

trackintel/analysis/tracking_quality.py
abcnishant007: done

abcnishant007:
@hong2223  I have addressed all your concerns. However, I have retained the dictionary-based approach instead of switching to the group-by option. Please let me know if this is acceptable. Thanks,

trackintel/analysis/tracking_quality.py
```diff
@@ -92,7 +93,9 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
         # get the tracked day relative to the first day
         start_date = df["started_at"].min().date()
         df["day"] = df["started_at"].apply(lambda x: (x.date() - start_date).days)
-        dict_day_to_date = dict(zip(df["day"], df["started_at"].apply(lambda x: x.date())))
+        dict_start_day_to_date = dict(zip(df["day"], df["started_at"].apply(lambda x: x.date())))
+        dict_finish_day_to_date = dict(zip(df["day"], df["finished_at"].apply(lambda x: x.date())))
+        dict_day_to_date = {**dict_start_day_to_date, **dict_finish_day_to_date}
```
hong2223, line 98: Does this basically concatenate the two dict?

trackintel/analysis/tracking_quality.py
```diff
@@ -92,7 +93,9 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
         # get the tracked day relative to the first day
         start_date = df["started_at"].min().date()
         df["day"] = df["started_at"].apply(lambda x: (x.date() - start_date).days)
-        dict_day_to_date = dict(zip(df["day"], df["started_at"].apply(lambda x: x.date())))
+        dict_start_day_to_date = dict(zip(df["day"], df["started_at"].apply(lambda x: x.date())))
+        dict_finish_day_to_date = dict(zip(df["day"], df["finished_at"].apply(lambda x: x.date())))
```
hong2223, line 97: why do we need two dict? if `granularity == "day"`, we first split the original df over multiple days, such that for each record `started_at` day == `finished_at` day. These two dict are essentially the same?

trackintel/analysis/tracking_quality.py
```diff
@@ -106,13 +109,19 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
         # get the tracked week relative to the first day
         start_date = df["started_at"].min().date()
         df["week"] = df["started_at"].apply(lambda x: (x.date() - start_date).days // 7)
-        dict_week_num = dict(zip(df["week"], df["started_at"].apply(lambda x: pd.to_datetime(x).week)))
+        dict_start_week_num = dict(
+            zip(df["week"], df["started_at"].apply(lambda x: pd.to_datetime(x).isocalendar()[1]))
+        )
+        dict_finish_week_num = dict(
```
hong2223, line 115: same as above, I think `dict_start_week_num == dict_finish_week_num`

trackintel/analysis/tracking_quality.py
```diff
@@ -106,13 +109,19 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
...
+            zip(df["week"], df["finished_at"].apply(lambda x: pd.to_datetime(x).isocalendar()[1]))
+        )
+        dict_week_num = {**dict_start_week_num, **dict_finish_week_num}
 
         # calculate per-user per-week raw tracking quality
         raw_quality = df.groupby(["user_id", "week"], as_index=False).apply(_get_tracking_quality_user, granularity)
         # add quality = 0 records
         quality = _get_all_quality(df, raw_quality, granularity)
-        quality["week_number"] = quality["week"].map(dict_week_num)
+        quality["week_number"] = quality["week"].map(dict_week_num).astype("int32")
```
hong2223, line 124: Last time we agree that we still return a DateTime format value when `granularity == "week"`? This means this should be something like the DateTime object of the first day of the week, which ensures flexibility. 

This would also mean we need to change the  `"week_number"` into something like `week_monday`

tests/analysis/test_tracking_quality.py
```diff
@@ -221,6 +221,56 @@ def test_non_positive_duration_warning(self):
...
+            {"user_id": 0, "started_at": t1, "finished_at": t1, "geom": p1},  # duration 0 at midnight
+            {"user_id": 0, "started_at": t1, "finished_at": t1 - ten_days, "geom": p1},  # negative duration
+            {"user_id": 0, "started_at": t1, "finished_at": t1 + ten_days, "geom": p1},  # positive duration
+        ]
+
+        sp = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
+        sp.index.name = "id"
+
+        granularity_ls = ["day"]
+        correct_quality_ls = [1 / 24]
```
hong2223, line 242: we do not need the for loop if we just have one test?

tests/analysis/test_tracking_quality.py
```diff
@@ -221,6 +221,56 @@ def test_non_positive_duration_warning(self):
...
+            {"user_id": 0, "started_at": t1, "finished_at": t1, "geom": p1},  # duration 0 at midnight
+            {"user_id": 0, "started_at": t1, "finished_at": t1 - ten_weeks, "geom": p1},  # negative duration
+            {"user_id": 0, "started_at": t1, "finished_at": t1 + ten_weeks, "geom": p1},  # positive duration
+        ]
+
+        sp = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
+        sp.index.name = "id"
+
+        granularity_ls = ["week"]
+        correct_quality_ls = [1 / 24 / 7]
```
hong2223, line 267: same as above, we do not need the for loop if we just have one test?

hong2223 (changes requested):
Thanks for the work, it took me a bit longer to get back to trackintel... 

I think the dict method is fine, but I cannot fully understand the procedure in between. The specific comments are listed below

trackintel/analysis/tracking_quality.py
abcnishant007: Yes, just a union, same as | operator in python 3.9, 
I did this because some dates were missing in the dictionary map.
This is correct if the test function makes sense, same goes for the week option. Please review this again. We should be on the same page.

Other comments are clear. I will fix in the next commit.

trackintel/analysis/tracking_quality.py
abcnishant007: no, they were not the same in my test case, one day was missing, so I used other dict to ensure all days are present, the previous [comment ](https://github.com/mie-lab/trackintel/pull/328#discussion_r719583534) addresses this. 
Summarily, we should agree on the test case.

trackintel/analysis/tracking_quality.py
abcnishant007: same discussion, will be resolved if we resolve the first [comment](https://github.com/mie-lab/trackintel/pull/328#discussion_r719583534)

trackintel/analysis/tracking_quality.py
abcnishant007: done, the doctoring updated accordingly

tests/analysis/test_tracking_quality.py
abcnishant007: done

tests/analysis/test_tracking_quality.py
abcnishant007: done

trackintel/analysis/tracking_quality.py
hong2223: The problem with the test case is that the last finished_at is at midnight the next day, so if we calculate the quality according to both the started_at and finished_at time, we will always get the last day with 0 quality. 

So I think we should avoid using finished_at. Since we split the records before, the started_at time will always cover all the days (except for the last day with quality 0 in the extreme case )

trackintel/analysis/tracking_quality.py
hong2223: I think this still does not work on my PC... the returned date is in format datatime.date, and not datatime.datetime (the pandas datatype for time). The returned column is therefore not usable using .dt method. I do not know how to change this, but I find using pd.Grouper pandas handled this for us

hong2223:
Wops seems that I messed up with the commits....

But I had a look at the pd.Grouper and find a nice and concise way to rewrite the `temporal_tracking_quality` function, to include the discussions we had in this PR. Your tests are also added.

Probably I should close this one and open a PR for that...

trackintel/analysis/tracking_quality.py
abcnishant007: yes, this is what was led me to have that extra one day! thanks for looking into it!

abcnishant007:
agreed, closed this one.

PR #331, opened 2021-09-11
BUG: Fix reading multiple timezones from postgis 

### LINKED ISSUE #249: trackintel-postgis connector fails when using mixed timezones
reading staypoints from a postgis database with timestamps with mixed timezones results in the following error when using `assert spts.as_staypoints`: 
```
  File "C:\Users\henry\.conda\envs\graph_rep\lib\site-packages\trackintel\model\staypoints.py", line 60, in _validate
    assert pd.api.types.is_datetime64tz_dtype(
AssertionError: dtype of started_at is object but has to be tz aware datetime64
```
This seems to be a pandas-posgresql problem where we can not do much about it but we can solve the problem by transforming data to UTC after reading it from the database.
```python
for col in ['started_at', 'finished_at']:
     spts[col] = pd.to_datetime(spts[col], utc=True)
```
I have ran some quick tests and what it does is that transforms the timezone to UTC without losing the time information (e.g., 13:33+02 -->11:33+00) however I am not sure what happens if no tz is provided

bifbof:
closes #249


tests/io/test_postgis.py
```diff
@@ -291,6 +291,28 @@ def test_non_standard_column_names(self, example_positionfixes, conn_postgis):
...
+        """Test if function can handle different tz informations in one column.
+
+        PostgreSQL saves all its datetimes in UTC and then on exports them to the local timezone.
+        That all works fine except when the local timezone changed in the past for example with daylight saving.
+        """
+        pfs = example_positionfixes.copy()
+        conn_string, conn = conn_postgis
+        table = "positionfixes"
+        sql = f"SELECT * FROM {table}"
+        t1 = pd.Timestamp("2021-08-01 16:00:00", tz="utc")  # summer time
```
henrymartin1, line 304: Did you check if they actually appear in the database as timestamps with different timezones? Timezone UTC should be without summer/winter time, shouldn't it?

henrymartin1 (commented):
Thanks a lot for the fix and sorry for the delay...
I had a small question that you should comment before I can merge it.

tests/io/test_postgis.py
bifbof: Yes, they don't appear in the database as timestamp with different timezones as everything is saved in UTC in postgres.
But I checked the export and there it created two UTC offsets one for the summer time and one for the winter time.

henrymartin1: Great, thanks!

PR #338, opened 2021-09-14
CLN: Turn `copy_docstring` to a protected function.

bifbof:
This function isn't meant to be externally visible.
Also add a docstring describing the function a bit.


hong2223:
👍

PR #347, opened 2021-09-15
CLN: update documentation and fix datatype issues in staypoint merging

### LINKED ISSUE #342: ENH for merge_staypoints() function
The `merge_staypoints()` function in `trackintel/preprocessing/staypoints.py` would need a location_id column. I think this is not yet reflected in the docstring.

Also, the function is not yet added to the documentation.

NinaWie:
closes #342 

The documentation is updated to include that the location id must be part of the staypoints dataframe. I also added an assert statement and a corresponding test. 
Also the issue of changing datatypes in the staypoints table should be fixed, but maybe you can test this with your data

### DIFF
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -233,7 +233,8 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
     Parameters
     ----------
     staypoints : GeoDataFrame (as trackintel staypoints)
-        The staypoints have to follow the standard definition for staypoints DataFrames.
+        The staypoints must contain a column `location_id` (see `generate_locations` function) and have to follow the
+        standard trackintel definition for staypoints DataFrames.
 
     triplegs: GeoDataFrame (as trackintel triplegs)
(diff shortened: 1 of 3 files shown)


trackintel/preprocessing/staypoints.py
```diff
@@ -285,6 +287,9 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
     tpls_merge = triplegs.copy()
     tpls_merge["type"] = "tripleg"
     sp_merge["type"] = "staypoint"
+    # convert datatypes in order to preserve the datatypes (especially ints) despite of NaNs during concat
+    sp_merge = sp_merge.convert_dtypes()
```
NinaWie, line 291: This converts the datatypes to suitable pandas datatypes that allow NaN values also in Int columns. I hope this also works in your case

hong2223:
Wow, thanks for the rapid fix. It also works in my case. I am using an empty tpls table to merge sp, so I think it's exactly what you said the dtype get transferred to float because of the nan. Now this is solved

PR #351, opened 2021-09-18
ENH: Add `geom_col` for read_csv functions

### LINKED ISSUE #343: Enable different geometry names for ``read_*_csv`` functions.
In the `read_*_csv` functions we can load geometries only if they have the column name "geom".
We dropped the requirement of a fixed geometry name some time ago, so we should allow a more flexible approach.
The simplest solution would be to introduce a new keyword "geom_col" to specify the geometry column.

bifbof:
closes #343 
Now we can read csv with different geometry names.


hong2223:
👍

PR #354, opened 2021-09-22
ENH: Redo `read_geolife` for smaller memory footprint

bifbof:
Splits `read_geolife` into two parts. 
I stumbled across the implementation of reading geolife data in [scikit-mobility](https://github.com/scikit-mobility/scikit-mobility/blob/7c50c09adda71338c19dcc03f0e11beb747ca71e/skmob/io/file.py#L82) , which in fact is very clean.
That I wanted also for our implementation. That worked out, at least for me the function seems simpler (but I am probably biased).

What is changed:

1. Dict of labels get generated in `_get_labels` for modularity.
2. `_get_df` now a generator to efficiently concat all these trajectories
  without building a huge list containing all data. &rarr; limiting the
  memory footprint.

### DIFF
--- a/trackintel/io/dataset_reader.py
+++ b/trackintel/io/dataset_reader.py
@@ -186,11 +186,10 @@ def _get_df(geolife_path, uids, print_progress):
         for traj_file in glob.glob(pattern):
             data = pd.read_csv(traj_file, skiprows=6, header=None, names=names, usecols=usecols)
             data["tracked_at"] = pd.to_datetime(data["date"] + " " + data["time"], format="%Y-%m-%d %H:%M:%S", utc=True)
-            data["geom"] = list(zip(data["longitude"], data["latitude"]))
-            data.drop(columns=["date", "time", "longitude", "latitude"], inplace=True)
-            data["geom"] = gpd.GeoSeries(data["geom"].apply(Point))
+            data["geom"] = gpd.points_from_xy(data["longitude"], data["latitude"])
             data["user_id"] = int(user_id)
             data["elevation"] = data["elevation"] * FEET2METER
+            data.drop(columns=["date", "time", "longitude", "latitude"], inplace=True)
             yield data
 
 
--- a/trackintel/io/dataset_reader.py
+++ b/trackintel/io/dataset_reader.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 
 import glob
-import ntpath
 import os
+from collections import defaultdict
+from functools import partial
 
 import geopandas as gpd
 import numpy as np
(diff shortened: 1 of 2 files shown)


bifbof:
With pygeos we may have to think about doing more work outside of these for loops.
Especially the functions on the columns could speed up quite a bit this way.
But if we decide to do that we should create another PR.

hong2223:
> Handed the creation of geometry to the `gpd.points_from_xy` function that uses pygeos if installed. With pygeos introduced we may have to think about doing more work outside of these for loops. Especially the functions on the columns could speed up quite a bit this way. But if we decide to do that we should create another PR.

Thanks for this change, makes the code much cleaner. I also like the `yield` solution now, this aligns the tqdm print time with the actual time (previously tqdm time is always shorter as most of the time is spent on the final geopandas conversion). 

Could you write an issue stating where in the function could be further speed up?

trackintel/io/dataset_reader.py
```diff
@@ -93,80 +92,105 @@ def read_geolife(geolife_path, print_progress=False):
...
-            data_this["geom"] = data_this["geom"].apply(Point)
+    Notes
+    -----
+    No further checks are done on user ids, they must be convertable to ints.
+    """
+    labels_rename = {"Start Time": "started_at", "End Time": "finished_at", "Transportation Mode": "mode"}
+    label_dict = defaultdict(
+        partial(pd.DataFrame, columns=["started_at", "finished_at", "mode"])
+    )  # output dict for the labels
+    # TODO: change geolife_add_modes_to_triplegs so that we can use a dict instead.
```
hong2223, line 143: I do not understand what this means... Is the `get_geolife_ori_pfs` still compatible with `geolife_add_modes_to_triplegs`?

Edit: I tested and the functions work fine with each other. What is the concern here?

hong2223 (approved):

trackintel/io/dataset_reader.py
bifbof: Yes, but it is a bit messy. To be able to split the labels up I created a defaultdict with an empty df. So everytime a user is called without labels ([here](https://github.com/mie-lab/trackintel/blob/feaf7f37b38c1a5659cd8f6ef6982c3b505110dd/trackintel/io/dataset_reader.py#L228)), a new entry is created with an empty dataframe. It would also be possible just to iterate over all keys, but for that we need to change the `geolife_add_modes_to_tripleg` iterating structure.

hong2223: Thanks for the explanation. I think we should change the iterating structure as you said.

bifbof: +1 I wrote an issue for that.

hong2223:
I will now merge this PR (after checking the test errors). Things in the TODO list after this PR:

- pygeos integration for more efficient handling of data
- issue #358

PR #355, opened 2021-09-27
ENH: Add pygeos to requirements

### LINKED ISSUE #353: pygeos for trackintel
pygeos is a package that vectorizes the geometry in GeoPandas using Cython. It is maintained basically by the same people that maintain geopandas. As such they've implemented a lot of functions in geopandas to run faster if pygeos is installed. That would give us a free speedup and further allow us to use vectorized functions on geometryArrays similar to ufunc on numpy arrays. 

Further informations:
- https://jorisvandenbossche.github.io/blog/2017/09/19/geopandas-cython/
- https://github.com/pygeos/pygeos

One use case would be the generation of points out of longitude and latitude. Instead of using the relatively slow `.apply(Points)` we can use directly on arrays `pygeos.points(x, y)`. 
This way we get the following times on my laptop.

Preparation
````python
import numpy as np
import pandas as pd

a = np.arange(100_000)
b = np.arange(100_000, 200_000)
df = pd.DataFrame({"a": a, "b": b})
````

`pygeos`
````python
%timeit pygeos.points(df["a"], df["b"])
57.4 ms ± 4.71 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
````

Our way so far
````python
%%timeit
df["geom"] = list(zip(df["a"], df["b"]))
df["geom"] = df["geom"].apply(Point)
1.08 s ± 31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
````

bifbof:
closes #353 

See #353 for more informations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ ti.print_version()
 * tqdm
 * OSMnx
 * similaritymeasures
+* pygeos
 
 ## Development
 You can find the development roadmap under `ROADMAP.md` and further development guidelines under `CONTRIBUTING.md`.
--- a/ci/envs/37-latest-conda-forge.yaml
+++ b/ci/envs/37-latest-conda-forge.yaml
@@ -20,4 +20,5 @@ dependencies:
 - pytest-xdist
 - tqdm
 - similaritymeasures
+- pygeos
 - jupyter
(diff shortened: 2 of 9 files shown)


hong2223:
This only adds `pygeos` as a dependency and do not change our code, right? I think it might be good to change the code in this PR as well

bifbof:
> This only adds `pygeos` as a dependency and do not change our code, right? I think it might be good to change the code in this PR as well

That is a good idea. Unfortunatly I wanted to add it first was for the dataset reader, but there is already a PR that would conflict with any changes introduced here.
As for now I would just merge this pull request. Then I put the changes into the other PR.

hong2223:
So I would merge this now

PR #359, opened 2021-10-01
ENH: use pd.Grouper for tracking quality

### LINKED ISSUE #258: How to use temporal_tracking_quality to delete bad tracking days

hong2223:
closes #258

main changes:
- We use pd.Grouper to group days or weeks and return the datatime object. The granularity == "day" and "week" now only returns the absolute date. 
- As we have no relative days, we no longer fill in quality == 0 records (_get_all_quality() is now unused).
- Description added to the docstrings

### DIFF
--- a/tests/analysis/test_tracking_quality.py
+++ b/tests/analysis/test_tracking_quality.py
@@ -221,6 +221,48 @@ class TestTemporal_tracking_quality:
         with pytest.warns(UserWarning):
             ti.analysis.tracking_quality.temporal_tracking_quality(sp)
 
+    def test_last_day_same_as_last_day_in_record(self):
+        """Test the take the last tracking record's day and assert it is the same as the
+        last absolute date in the quality."""
+        p1 = Point(8.5067847, 47.4)
+        t1 = pd.Timestamp("1971-01-01 00:00:00", tz="utc")
+
+        ten_days = pd.Timedelta(days=10)
+
+        list_dict = [
+            {"user_id": 0, "started_at": t1, "finished_at": t1, "geom": p1},  # duration 0 at midnight
+            {"user_id": 0, "started_at": t1, "finished_at": t1 - ten_days, "geom": p1},  # negative duration
+            {"user_id": 0, "started_at": t1, "finished_at": t1 + ten_days, "geom": p1},  # positive duration
+        ]
+
+        sp = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
+        sp.index.name = "id"
+
+        quality = ti.analysis.tracking_quality.temporal_tracking_quality(sp, granularity="day")
+        # get the "date" of the last record and compare to the last "date" in data
+        assert quality.values[-1][-2] == (t1 + ten_days - pd.Timedelta(days=1))
+
+    def test_last_day_same_as_last_week_in_record(self):
+        """Test the take the last tracking record's day and assert it is the same as the
+        last absolute date in the quality."""
+        p1 = Point(8.5067847, 47.4)
+        t1 = pd.Timestamp("1971-01-01 00:00:00", tz="utc")
+
+        four_weeks = pd.Timedelta(weeks=4)
+
+        list_dict = [
+            {"user_id": 0, "started_at": t1, "finished_at": t1, "geom": p1},  # duration 0 at midnight
+            {"user_id": 0, "started_at": t1, "finished_at": t1 - four_weeks, "geom": p1},  # negative duration
+            {"user_id": 0, "started_at": t1, "finished_at": t1 + four_weeks, "geom": p1},  # positive duration
+        ]
+
+        sp = gpd.GeoDataFrame(data=list_dict, geometry="geom", crs="EPSG:4326")
+        sp.index.name = "id"
+
+        quality = ti.analysis.tracking_quality.temporal_tracking_quality(sp, granularity="week")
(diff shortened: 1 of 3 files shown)


hong2223:
Please check if this PR achieve the functionalities we discussed @abcnishant007

trackintel/analysis/tracking_quality.py
```diff
@@ -40,17 +40,33 @@ def temporal_tracking_quality(source, granularity="all", max_iter=60):
...
+
+    - ``weekday``
+
+        - time extent: the whole day (86400 sec) * number of tracked weeks for each user for each user
+        - columns: ``['user_id', 'weekday', 'quality']``
+
+    - ``hour``:
+
+        - time extent: the whole hour (3600 sec) * number of tracked days for each user
+        - columns: ``['user_id', 'hour', 'quality']``
```
abcnishant007, line 70: +1 to the docstring update, this summarises everything.
Also, since you did not change the order of the `quality` column, other test functions need not be updated as well (for example, I had to make it -2 at some places instead of -1). Now it can be -1 throughout. 

I think this is a better approach than what I was doing (returning both items: relative and absolute.)

tests/analysis/test_tracking_quality.py
```diff
@@ -69,8 +69,8 @@ def test_tracking_quality_day(self, testdata_sp_tpls_geolife_long):
         # test if the result of the user agrees
         quality = ti.analysis.tracking_quality.temporal_tracking_quality(sp_tpls, granularity="day")
 
-        assert quality_manual == quality.loc[(quality["user_id"] == 0) & (quality["day"] == 0), "quality"].values[0]
-        assert (quality["quality"] < 1).all()
+        assert quality_manual == quality.iloc[0]["quality"]
+        assert (quality["quality"] <= 1).all()
```
abcnishant007, line 73: Why did we need `<=` instead of `<` that was there before?

tests/analysis/test_tracking_quality.py
```diff
@@ -92,7 +92,7 @@ def test_tracking_quality_week(self, testdata_sp_tpls_geolife_long):
         quality = ti.analysis.tracking_quality.temporal_tracking_quality(sp_tpls, granularity="week")
 
         assert quality_manual == quality.loc[(quality["user_id"] == 0), "quality"].values[0]
-        assert (quality["quality"] < 1).all()
+        assert (quality["quality"] <= 1).all()
```
abcnishant007, line 95: same as previous comment

tests/analysis/test_tracking_quality.py
```diff
@@ -221,6 +221,48 @@ def test_non_positive_duration_warning(self):
...
+
+        quality = ti.analysis.tracking_quality.temporal_tracking_quality(sp, granularity="day")
+        # get the "date" of the last record and compare to the last "date" in data
+        assert quality.values[-1][-2] == (t1 + ten_days - pd.Timedelta(days=1))
+
+    def test_last_day_same_as_last_week_in_record(self):
+        """Test the take the last tracking record's day and assert it is the same as the
+        last absolute date in the quality."""
+        p1 = Point(8.5067847, 47.4)
+        t1 = pd.Timestamp("1971-01-01 00:00:00", tz="utc")
```
abcnishant007, line 249: Should we add another test case where we have a time stamp starting NOT at `00:00:00`so that we can be sure that it works as expected for non-boundary cases? 
Reminder: I am referring to the issue which I had in my solution, extra day due to starting at 00:00:00

abcnishant007 (commented):
Added some pointers to draw your attention to some minor points. Otherwise this is good!

tests/analysis/test_tracking_quality.py
hong2223: This relates to a bug we introduced before, check #221. So quality theoretically could take the value 1. The change here is just for the "correctness" and did not influence test results

hong2223 (commented):
Thanks for taking a look. I separated the function for generating test records and added one test case for a time starting in the middle of the day. Shall be ready for review

abcnishant007:
Thanks for making the changes. 

Looking at the report, I don't see coverage being reduced. Is the `codecov` red flag due to some codecov bug? If yes, please feel free to merge.

hong2223:
I am not sure what causes the failure of our tests. We need to fix this before merging

henrymartin1:
> 
> 
> I am not sure what causes the failure of our tests. We need to fix this before merging

I restarted the test job. There was a similar problem on the master branch that was solved by re-running the tests ;-)

abcnishant007:
> > I am not sure what causes the failure of our tests. We need to fix this before merging
> 
> I restarted the test job. There was a similar problem on the master branch that was solved by re-running the tests ;-)

That used to work but didn't work this time it seems. I will check if pushing another dummy commit fixes it.

PR #360, opened 2021-10-02
CLN: Change labels from defaultdict to dict

### LINKED ISSUE #358: CLN: Iterate over label keys in `geolife_add_modes_to_triplegs`
In #354 the type of the labels dict changed to defaultdict. Now if we iterate over all users we create a lot of empty dataframes in the defaultdict. Two proposed changes to change that:

1. change type back to dict
2. iterate [here](https://github.com/mie-lab/trackintel/blob/feaf7f37b38c1a5659cd8f6ef6982c3b505110dd/trackintel/io/dataset_reader.py#L226) over labels key instead of unique user_ids

bifbof:
closes #358


hong2223:
Thanks. This works nicely on my PC. Merging this now

PR #361, opened 2021-10-19
ENH: Computing speed per positionfix and mean speed per tripleg

### LINKED ISSUE #191: ENH: Speed calculation for triplegs/positionfixes
during my internship I once used movingpandas to calculate speeds per positionfix and mean speeds for triplegs (for delay classification of measurement on a bus).

In the last meeting we agreed that this could also be useful in the future for trackintel 2.0

at the moment the transport mode prediction does a mean speed calculation for quick classification of transport modes

NinaWie:
closes #191 

I have added functions to compute the speed per positionfix and the average for a tripleg. The pfs-wise speed is computed similar to the implementation in movingpandas [here](https://github.com/anitagraser/movingpandas/blob/ea9f9947dab756271a2910bd43f9127c27135d19/movingpandas/trajectory.py#L611). 

Main open questions: 
* Where should these functions be added in the code? I currently put them in a new file in `trackintel/analysis/speed.py` because it didn't seem to fit well anywhere. Maybe we can discuss this shortly.
* See my comment to the issue [here](https://github.com/mie-lab/trackintel/issues/191#issuecomment-946604676)  - do we need functions for space_mean and time_mean?
* I need to improve the tests, but I find it a bit hard to design test data with ground truth values for speed calculation. Let me know if you have any ideas.


NinaWie:
@henrymartin1  feel free to assign someone else for this PR review

hong2223:
I can try to review this PR

trackintel/analysis/speed.py
```diff
@@ -0,0 +1,83 @@
...
+    pfs: GeoDataFrame (as trackintel positionfixes)
+        The original positionfixes with a new column ``[`speed`]``. The speed is given in km/h
+
+    Notes
+    -----
+    The speed at one positionfix is computed from the distance and time since the previous positionfix. For the first
+    positionfix, the speed is set to the same value as for the second one.
+    """
+    pfs = positionfixes.copy()
+    crs_is_projected = ti.geogr.distances.check_gdf_planar(pfs)
```
hong2223, line 26: I would call this `if_planer_crs`, as it is called in `_predict_transport_mode_simple_coarse()` in labelling.
`crs_is_projected` might cause confusion in "is projected" or "is in the projected coordinate system".

trackintel/analysis/speed.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if crs_is_projected:
+        dist_function = lambda x: x.geom.distance(x.prev_geom)
+    else:
+        dist_function = lambda x: ti.geogr.point_distances.haversine_dist(
+            x.geom.x, x.geom.y, x.prev_geom.x, x.prev_geom.y
+        )[0]
```
hong2223, line 33: why do we need the [0]? the return of `ti.geogr.point_distances.haversine_dist` is only a float number?

trackintel/analysis/speed.py
```diff
@@ -0,0 +1,83 @@
...
+    positionfix, the speed is set to the same value as for the second one.
+    """
+    pfs = positionfixes.copy()
+    crs_is_projected = ti.geogr.distances.check_gdf_planar(pfs)
+
+    if crs_is_projected:
+        dist_function = lambda x: x.geom.distance(x.prev_geom)
+    else:
+        dist_function = lambda x: ti.geogr.point_distances.haversine_dist(
+            x.geom.x, x.geom.y, x.prev_geom.x, x.prev_geom.y
```
hong2223, line 32: For readability, I would put this whole block (`check_gdf_planar` block) after line 38, where `prev_geom` is defined. And here maybe `lambda point` instead of `lambda x`?

trackintel/analysis/speed.py
```diff
@@ -0,0 +1,83 @@
...
+    pfs.loc[pfs.index[0], "prev_geom"] = pfs.loc[pfs.index[0], "geom"]
+    pfs["prev_tracked_at"] = pfs["tracked_at"].shift(1)
+
+    # get distance and time difference
+    pfs["dist"] = pfs.apply(dist_function, axis=1)
+    pfs["time_diff"] = pfs.apply(lambda x: x.tracked_at - x.prev_tracked_at, axis=1).dt.total_seconds()
+
+    # compute speed and convert to kmh
+    pfs["speed"] = 3.6 * pfs["dist"] / pfs["time_diff"]
+    pfs.loc[pfs.index[0], "speed"] = pfs.iloc[1]["speed"]
```
hong2223, line 46: This is for assigning the first point speed, right?

trackintel/analysis/speed.py
```diff
@@ -0,0 +1,83 @@
...
+    grouped_pfs = positionfixes.groupby("tripleg_id").apply(_single_tripleg_mean_speed)
+    # add the speed values to the triplegs column
+    tpls = pd.merge(triplegs, grouped_pfs.rename("mean_speed"), left_index=True, right_index=True)
+    tpls.index = tpls.index.astype("int64")
+    return tpls
+
+
+def _single_tripleg_mean_speed(positionfixes):
+    pfs_sorted = positionfixes.sort_values(by="tracked_at")
+    pfs_speed = speed_positionfixes(pfs_sorted)
```
hong2223, line 82: We have two ways to calculate tripleg speed:
- We take the total length and divide by the total duration (as implemented in `_predict_transport_mode_simple_coarse` in labelling)
- We take the mean of point speed (as implemented here).

In general, I think the first way is more natural, as the second method does not consider the weight (length) of each pfs?

tests/analysis/test_speed.py
```diff
@@ -0,0 +1,102 @@
...
+    sp = sp.as_staypoints.create_activity_flag(time_threshold=15)
+    pfs, tpls = pfs.as_positionfixes.generate_triplegs(sp)
+    return pfs, sp, tpls
+
+
+@pytest.fixture
+def load_positionfixes():
+    """Load test positionfixes"""
+    pfs = ti.io.file.read_positionfixes_csv(os.path.join("tests", "data", "positionfixes.csv"), sep=";", index_col="id")
+    correct_speeds = np.array([8.82100607, 8.82100607, 0.36585538, 1.93127652, 19.60643425, 2.07086017])
```
hong2223, line 24: Do you calculate this using a calculator or e.g., QGIS? Could you document it somewhere? But this is a good comparison:)

tests/analysis/test_speed.py
```diff
@@ -0,0 +1,102 @@
...
+        pfs, _ = load_positionfixes
+        with pytest.raises(Exception) as e_info:
+            _ = ti.analysis.speed.mean_speed_triplegs(pfs, None)
+            assert e_info == "Positionfixes must include column tripleg_id"
+
+    def test_tripleg_speed_correct(self, example_triplegs):
+        """Test whether the computed mean speed values correspond to the one yielded from linestrings"""
+        pfs, _, tpls = example_triplegs
+        tpls_speed = ti.analysis.speed.mean_speed_triplegs(pfs, tpls)
+        ground_truth_speed = [
```
hong2223, line 78: Same as above, document the origin

hong2223 (changes requested):
Thanks for the implementation. I think the code is quite complete; we might need to discuss the implementation of tripleg speed. I have some general comments:
- I would think this module ("speed") more of a helper module, where you can call the speed calculation within each model (something like pfs.calculate_speed()). This means 1. we need to register the method in ti.model.positionfixes and ti.model.triplegs. and 2, we need to discuss where to place them (I think maybe in model.utils is better than in analysis?)
- Would it be possible to rewrite the `identify_mode()` function in labelling, to use the speed calculation function here? I think we do not need two speed implementation

tests/analysis/test_speed.py
NinaWie: I just computed it with python and then double checked it on several values. I'm always unsure what is the best method.. I haven't worked with qgis so that would be a huge overhead for me.

trackintel/analysis/speed.py
NinaWie: Done

trackintel/analysis/speed.py
NinaWie: Good points, I did both in the new version

trackintel/analysis/speed.py
NinaWie: As you can see [here](https://github.com/mie-lab/trackintel/blob/3a67d0c7dd2ac1f305b37c81a17e49365d693d79/trackintel/geogr/distances.py#L56), haversine_dist always returns an array.

trackintel/analysis/speed.py
NinaWie: Yes, I added a comment

trackintel/analysis/speed.py
NinaWie: There are now both versions in the new code. You are right though that the positionfix-wise speed does not make that much sense. Maybe we should just delete it. Let me know if you'd keep both.

NinaWie (commented):
I have refactored everything such that it's now in `model/util.py` and it's also available via the accessor. Also, I've added a function for the simple tripleg speed calculation based on overall distance and duration. The labelling function is adapted accordingly. Since it is refactored, it might be a bit hard to see the changes I did compared to the first version, but I answererd to your comments to address them. Let me know what you think

NinaWie:
The tests work on my machine, I think it's a precision error. I'll take care of it in the next iteration

trackintel/model/util.py
```diff
@@ -1,4 +1,102 @@
...
+    pfs["speed"] = 3.6 * pfs["dist"] / pfs["time_diff"]
+    # The first point speed is imputed
+    pfs.loc[pfs.index[0], "speed"] = pfs.iloc[1]["speed"]
+
+    pfs.drop(columns=["prev_geom", "prev_tracked_at", "dist", "time_diff"], inplace=True)
+
+    return pfs
+
+
+def speed_triplegs(triplegs):
```
hong2223, line 55: Docstring missing. You can copy the docstring from below and slightly change the description

hong2223 (approved):
Thanks for the fix. I only have minor points:

- I am wondering about the function names, currently it would look like `example_pfs.as_positionfix.speed_positionfixes()` when we call this function. Would it be better to rename it to `example_pfs.as_positionfix.get_speed()` for both pfs and tpls?
- `speed_triplegs` misses docstring
- I am fine with keeping `pfs_mean_speed_triplegs`, but I can hardly imagine I need to write `example_tpl.as_triplegs.pfs_mean_speed_triplegs()` to call the function. I think it would be better if both functions are called using `get_speed`, but distinguished using a input parameter like `method = {"pfs_mean_speed", "tpl_speed"}`.

Let me know if these comments make sense.

NinaWie:
Okay thanks for the suggestions, I think everything is now changed accordingly :) Maybe you can double check, but feel free to merge.

NinaWie:
The tests I had leaded to precision errors that I could not reproduce on my machine. I now replaced two tests with other ones that just recompute one speed manually. Hope that's also fine, or do you have any idea where the differences in values could come from? I already compared my geopandas / numpy versions to the one used here, but that didn't change anything. If you are okay with the current tests, feel free to merge now.

hong2223:
> The tests I had leaded to precision errors that I could not reproduce on my machine. I now replaced two tests with other ones that just recompute one speed manually. Hope that's also fine, or do you have any idea where the differences in values could come from? I already compared my geopandas / numpy versions to the one used here, but that didn't change anything. If you are okay with the current tests, feel free to merge now.

I had a similar issue some time ago with another test in trackintel, but I also do not know the cause... 
I am fine with this PR and am now merging it.

PR #364, opened 2021-12-08
BLD: release 1.1.5

### LINKED ISSUE #363: Tour generation not available in current release
We should create a new release that includes tour generation.

NinaWie:
Increase version number of new release (closes #363 )

@henrymartin1  or @hong2223 , can you check whether my draft for the release looks okay and whether there is something else that I forgot to do for the release? Thanks :)


hong2223:
thanks, I think this looks fine. i will merge this now

PR #367, opened 2022-01-08
BUG: Load extent correctly from postgis

### LINKED ISSUE #365: Bug in reading locations with extent from postgis
I wrote a `locations` table to postgis with the function 
```
locs.as_locations.to_postgis(con=engine, name="locations", ...    )
```
I try to read it with 
```
sql = "SELECT * FROM schema.locations"
locs = ti.io.read_locations_postgis(sql, con=engine, extent="extent")
```
However, the extent is not a geometry in the output `locs`. I had to add the following line to make it work (the problem was the wkb encoding):
```
 locs["extent"] = locs["extent"].apply(lambda x: wkb.loads(bytes.fromhex(x)))
```

So we probably have to include this in the `read_locations_postgis` function. Please let me know if I just overlooked something, or whether I should go ahead with the changes.

bifbof:
Bugfix for reading extent correctly from postgis.
Closes #365

### DIFF
--- a/tests/io/test_postgis.py
+++ b/tests/io/test_postgis.py
@@ -517,11 +517,11 @@ class TestLocations:
             assert geom_schema_extent in dtypes
         finally:
             del_table(conn, table)
-    
+
     def test_read_extent(self, example_locations, conn_postgis):
         """Test if extent geometry can be read correctly."""
         conn_string, conn = conn_postgis
-        table= "locations"
+        table = "locations"
         sql = f"SELECT * FROM {table}"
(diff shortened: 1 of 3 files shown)


bifbof:
It seems that we have an unrelated bug. The strange thing is that I cannot reproduce this bug on my local machine.

hong2223:
> It seems that we have an unrelated bug.

Thanks for the PR, I'll take a look at our test error

bifbof:
> Thanks for the PR, I'll take a look at our test error

Nice thanks for the fix.

NinaWie:
Thanks for taking care of this @bifbof! Looks good to me, I already merged it

PR #372, opened 2022-02-05
TST/BLD: Fix ci

### LINKED ISSUE #370: BUG/TST: Github action failed
Our daily GitHub action failed 8 days ago when we had our last pull request. <s>I suspect that some changes from #361 is causing the error.</s>

hong2223:
closes #370.

Basically deleted the requirements of `pandas<=1.2.5`. The test now only includes `python-version: ["3.9"]`, dependency issues still exist for python 3.7.

Two tests would fail with the newer version of pandas. I temporarily commented them out.


hong2223:
I decided to only include tests with python=3.9 and packages installed from the conda-forge channel. This seems to be used by similar packages

bifbof:
I have a small question regarding this pull request. :)
Why was pygeos removed?

hong2223:
> I have a small question regarding this pull request. :) Why was pygeos removed?

I find pygeos never used in our repo. It is also a required dependency for geopandas that shall be already installed

bifbof:
Ah okay that makes sense, thanks for the answer. Geopandas uses it for io but they didn't included it in their dependencies back then.

PR #377, opened 2022-02-20
ENH: 5x speed increase for staypoint generation

bifbof:
I tinkered around a bit and looked at the code of the staypoint generation as that takes always a lot of time. 
Found out that most of the time is spent accessing the x and y coordinates (shapely is slow) so I have thrown the coordinates into numpy arrays. That way the access takes nanoseconds instead of microseconds.
I also moved some calculation outside of the for-loop to speedup the time comparisons but those improvements are rather minor.
Now the limiting factor is the distance function. For this reason I replaced the casting to radians with `np.deg2rad` as this is some microseconds faster.

This results in a speed up of around a factor 5. (Tested on my laptop with users 51 and 52)
- Before the changes it took around 52 seconds
- After the changes it takes around 11 seconds

If we want to increase the speed of the function further we probably have to use either cython or numba.
Another way would be to do the distance calculations in "blocks" instead of single entries that way we can use the speedup from numpy more.

### DIFF
--- a/trackintel/preprocessing/positionfixes.py
+++ b/trackintel/preprocessing/positionfixes.py
@@ -415,50 +415,48 @@ def _generate_staypoints_sliding_user(
     else:
         raise AttributeError("distance_metric unknown. We only support ['haversine']. " f"You passed {distance_metric}")
 
-    df = df.sort_index(kind="mergesort").sort_values(by=["tracked_at"], kind="mergesort")
-    # pfs id should be in index, create separate idx for storing the matching
-    pfs = df.to_dict("records")
-    idx = df.index.to_list()
+    df = df.sort_index(kind="stable").sort_values(by=["tracked_at"], kind="stable")
+    # transform times to pandas Timedelta to simplify comparisons
+    gap_threshold = pd.Timedelta(gap_threshold, unit="minutes")
+    # precalulate variables to not do that in for loop
+    time_threshold = time_threshold * 60
+    last_curr = len(df.index) - 1
+    # to numpy as access time of numpy numpy is faster than pandas array
+    gap_times = pd.eval("((df.tracked_at - df.tracked_at.shift(1)) > gap_threshold)").to_numpy()
+
+    # put x and y into numpy arrays to speed up the access in the for loop (shapely is slow)
+    x = df[geo_col].x.to_numpy()
+    y = df[geo_col].y.to_numpy()
 
     ret_sp = []
     start = 0
+    for curr in range(1, len(df)):
 
-    # as start begin from 0, curr begin from 1
-    for i in range(1, len(pfs)):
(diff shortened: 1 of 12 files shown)


trackintel/preprocessing/positionfixes.py
```diff
@@ -415,78 +415,72 @@ def _generate_staypoints_sliding_user(
...
-    pfs = df.to_dict("records")
-    idx = df.index.to_list()
+    df = df.sort_index(kind="stable").sort_values(by=["tracked_at"], kind="stable")
+    # transform times to pandas Timedelta to simplify comparisons
+    gap_threshold = pd.Timedelta(gap_threshold, unit="minutes")
+    # precalulate variables to not do that in for loop
+    time_threshold = time_threshold * 60
+    last_curr = len(df.index) - 1
+    # to numpy as access time of numpy numpy is faster than pandas array
+    gap_times = pd.eval("((df.tracked_at - df.tracked_at.shift(1)) > gap_threshold)").to_numpy()
```
hong2223, line 426: Nice solution!

hong2223 (approved):
Thanks for the nice solution! This is a really important update.

I tested with the whole geolife dataset and this PR will create a major speedup!

trackintel/preprocessing/positionfixes.py
```diff
@@ -415,78 +415,72 @@ def _generate_staypoints_sliding_user(
...
 
-    df = df.sort_index(kind="mergesort").sort_values(by=["tracked_at"], kind="mergesort")
-    # pfs id should be in index, create separate idx for storing the matching
-    pfs = df.to_dict("records")
-    idx = df.index.to_list()
+    df = df.sort_index(kind="stable").sort_values(by=["tracked_at"], kind="stable")
+    # transform times to pandas Timedelta to simplify comparisons
+    gap_threshold = pd.Timedelta(gap_threshold, unit="minutes")
+    # precalulate variables to not do that in for loop
+    time_threshold = time_threshold * 60
```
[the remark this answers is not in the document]
bifbof: Yes I think it also fine like this, I just wanted to move the calculation outside of the function.

trackintel/preprocessing/positionfixes.py
```diff
@@ -416,11 +416,12 @@ def _generate_staypoints_sliding_user(
         raise AttributeError("distance_metric unknown. We only support ['haversine']. " f"You passed {distance_metric}")
 
     df = df.sort_index(kind="stable").sort_values(by=["tracked_at"], kind="stable")
-    # transform times to pandas Timedelta to simplify comparisons
-    gap_threshold = pd.Timedelta(gap_threshold, unit="minutes")
```
bifbof: Does the comparison still work if we delete that?

PR #384, opened 2022-03-18
ENH: `_split_overlap` splits now exactly on border

### LINKED ISSUE #369: ENH: _split_overlap() function does not deal with miliseconds
The splitting of records implemented in the `_split_overlap()` function only replaces the time larger than 1 second, causing the newly created records not to start exactly at 12 midnight. The codes [here](https://github.com/mie-lab/trackintel/blob/1ff2323c868cfb82c0d0497cf57281c821e044d1/trackintel/analysis/tracking_quality.py#L253) and [here](https://github.com/mie-lab/trackintel/blob/1ff2323c868cfb82c0d0497cf57281c821e044d1/trackintel/analysis/tracking_quality.py#L257) need to be changed.

image

bifbof:
This PR does quite a lot because some changes where necessary to generalize this method.
- use `dt.ceil()` and `dt.floor()` to handle the splits more generally.
- fixed multiple bugs that come from using `.date`/`.hour` (if on same date on different months it wouldn't be recongized)
- wrote test for corner cases in `_split_overlap`
- wrote test for helpfunction `_get_split_index`.
- some cleanup and speedup here and there

closes #369

### DIFF
--- a/tests/analysis/test_tracking_quality.py
+++ b/tests/analysis/test_tracking_quality.py
@@ -7,6 +7,7 @@ import geopandas as gpd
 from shapely.geometry import Point
 
 import trackintel as ti
+from trackintel.analysis.tracking_quality import _get_split_index
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


hong2223:
Thanks for the nice fix! This issue has long bothered me and is finally solved!

PR #387, opened 2022-04-02
ENH: Replace df.append with pd.concat (Future warning)

bifbof:
In future versions of Pandas, `concat` will be deprecated. 
Read more about this [here](https://github.com/pandas-dev/pandas/issues/35407).


hong2223:
I checked all the changes, and they are fine. Thanks for the fix!

PR #392, opened 2022-04-21
ENH: add code for case study comparing four datasets

NinaWie:
This is a first clean version of the code for the case study in our Trackintel paper. 

Note:
* The code is not executable for someone external because it requires access to our database. An alternative would be to specify `studies=["geolife"]` on top of the notebook such that all anaylsis is only done for Geolife. However, this will most likely look super weird in the plots, which are designed to compare multiple datasets.
* We should probably remove the outputs, i.e. the plots and tables, from the notebook because we shouldn't upload such a large file to github. However, if people cannot execute the notebook, then it's even harder to understand the code / use it if there are no outputs.
* Many of the steps take a long time to execute, because they for example require loading all the triplegs for all four studies (e.g. to compute the distance). 
* The notebook is now structured into clear sections, but it's not commented very well. Not sure how much work we need to put into this. I guess for the paper submission it wouldn't be necessary. 

In any case, this PR is rather a way for you to be able to see the code so that we can discuss the further todos.


henrymartin1 (approved):
Awesome! 🎉

PR #393, opened 2022-05-01
ENH: Simplified logic in `_split_overlaps`

bifbof:
When I updated `split_overlaps` in #384, I had an idea for an alternative implementation.
I am not completely happy with mine either because I had to use the `apply` method. I did not find another way with the two datetime columns ["started_at", "finished_at"]. The logic is simpler, but I haven`t looked at the performance.
So this is more a suggestion, if you think this is a bad idea, we can drop it.

This PR contains:
- removed unused method `_get_all_quality`
- removed `max_iter` keyword argument.
- created new function `_get_times` that returns list of start-end times.
- applied this function to all rows for a more elegant split.
- wrote test for new `_get_times` function.

### DIFF
--- a/tests/analysis/test_tracking_quality.py
+++ b/tests/analysis/test_tracking_quality.py
@@ -7,7 +7,7 @@ import geopandas as gpd
 from shapely.geometry import Point
 
 import trackintel as ti
-from trackintel.analysis.tracking_quality import _get_split_index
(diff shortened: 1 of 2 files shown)


henrymartin1 (approved):
Nice work!
I personally liked the iterate+shift style but I agree that this is cleaner ;-)
Thanks for the contribution.

PR #405, opened 2022-05-19
ENH: Add test for check_gdf_crs

bifbof:
Rewrite logic of `check_gdf_crs` to be more flat and added test for missed testcase.


henrymartin1 (approved):
I remember that we had some discussions with this function because it was always a bit weird. It looks better now, thanks!

PR #406, opened 2022-05-19
ENH: increase coverage for calc_temp_overlap

bifbof:
Rewrote the `calc_temp_overlap` function to make it simpler and to eliminate states that are not reachable &rarr; increasing the coverage of the code.


henrymartin1 (commented):
Thanks for the clean-up! This is now probably the function with the most tests per line of code in trackintel ;-)

henrymartin1 (approved):

PR #408, opened 2022-05-20
BUG: Set index for mask in location_identifier

### LINKED ISSUE #403: Bug in location_identifier for pre_filter=False and non-serial index
The function [location_identifier](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L7) assumes that an index is serial e.g., incremental from 0 to n. It throws an error in fthe index is not serial and sorts the dataframe (may not be intended) if the index is serial but not in order.

The problem is that for `pre_filter=False` [here](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L65) a boolean series with the same length as the staypoint dataframe and serial index is created. 

From the geopandas documentation for what can be input to `loc` for indexing: 
_An alignable boolean Series. The index of the key will be aligned before masking._ [source](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html)

A fix could be to either use the original index when creating `f` in line [65](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L65), to use a boolean array without index or to use iloc instead of loc.

bifbof:
closes #403

### DIFF
--- a/tests/analysis/test_location_identification.py
+++ b/tests/analysis/test_location_identification.py
@@ -253,6 +253,15 @@ class TestLocation_Identifier:
         osna = osna_method(example_osna)
         assert_geodataframe_equal(li, osna)
 
+    def test_pre_filter_index(self, example_freq):
+        """Test if pre_filter=False works with non-serial index"""
+        # issue-#403
+        example_freq.index = *reversed(example_freq.index[1:]), example_freq.index[0]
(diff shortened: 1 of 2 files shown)


henrymartin1 (approved):
Perfect, thanks!

PR #411, opened 2022-05-22
BUG: Error axis argument ``plot_modal_split``

### LINKED ISSUE #379: BUG: error in modal split plot when passing axis
The modal split plotting function throws an error if the `axis` argument is not None. The error occurs in [this line](https://github.com/mie-lab/trackintel/blob/ca4300e6056b8500a27907faca515478ca268524/trackintel/visualization/modal_split.py#L112)
because if an axis object is passed as an argument, the `fig` variable does not exist. A simple fix is probably to put that line into an if clause as well.

bifbof:
closes #379


NinaWie:
Nice fixes and especially nice tests for the plotting functions! Looks good to me, I can't find any problems. Merging :)

PR #418, opened 2022-05-25
BUG: access geometry in smoothing correctly

bifbof:
The `smoothen_triplegs` has the bug that it tries accessing the geometry over the "geom" column. 
This PR fixes this. But generally I would also suggest removing this function as the wrapper is so thin that I don't see the advantage in adding it to trackintel.

### DIFF
--- a/tests/preprocessing/test_triplegs.py
+++ b/tests/preprocessing/test_triplegs.py
@@ -8,12 +8,11 @@ import pytest
 from geopandas.testing import assert_geodataframe_equal
 from pandas.testing import assert_frame_equal, assert_series_equal, assert_index_equal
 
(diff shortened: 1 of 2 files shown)


henrymartin1:
Yes, I agree. The wrapper is really thin and by now it is also implemented in [Geopandas](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.simplify.html#geopandas-geoseries-simplify) (using an equally thin wrapper :D). The implementation comes from a time when the simplification was not included in shapely but we then updated it to shapely when they included it too reduce our dependencies.

henrymartin1 (approved):

PR #419, opened 2022-05-25
BUG: handle geometry names in `get_speed_positionfixes`

bifbof:
Same bug as in #418. "geom" was used instead of `.geometry`.
Additionally replaced the `.apply` methods with vectorized ones and added a test for planar crs.


henrymartin1 (approved):
Nice work! It is a shame that we never implemented benchmarks. Would be nice to see how much faster this method is ;-)

Issue #409, opened 2022-05-20
Cannot rename geometry column with postgis IO functions


henrymartin1:
We are still on the hunt for the postgis integration that allows to read any data directly into a trackintel class using only one line of code ;-) 

We are very close but I noticed another problem. I have a database with staypoints that have a geometry column named `geometry_raw` which I want to rename to `geom`. 
The challange is that this involves three steps: 
- Reading a geopandas dataframe with the geometry column `geometry_raw` from the database
- Renaming the geometry column to `geom`
- Setting the geometry column to `geom` 

Renaming is now supported by using the passing the `mapping` argument down to `read_staypoints_gpd` However, I cannot pass the `geom_col` argument as `**kwds` to `read_staypoints_gpd` because `read_staypoints_postgis` also has a paramter called `geom_col`. 
I think the standard solution is to have an explicit parameter that might be called `read_gpd_kwds` and that is unpacked directly to the wrapped function. 
Examples: 
- [Discussion on SO](https://stackoverflow.com/questions/61746117/how-to-pass-kwargs-with-the-same-name-as-a-positional-argument-of-a-function) 
- See seaborn [histplot](https://seaborn.pydata.org/generated/seaborn.histplot.html#seaborn.histplot) the `kde_kws` parameter

PR #428, opened 2022-06-01
ENH: add `**kwargs` to `plot_modal_split`

### LINKED ISSUE #174: Enhancement for modal_split
- Giving an invalid 'metric' should return an informative error.

bifbof:
closes #174

Small PR that adds description to an argument and passes keyword arguments to `DataFrame.plot.bar`.

### DIFF
--- a/trackintel/visualization/modal_split.py
+++ b/trackintel/visualization/modal_split.py
@@ -19,6 +19,7 @@ def plot_modal_split(
     skip_xticks=0,
     n_col_legend=5,
     borderaxespad=0.5,
+    **kwargs
 ):
     """
     Plot modal split as returned by `trackintel.analysis.modal_split.calculate_modal_split`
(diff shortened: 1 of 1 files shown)


henrymartin1 (commented):
I think it might be better to tunnel the optional keywords with a dedicated dictionary as discussed in #409.

> I think the standard solution is to have an explicit parameter that might be called `read_gpd_kwds` and that is unpacked directly to the wrapped function. 
> Examples: 
> - [Discussion on SO](https://stackoverflow.com/questions/61746117/how-to-pass-kwargs-with-the-same-name-as-a-positional-argument-of-a-function) 
> - See seaborn [histplot](https://seaborn.pydata.org/generated/seaborn.histplot.html#seaborn.histplot) the `kde_kws` parameter

bifbof:
> I think it might be better to tunnel the optional keywords with a dedicated dictionary as discussed in #409.

Agree the this is cleaner.  👍🏼

bifbof:
> @bifbof is this ready to be merged?

Yes it is ready to merge.

PR #430, opened 2022-06-01
ENH: Remove `.apply` in `get_speed_triplegs`

bifbof:
Minor performance increase. Replaced a `apply` call with the direct vectorization.
Additionally replaced the asserts with raised errors as asserts should be used to detect bugs and not to communicate missing parameters. (e.g. https://stackoverflow.com/questions/944592/best-practice-for-using-assert?noredirect=1&lq=1)


hong2223:
Thanks, I will merge this now

PR #431, opened 2022-06-01
ENH: generalize `_generate_locations_per_user`

bifbof:
I want finally to get rid of these shapely warnings in `preprocessing/staypoints`. 
But for that I want to clean up the method first. I saw that we could easily reuse one function.
So this PR reuses the ``_generate_locations_per_user`` function and refactors there a bit.

### DIFF
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -1,5 +1,3 @@
-from math import radians
-
 import numpy as np
 import geopandas as gpd
 import pandas as pd
@@ -95,10 +93,9 @@ def generate_locations(
         if agg_level == "user":
             sp = applyParallel(
                 sp.groupby("user_id", as_index=False),
-                _generate_locations_per_user,
+                _gen_locs_dbscan,
                 n_jobs=n_jobs,
                 print_progress=print_progress,
-                geo_col=geo_col,
                 distance_metric=distance_metric,
                 db=db,
             )
@@ -126,14 +123,7 @@ def generate_locations(
             sp.sort_values(["user_id", "started_at"], inplace=True)
 
         else:
-            if distance_metric == "haversine":
-                # the input is converted to list of (lat, lon) tuples in radians unit
-                p = np.array([[radians(g.y), radians(g.x)] for g in sp.geometry])
-            else:
-                p = np.array([[g.x, g.y] for g in sp.geometry])
-            labels = db.fit_predict(p)
-
-            sp["location_id"] = labels
+            _gen_locs_dbscan(sp, db=db, distance_metric=distance_metric)
 
         ### create locations as grouped staypoints
         temp_sp = sp[["user_id", "location_id", sp.geometry.name]]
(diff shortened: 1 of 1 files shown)


hong2223 (changes requested):
Thanks, I think the simplification is nice.

I only have one comment: you removed the `geo_col` input parameter in the function. I think this related to a geopandas bug we observed [before](https://github.com/mie-lab/trackintel/issues/180#issuecomment-839670952): namely, the geom column cannot be called using `.geometry` after the groupby option. I checked and the [geopandas issue](https://github.com/geopandas/geopandas/issues/1777) is still open. Can you checked if this problem still exist after the change?

bifbof:
> Thanks, I think the simplification is nice.
> 
> I only have one comment: you removed the `geo_col` input parameter in the function. I think this related to a geopandas bug we observed [before](https://github.com/mie-lab/trackintel/issues/180#issuecomment-839670952): namely, the geom column cannot be called using `.geometry` after the groupby option. I checked and the [geopandas issue](https://github.com/geopandas/geopandas/issues/1777) is still open. Can you checked if this problem still exist after the change?

Oh interesting, good catch. I'll write a test and have a look.

bifbof:
So I wrote a little script testing that error and I cannot reproduce it for the `.apply` function.
Here I have a little script to test that.
````python
def fun(gdf):
    print("crs", repr(gdf.geometry.crs))
    print("name", repr(gdf.geometry.name))
    return gdf

world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
world.crs = "EPSG:2056"
world = world.rename(columns={"geometry": "cool_geometry"})
world = world.set_geometry("cool_geometry")
world.groupby("continent").apply(func)
````
It still does not work for `first()` and similar functions that is a bit strange

````python
ret = world.groupby("continent").first()
print(repr(ret.geometry.crs))
````

hong2223:
> So I wrote a little script testing that error and I cannot reproduce it for the `.apply` function. Here I have a little script to test that.
> 
> ```python
> def fun(gdf):
>     print("crs", repr(gdf.geometry.crs))
>     print("name", repr(gdf.geometry.name))
>     return gdf
> 
> world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
> world.crs = "EPSG:2056"
> world = world.rename(columns={"geometry": "cool_geometry"})
> world = world.set_geometry("cool_geometry")
> world.groupby("continent").apply(func)
> ```
> 
> It still does not work for `first()` and similar functions that is a bit strange
> 
> ```python
> ret = world.groupby("continent").first()
> print(repr(ret.geometry.crs))
> ```

What is the status now? Can I understand as the `.geometry` works for some apply functions and failed for others? 

If in general this works for our case I think we can get rid of the "geom" input parameter. This also applies for the input parameter "geo_col" of the "_generate_staypoints_sliding_user()" function

bifbof:
> What is the status now? Can I understand as the `.geometry` works for some apply functions and failed for others?
> 
> If in general this works for our case I think we can get rid of the `geo_col` input parameter. This also applies for the `_generate_staypoints_sliding_user()` function

I would say it works in this case and we can merge it. Functions like `.first()` are implemented in pandas that overrides some geopandas stuff and `.apply()` does not do anything before it is handed to you and then we control it.

hong2223:
> > What is the status now? Can I understand as the `.geometry` works for some apply functions and failed for others?
> > If in general this works for our case I think we can get rid of the `geo_col` input parameter. This also applies for the `_generate_staypoints_sliding_user()` function
> 
> I would say it works in this case and we can merge it. Functions like `.first()` are implemented in pandas that overrides some geopandas stuff and `.apply()` does not do anything before it is handed to you and then we control it.

Thanks! I will merge this after the tests

PR #435, opened 2022-06-08
PERF: faster mode for only floats in haversine distance

bifbof:
This PR increases the performance of `staypoint_generation` by 4x. We achieved that by adding an extra mode to `haversine_distance` that uses `math` functions functions that avoid the overhead of `numpy` functions and is that way 10x faster.
That is of course only possible for float values and not for arrays therefore an extra flag is added that can be set to get this speedup. 

Additionally made some minor cleanups in the function.


hong2223:
Thanks for the clean-up. This looks great and can be merged now.

PR #440, opened 2022-07-07
ENH: Airspeed velocity based continuous integration

### LINKED ISSUE #183: Profiling of trackintel - continuous integration
Based on our discussion, I think it is better to have a memory profiler for trackintel. It will help us ensure that new pull requests do not significantly worsen the memory requirements. Such a profiler will become more relevant as we introduce multithreaded options for trackintel in general.

abcnishant007:
closes #183 
Stats are live at https://mie-lab.github.io/trackintel/


hong2223:
Sorry for the late review :)
From what I saw in the demo, this is great: we finally have some benchmarks to compare the PRs! 

What is currently confusing is maybe a readme on how to maintain this. I read through the related issue and am not sure if I understand everything. The most important points could, for example, be documented in the CONTRIBUTING.md file. Points including:

- how to add more benchmarks/functions?
- is this automatic? if not, how to rerun the benchmarks?
- this needs to be run on a physical computer, and the performance need to be uploaded?
- how can we transfer this to mielab/trackintel.github.io?

abcnishant007:
@hong2223 Thanks for the review! Please find my answers below. 

**1. how to add more benchmarks/functions?** 

> 
> Manually is the best way to go. Why?
> - I read lots of examples where some repositories tried to automatically convert test functions to benchmark functions, but it becomes unnecessary in the long run. Given the fact that for benchmarks, we need bigger input files, unlike the unit tests, running hundreds of benchmarks becomes very slow. 
> - If we must make it automatic, [pytest-benchmarks](https://pypi.org/project/pytest-benchmark/) can be used but the interface and information provided in the output seem rudimentary in comparison to [airspeedvelocity](https://github.com/airspeed-velocity/asv). For example, in the benchmarks which I wrote using the ASV template, I use three aspects, `max_memory`, `returned memory`, and `runtime`. Each of them runs multiple times to get the mean value. There is a common function which does the computation, and then we have three wrapper-like functions which compute these three parameters. 

**2. is this automatic? if not, how to rerun the benchmarks?**

> 
> No, it is not automatic. But it can be automated to some degree depending on the need and urgency. 
> 
> - Imporant information: Two parallel developments are underway. (i) ASV might be available as a github action soon (ii) Github [action benchmark ](https://github.com/benchmark-action/github-action-benchmark ) which currently works only with pytest-benchmark, is likely to include ASV in their automation. It is listed in their future works. If any of them is incorporated in the near future, our job is done. We do not need to reinvent the wheel. 
> - Some repositories who use ASV suggested on stackoverflow to use a sandbox to run this automatically. But most of them are waiting for either of those two developments to be completed. 

**> How to rerun the benchmarks?** 
This is my to-do item. I have all the ingredients scattered in the comments on #183. I can summarise them in an easy to read format. Then I can tag you to check if you can do exactly as in the readme and regenerate the results. Would that work?

**3.  this needs to be run on a physical computer, and the performance need to be uploaded?**
> Yes, needs to run on a physical computer due to the sheer amount of computation.  
> The performance results however, can be uploaded automatically. Still, it will involve some manual work when running on a new computer due to gitHub account setup,  private key etc.. 
> The real solution is possible only when one of those two developments listed above is accomplished. 

**4. how can we transfer this to mielab/trackintel.github.io?** 
> done :) 
> the generated htmls need to be uploaded to gh-pages branch.

hong2223:
Thanks for the answer!

> **How to rerun the benchmarks?**
> 
> > This is my to-do item. I have all the ingredients scattered in the comments on #183. I can summarise them in an easy to read format. Then I can tag you to check if you can do exactly as in the readme and regenerate the results. Would that work?

Yes exactly. I think we need a readme on how to execute the steps.

abcnishant007:
> Thanks for the answer!
> 
> > **How to rerun the benchmarks?**
> > > This is my to-do item. I have all the ingredients scattered in the comments on #183. I can summarise them in an easy to read format. Then I can tag you to check if you can do exactly as in the readme and regenerate the results. Would that work?
> 
> Yes exactly. I think we need a readme on how to execute the steps.

I have crated a readMe which, can be found [here](https://github.com/abcnishant007/trackintel/blob/asv-trackintel/ASV-BENCHMARKING.md). Ideally, you should not need any other input to run it succefully. If you need to look up anything, please keep track of them. I will include them as well. Two items in the readme are temporary and need to be removed after pull request is successful. I have tagged them with your name :) please let me know if you have any trouble running it.

ASV-BENCHMARKING.md
```diff
@@ -0,0 +1,82 @@
+Summarised below are the steps required to run the asv benchmark on a new computer.
+
+1. Clone trackintel and checkout to master. Add the trackintel root to python path 
+
+>(**For testing purposes only,** clone the abcnishant007/trackintel and checkout to `asv-trackintel` branch). 
+*@Ye, we need to edit this line after pull request is approved. *
+
+2. Download the larger geolife data from [this branch](https://github.com/abcnishant007/trackintel/tree/benchmark-files) and place it inside `tests/data/geolife_long_10_MB`. This dataset will not be used by default. The dataset option is hardcoded in the benchmark files as shown [here](https://github.com/abcnishant007/trackintel/blob/7b8c2ee2f12d98b59578cd0519aae6a5240ade4c/benchmarks/preprocessing_benchmarks.py#L6). The hardcoding exists temporarily only to ensure that we can switch to the smaller dataset for asv testing purposes on a new computer. Once the whole setup is complete, we can start using the bigger dataset. The hardcoding can be removed (or retained for future setups) and only the `geolife_long_10_MB` data set should be used for actual benchmarks.
```
hong2223, line 8: I am not sure I understand here. From the line of code, which dataset are we using now and which in the future?

ASV-BENCHMARKING.md
```diff
@@ -0,0 +1,82 @@
...
+>> Solution:
+```
+pip install --upgrade --user pyqtwebengine==5.12
+pip install --upgrade --user pyqt5==5.12
+```
+
+4. Install asv using `pip install asv`
+5. Change current directory to the root of trackintel. 
+6. Get the list of commits in short form (last 7 characters of commit id) which were merge commits in master using the shell command below: 
+```git log | grep -B 1 “Merge” | grep “commit” | sed ‘s/commit //g’ | cut -c1-7 | head -n 3 > commits.txt```
```
hong2223, line 33: I am actually stuck on this line; my command line (mac) returns `s/commit: invalid command code`. What is this line doing?

hong2223 (changes requested):
Hi, thanks for all the work! I tried to follow the steps and got an error in step 6, could you check?

ASV-BENCHMARKING.md
hong2223: I changed the `‘s/commit //g’` to `"s/commit //g"` and it works! But the generated `commit.txt` file does not contain any lines

ASV-BENCHMARKING.md
abcnishant007: @hong2223 The stats which are live at the trackintel link are based on the big 10 MB data set. 
However, for testing purposes, I put an option to use the dataset which is readily available in the tests folder ( the geolife_long dataset) so that the results are obtained faster. I have updated the readMe with the following permalink:
https://github.com/abcnishant007/trackintel/blob/7b8c2ee2f12d98b59578cd0519aae6a5240ade4c/benchmarks/preprocessing_benchmarks.py#L6-L7 
where we assign the dataset name to the variable `bm_dataset`.

ASV-BENCHMARKING.md
abcnishant007: On the bash shell, all quotes should be straight, (single or double doesn't matter). Please feel free to try now. I replaced the version with straight ticks and there is a copy option now. Additionally, the whole document has been updated to put the commands in separate code blocks so that they can be copied with a click instead of drag and click copy. That reduces the chances of error. 

I tested for MacOS
 Screenshot 2022-08-01 at 21 41 10

The 
```
sed s/`some stuff`/`new stuff`/g 
```
replaced the `<stuff>` with the `<new stuff>`.

abcnishant007: > I am actually stuck on this line; my command line (mac) returns `s/commit: invalid command code`. What is this line doing?

This line parses the output from the `git log` and extracts the merge commits from them. Finally, cuts them to the last 7 chracters. Why last 7? because asv did not work with full commits; I have no idea why :(

hong2223 (approved):

PR #446, opened 2022-08-10
BUG: fixed setting crs for tripleg generation

### LINKED ISSUE #442: BUG: crs not set for trips or locations
When I generate trips from staypoints and triplegs, the resulting Geodataframe has no crs defined, even though the crs was set for both staypoints and triplegs. 
e.g.:
```
>> print(sp.crs, legs.crs)
EPSG:2056 EPSG:2056
>> sp, legs, trips = ti.preprocessing.triplegs.generate_trips(act_sp, legs, gap_threshold=15, add_geometry=True)
>> print(trips.crs)
None
```
The same happens for generating locations. This is not intended, is it? For further processing of the data, I'd expect that the crs is adopted.

NinaWie:
closes #442 

As described in #442 , the crs was not set for trips or locations generation in previous versions of trackintel. 

In the meantime, the problem was already solved for location generation (I could not reproduce the error with the new version of trackintel, only with an older version). I therefore just added a test. Also, since the crs can only be set on a dataframe-level, I don't think it's a problem that there are two geometries - the locs crs just needs to be the same as the sp crs in the end.

Regarding the trips generation it is more complicated because they use both sp and triplegs. I therefore added 
1) an error if both crs are set but they contradict (I think we shouldn't concat the geometries into one in that case)
2) a warning if one or both of them are None
3) trips.crs is set equal to whatever crs is available (either sp or tpls or None)

Let me know what you think about that logic.

### DIFF
--- a/tests/preprocessing/test_staypoints.py
+++ b/tests/preprocessing/test_staypoints.py
@@ -245,6 +245,12 @@ class TestGenerate_locations:
         assert loc_dataset_num == 1
         assert loc_user_num == 2
 
+    def test_crs(self, example_staypoints):
+        """Test whether the crs of the output locations is set correctly."""
+        sp = example_staypoints
+        sp, locs = sp.as_staypoints.generate_locations(method="dbscan", epsilon=20, num_samples=1)
+        assert locs.crs == sp.crs
+
     def test_dbscan_min(self):
         """Test with small epsilon parameter."""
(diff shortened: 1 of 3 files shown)


henrymartin1 (approved):
Nice work! I like that it still works but warns if both crs are None. Also thanks for the test for the location crs.

PR #453, opened 2022-12-05
ENH: add centroid method for angles

### LINKED ISSUE #437: BUG: `.centroid` does not wrap for angles
In #431 we discovered an error in calculating the mean angle for a location.
We calculate the average coordinates with `.centroid`. This function assumes that the coordinate system is projected and therefor does not wrap around in a circle.
For example, if we compute the centroid of the multipoint `[(0, 179), (0, -179)]`, we get the point `(0, 0)` instead of the desired `(0, 180)`.
This is equivalent to this problem (https://rosettacode.org/wiki/Averages/Mean_angle), for a geographical explanation see (https://carto.com/blog/center-of-points/).
A suggested solution would be to implement a new function in `util.py`. You can also take a look at https://github.com/mie-lab/trackintel/blob/4f4868b6547bfceb05447e221bb681f8cd389fd7/trackintel/preprocessing/positionfixes.py#L455
That suffers from the same bug.

bifbof:
Fix #437. 
Such that centroids can be calculated for angles that pass over the -180, 180 wrapping border.


tests/preprocessing/test_util.py
```diff
@@ -88,3 +91,15 @@ def test_index_dtype_with_None(self):
...
+
+
+class TestAngleCentroidMultipoints:
+    """Test util method angle_centroid_multipoints"""
+
+    a = Point((130, 45))
+    b = MultiPoint([[160, 10], [-170, 20]])
+    c = MultiPoint([[20, 0], [30, 10], [40, 20]])
+    g = gpd.GeoSeries([a, b, c])
+    g_solution = gpd.GeoSeries([a, Point([175, 15]), Point([30, 10])])
```
henrymartin1, line 103: It would be nice if you (also) use an example from a referenced source. E.g., [https://rosettacode.org/wiki/Averages/Mean_angle](https://rosettacode.org/wiki/Averages/Mean_angle).
Or if your example is already from a credible source than to add the reference

henrymartin1 (commented):
Looks really good, thanks for the fix! 
I have a small comment for the test and didn't you mention in the error report that the same error applies [here](https://github.com/mie-lab/trackintel/blob/4f4868b6547bfceb05447e221bb681f8cd389fd7/trackintel/preprocessing/positionfixes.py#L455). Then we should fix this as well, shouldn't we?

PR #462, opened 2023-02-11
TXT: Fix warnings in tests

hong2223:
Ready for the review!

### DIFF
--- a/trackintel/model/util.py
+++ b/trackintel/model/util.py
@@ -2,6 +2,8 @@ from functools import partial, update_wrapper
 import trackintel as ti
 import numpy as np
 import pandas as pd
+import warnings
+
 from trackintel.geogr.distances import calculate_haversine_length, check_gdf_planar
 from trackintel.geogr.point_distances import haversine_dist
 
@@ -31,7 +33,10 @@ def get_speed_positionfixes(positionfixes):
     g = pfs.geometry
     # get distance and time difference
     if is_planar_crs:
-        dist = g.distance(g.shift(1)).to_numpy()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
+
+            dist = g.distance(g.shift(1)).to_numpy()
     else:
         x = g.x.to_numpy()
         y = g.y.to_numpy()
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -270,7 +270,10 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
 
     # a joined dataframe sp_tpls is constructed to add the columns 'type' and 'next_type' to the 'sp_merge' table
     # concat and sort by time
-    sp_tpls = pd.concat([sp_merge, tpls_merge]).sort_values(by=["user_id", "started_at"])
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
+
+        sp_tpls = pd.concat([sp_merge, tpls_merge]).sort_values(by=["user_id", "started_at"])
     sp_tpls.index.rename(index_name, inplace=True)
     # get information whether the there is a tripleg after a staypoint
     sp_tpls["next_type"] = sp_tpls["type"].shift(-1)
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -273,6 +273,8 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
 
+        # TODO: the warning relates to how we process when no triplegs is provided. In future versions,
+        # we want to seperate the two scenarios. See issue #463.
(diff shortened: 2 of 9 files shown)


trackintel/model/util.py
```diff
@@ -31,7 +33,10 @@ def get_speed_positionfixes(positionfixes):
     g = pfs.geometry
     # get distance and time difference
     if is_planar_crs:
-        dist = g.distance(g.shift(1)).to_numpy()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
```
NinaWie, line 37: Why are the warnings filtered here? Isn't it a problem if the crs is not set for the positionfixes?

trackintel/preprocessing/staypoints.py
```diff
@@ -270,7 +270,10 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
 
     # a joined dataframe sp_tpls is constructed to add the columns 'type' and 'next_type' to the 'sp_merge' table
     # concat and sort by time
-    sp_tpls = pd.concat([sp_merge, tpls_merge]).sort_values(by=["user_id", "started_at"])
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
+
+        sp_tpls = pd.concat([sp_merge, tpls_merge]).sort_values(by=["user_id", "started_at"])
```
NinaWie, line 276: This is related to the issue you opened, right? I think it would be better to have a separate case for the empty tpls case, as you suggested :)

trackintel/preprocessing/triplegs.py
```diff
@@ -132,7 +132,9 @@ def _seperate_ids(row):
     trips.drop(columns=["type", "sp_tpls_id"], inplace=True)  # make space so no overlap with activity "sp_tpls_id"
     # Inserting `gaps` and `user_change` into the dataframe creates buffers that catch shifted
     # "staypoint_id" and "trip_id" from corrupting staypoints/trips.
-    trips_with_act = pd.concat((trips, sp_tpls_only_act, gaps, user_change), axis=0, ignore_index=True)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
+        trips_with_act = pd.concat((trips, sp_tpls_only_act, gaps, user_change), axis=0, ignore_index=True)
```
NinaWie, line 137: Is this also related to empty dataframes?

trackintel/model/util.py
NinaWie: Ah, is this problem due to the g.shift? After it's shifted, there is no crs defined anymore?

NinaWie (commented):
Awesome that so many of the warnings are gone! 
I had a few questions but looks good in general :)

trackintel/preprocessing/triplegs.py
hong2223: I think all dfs are not empty in our test cases. And this case is more complicated :)
This is the initial place that I tried to find what is causing the warning `CRS not set for some of the concatenation inputs.*`, and I still do not know how it is triggered :). The warning persisted when I set all dfs as pandas dataframe, so I guess a df with geom column (with crs) is still considered a gdf for some functions.

I simply silence the warning here, as this warning is only for information purposes, and I think we should not substantially change the code to fix it.

trackintel/preprocessing/staypoints.py
hong2223: Yes! if a df is empty it will be considered as having no crs. setting a crs to an empty df will also cause an error. 

A more natural fix would be separating the cases.

trackintel/model/util.py
hong2223: You might be right. But I am not sure why a warning is raised :), see below (or above) comments.

hong2223 (commented):
Thanks for the review! The places with comments are all "temporary" or "compromising" fixes. These warnings are all for information purposes, and I did not spend too much time figuring out their real causes.

trackintel/model/util.py
NinaWie: Seems like this was a bug in the past but is fixed now: https://github.com/geopandas/geopandas/issues/1648

trackintel/preprocessing/staypoints.py
NinaWie: So you would add the warning for now, and then we work on the issue in another PR?

NinaWie (approved):
So, I had a closer look at this, but actually I can't reproduce any of the "CRS not set for some of the concatenation inputs" - warnings :D So it seems that it's again a version problem. I'm using geopandas 0.10.2, maybe they fixed the crs problems now.
I guess we can leave the catch_warnings() statements and add a comment that this is due to bugs in the old geopandas version? Otherwise it's fine for me to merge

trackintel/preprocessing/staypoints.py
hong2223: Yes! a todo comment will be added

trackintel/model/util.py
hong2223: The warning persists in our test environment that uses geopandas 0.12.2

hong2223 (commented):
This issue might be related, but the warning persists in our test environment that uses geopandas 0.12.2. I can also reproduce it on my location machine with the same version.

NinaWie (approved):
Looks good, thanks! Let's merge :)

PR #468, opened 2023-04-03
Issue 467

### LINKED ISSUE #467: Allow empty geometries in datamodel
The trackintel accessors check the validity of the geometry column but do not allow empty geometries which can occur in datasets with missing tracking information (e.g., #448). 
We have discussed that we want to allow datasets with missing geometries for the main accessor but that we won't support it for now in the all preprocessing functions.

henrymartin1:
fixes #467

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -4,9 +4,19 @@ channels:
   - defaults
 dependencies:
 # - python=3.8
-- geopandas
-- pygeos
+- geopandas>=0.9.0
 - pip:
+  - numpy
+  - matplotlib
+  - pandas
+  - scikit-learn
+  - networkx
+  - pint
+  - geoalchemy2
+  - osmnx
+  - psycopg2
   - sphinx
   - sphinx_rtd_theme
-  
+  - tqdm
+  - similaritymeasures
+  - pygeos>=0.10.0


trackintel/model/positionfixes.py
```diff
@@ -54,19 +56,32 @@ def _validate(obj):
...
-            "Not all geometries are valid. Try x[~ x.geometry.is_valid] " "where x is you GeoDataFrame"
-        )
+        assert (
+            geom_isna | obj.geometry.is_valid
+        ).all(), "Not all geometries are valid. Try x[~ x.geometry.is_valid] where x is you GeoDataFrame"
 
-        if obj.geometry.iloc[0].geom_type != "Point":
-            raise AttributeError("The geometry must be a Point (only first checked).")
+        if "Point" not in obj.geometry.geom_type.values:
+            raise AttributeError("The geometry must be a Point (checked for any occurrence).")
```
NinaWie, line 67: nice idea to change this :) I think it would have to be `assert all(obj.geometry.geom_type.values == "Point")`, right? 
Or probably we would have to consider the missing values again? I just tested it with some other data, seems like the type of an empty point is GeometryCollection, so it would have to be 
```
assert all(obj.geometry.geom_type.isin(["Point", None, "GeometryCollection"]))
```

NinaWie (changes requested):
Nice tests and changes! I added some comments, but one general question: Shouldn't we add some checks in the other functions to check for empty or missing geometries then? Your plan now is to support missing/empty geometries in the data model, but not in the preprocessing functions, right? Then we should probably add checks in the functions generate_staypoints, generate triplegs etc, to avoid that they fail with data that is loaded in the model?

PR #470, opened 2023-04-13
TST: test error fixes with pandas 2.0

hong2223:
Pandas released v2.0 which causes some errors in our test functions:

- `append` function deprecated and shall be replaced by `concat`:

> Removed deprecated Series.append(), DataFrame.append(), use [concat()](https://pandas.pydata.org/docs/reference/api/pandas.concat.html#pandas.concat) instead ([GH35407](https://github.com/pandas-dev/pandas/issues/35407))

- More flexible dtypes for pandas.index, [link](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#index-can-now-hold-numpy-numeric-dtypes)

In addition, `infer_datetime_format` argument is depricated:

> Deprecated argument infer_datetime_format in [to_datetime()](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html#pandas.to_datetime) and [read_csv()](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html#pandas.read_csv), as a strict version of it is now the default ([GH48621](https://github.com/pandas-dev/pandas/issues/48621))

### DIFF
--- a/trackintel/preprocessing/util.py
+++ b/trackintel/preprocessing/util.py
@@ -107,7 +107,11 @@ def _explode_agg(column, agg, orig_df, agg_df):
     temp = agg_df.explode(column)
     temp = temp[temp[column].notna()]
     temp.index = temp[column]
-    return orig_df.join(temp[agg], how="left")
+
+    return_df = orig_df.join(temp[agg], how="left")
+    # ensure index dtype the same as input
+    return_df.index = return_df.index.astype(orig_df.index.dtype)
+    return return_df
 
 
 def angle_centroid_multipoints(geometry):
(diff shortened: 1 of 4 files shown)


trackintel/preprocessing/util.py
```diff
@@ -107,7 +107,11 @@ def _explode_agg(column, agg, orig_df, agg_df):
     temp = agg_df.explode(column)
     temp = temp[temp[column].notna()]
     temp.index = temp[column]
-    return orig_df.join(temp[agg], how="left")
+
+    return_df = orig_df.join(temp[agg], how="left")
```
hong2223, line 111: As dtype for index is now more flexible, the join function will create an `object` dtype index, even if the index does not change for `orig_df`. We add another line here to ensure consistency of index dtype

henrymartin1: good catch

henrymartin1 (approved):
Thanks for the fix!

PR #477, opened 2023-06-27
ENH: replace pygeos with calls to shapely 2.0

bifbof:
With pygeos being integrated into shapely we don't need this dependency anymore.
Thus this PR removes all calls to pygeos.


hongyeehh:
Thanks for the fix. This is great; we shall push a new version soon and update the conda dependencies accordingly.

hongyeehh (approved):

PR #478, opened 2023-06-28
ENH: remove as_staypoints calls from location_identification

### LINKED ISSUE #476: `obj.as_xyz` does not necessarily validate data
The pattern `sp.as_staypoints` is regularly used in the codebase to verify that the data is valid.
It is intended that it calls the [`StaypointsAccessor`](https://github.com/mie-lab/trackintel/blob/11ff487b1596d0da5f346446df0fd340df433210/trackintel/model/staypoints.py#L46) `__init__` function, which in turn calls the `_validate` function.
But the accessor classes are cached ([here](https://github.com/pandas-dev/pandas/blob/1a23779f09abc6ebf908d66ee88b973b767e2e3c/pandas/core/accessor.py#L151)). Thus the `StaypointsAccessor` class is only created once per object and all follow-up `sp.as_staypoints` calls will always suceed.

Let's consider the following example
```python
import pandas as pd
import geopandas as gpd
import trackintel as ti

p1 = Point(8.5067847, 47.4)
t1 = pd.Timestamp("1971-01-01 04:00:00", tz="utc")
pfs = gpd.GeoDataFrame(data=[{"user_id": 0, "tracked_at": t1, "geom": p1}], geometry="geom", crs="EPSG:4326")
pfs.index.name = "id"
pfs.as_positionfixes # does call "_verify"
pfs.drop(columns="tracked_at", inplace=True)
pfs.as_positionfixes # should not succeed as column "tracked_at" is required
```

Here the second assert goes through although the data is not valid anymore.

This leads to two potential bugs ([here](https://github.com/mie-lab/trackintel/blob/11ff487b1596d0da5f346446df0fd340df433210/trackintel/analysis/location_identification.py#L52) and [here](https://github.com/mie-lab/trackintel/blob/11ff487b1596d0da5f346446df0fd340df433210/trackintel/analysis/location_identification.py#L129)).
In this two places we should call `_validate` manually.

bifbof:
closes #476
This PR removes the `as_staypoints` assertions, as the staypoints need to be created initially and there this property is cached.
It doesn't really solve the verify problem, but my argument is that we require a `GeoDataFrame (as trackintel staypoints)` and thus can rely that we get the right data shape.
In all other places `as_xyz` works as intended, since it is always applied to newly created objects and thus is not yet cached.
However, I would like to make this a bit clearer with a proper class hierarchy (which is backwards compatible with our current `as_xyz` approach), but we should discuss that elsewhere.

### DIFF
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -48,9 +48,6 @@ def location_identifier(staypoints, method="FREQ", pre_filter=True, **pre_filter
     >>> from ti.analysis.location_identification import location_identifier
     >>> location_identifier(staypoints, pre_filter=True, method="FREQ")
     """
-    # assert validity of staypoints
-    staypoints.as_staypoints
-
     sp = staypoints.copy()
     if "location_id" not in sp.columns:
         raise KeyError(
@@ -125,9 +122,6 @@ def pre_filter_locations(
     >>> mask = pre_filter_locations(staypoints)
     >>> staypoints = staypoints[mask]
     """
-    # assert validity of staypoints
-    staypoints.as_staypoints
-
     sp = staypoints.copy()
     if isinstance(thresh_loc_time, str):
         thresh_loc_time = pd.to_timedelta(thresh_loc_time)


hongyeehh:
Thanks. I agree that we shall not check for validity, as we already stated in the documentation. I will merge this PR now.

hongyeehh (approved):

PR #480, opened 2023-06-28
ENH: add development environments

### LINKED ISSUE #407: We should have a list of dependencies required for development
Idealy we would also have environment_dev and requirements_dev files. We can use this issue to collect dependencies.

Here are some: 
- sphinx
- pytest
- black 
- sphinx_rtd_theme

bifbof:
closes #407 
Found a development environment file really useful. So I copied the style from geopandas and added it here.
If there are some packages missing, I'll gladly add them.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -1,0 +1,24 @@
+name: trackintel-dev
+channels:
+  - conda-forge
+  - defaults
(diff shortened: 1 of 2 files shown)


henrymartin1 (approved):
Fully agree!

PR #481, opened 2023-06-28
ENH: remove warning suppression

### LINKED ISSUE #463: Warning raised in trackintel.preprocessing.staypoints.merge_staypoint()
The function accepts sp and tpls as input. But if an `empty` or `pd.DataFrame` or `gpd.Geodataframe` tpls is given, this [line](https://github.com/mie-lab/trackintel/blob/136ad8b045bcd1e50258a4ff7371d2d9fb86b904/trackintel/preprocessing/staypoints.py#L273) will raise a warning as tpls has no crs. 

> UserWarning: CRS not set for some of the concatenation inputs. Setting output's CRS as WGS 84 (the single non-null crs provided).

In general, we do not enforce checking the correctness of sp and tpls inputs, and we explicitly mention tpls can be empty in the docstring. 

It would be good to check the format of sp and separate the scenarios in the code when tpls is given as empty (maybe set input `tpls=None` by default).

bifbof:
As of geopandas 0.12.0, these warnings have been fixed and should no longer be suppressed.
Further should close #463 as the discussed problem does not occur anymore, can be tested with this code:
````python
import geopandas as gpd
import pandas as pd

gdf = gpd.GeoDataFrame([{"a":1, "g": Point(1, 2)}], geometry="g", crs=4326)
gdf_empty = gpd.GeoDataFrame()
gdf_concat = pd.concat([gdf, gdf_empty])
assert gdf.crs == gdf_concat.crs
````

### DIFF
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -271,6 +271,7 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
     # a joined dataframe sp_tpls is constructed to add the columns 'type' and 'next_type' to the 'sp_merge' table
     # concat and sort by time
     sp_tpls = pd.concat([sp_merge, tpls_merge]).sort_values(by=["user_id", "started_at"])
+    # TODO: we want to make tpls as argumtent optional and adapt this logic here. See issue #463.
     sp_tpls.index.rename(index_name, inplace=True)
     # get information whether the there is a tripleg after a staypoint
     sp_tpls["next_type"] = sp_tpls["type"].shift(-1)
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -270,12 +270,7 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
 
     # a joined dataframe sp_tpls is constructed to add the columns 'type' and 'next_type' to the 'sp_merge' table
     # concat and sort by time
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
-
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: trackintel/preprocessing/staypoints.py [hunk 1]
```diff
@@ -270,12 +270,7 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
 
     # a joined dataframe sp_tpls is constructed to add the columns 'type' and 'next_type' to the 'sp_merge' table
     # concat and sort by time
-    with warnings.catch_warnings():
-        warnings.filterwarnings("ignore", message="CRS not set for some of the concatenation inputs.*")
-
-        # TODO: the warning relates to how we process when no triplegs is provided. In future versions,
-        # we want to seperate the two scenarios. See issue #463.
```
hongyeehh, line 277: The issue #463 concerns the seperation of the tripleg scenarios, which is not fix by this PR

hongyeehh:
Thanks for taking care of the warnings. The warning fixes are fine, I would just keep the issue #463 open (and the corresponding comments in the code) as it is not entirely solved

trackintel/preprocessing/staypoints.py
bifbof: I do not understand the problem here. Empty GeoDataFrames and `Nones` are both possible to concat without a warning.

hongyeehh: The issue is not only about the warning - we want to change the default input of the function such that it does not require `triplegs` as input. But I think #463's title is confusing and we shall open a new issue for that

hongyeehh (approved):

trackintel/preprocessing/staypoints.py
bifbof: Ah, that makes sense. Thanks for the clarification. I have added a TODO again.

PR #482, opened 2023-06-29
DOC: change contributing file for performance benchmarking

hongyeehh:
Update the instructions on Performance benchmarking and the usage of the `gh-pages` branch.


abcnishant007 (approved):
The link works, hosted on the mie-lab user instead of abcnishant007; Approving

Issue #491, opened 2023-07-26
Consistent handling of optional columns


bifbof:
The Trackintel model has two types of columns that it accesses.
The required ones, without which you can't create a Trackintel object, and the optional ones mentioned in the documentation.
The handling of the optional columns is rather confusing
- some like `accuracy` are mentioned, but nowhere used with any function
- some like `trips` in tours are not mentioned at all, although they are returned every time by the `generate_tours` function
- no information is given about what type these optional types should be
- basically no tests are written for them, e.g. for io functions

We should treat them consistently and define: which ones are really used, their datatypes etc.

PR #492, opened 2023-07-26
BUG: Read csv lists correctly

### LINKED ISSUE #485: List of trip ids in a tour dataframe is unstructured when reading it from csv.
similar to  #443 but easier to fix.
Tours store a list of the trips they aggregate in the field "trips". Unfortunately the datatype of this list is not stable when writing+reading it from a csv file.
This can be easily fixed in the function `ti.read_tours_csv` by adding the following line if the column `trips` is present
```python
tours['trips'] = tours['trips'].apply(eval)
```

bifbof:
fixes #485
Used `ast.literal_eval` as `eval` is a bit dangerous.
But I think we need to communicate that this is done for some reason, or that `trips` is a optional column.
This relates to #491.

### DIFF
--- a/tests/io/test_file.py
+++ b/tests/io/test_file.py
@@ -2,6 +2,7 @@ import filecmp
 import os
 import pytest
 import pandas as pd
+from pandas.testing import assert_frame_equal
 
 import trackintel as ti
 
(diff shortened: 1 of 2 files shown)


hongyeehh (approved):
Thanks for the solution. This is a nice fix for the issue. 
And yes I agree we need to find a way to handle optional columns: here "trips" does not appear in any input parameters or docstrings. But let's discuss that in #491

PR #493, opened 2023-07-28
ENH: add kwarg to generate_locations to filter for activities

### LINKED ISSUE #427: Add option to generate locations only based on activity staypoints
I think that the way we define staypoints it would make a lot of sense to create locations only based on staypoints that are activities. For this we could include an `activity_only` option in the `generate_location` function.

E.g., in the paper we write: 
```
Activity staypoints represent individual visits at places that are significant to the visitor. Trackintel models
these significant places using the location class to enable the characterization of the place that is visited.
```

bifbof:
closes #427

### DIFF
--- a/tests/preprocessing/test_staypoints.py
+++ b/tests/preprocessing/test_staypoints.py
@@ -454,6 +454,25 @@ class TestGenerate_locations:
             example_staypoints.as_staypoints.generate_locations(method="unknown")
             assert error_msg == str(e.value)
 
+    def test_activity_flag(self, example_staypoints):
+        """Test if only activity staypoints are used if flag is set."""
+        # take out staypoint 6 that should have been merged with 2, 15
+        sp = example_staypoints
+        data = [True, True, True, True, False, True, True, True]
+        idx = [1, 2, 3, 5, 6, 7, 15, 80]
+        activities = pd.Series(data, index=idx)
+        sp["activity"] = activities
+        sp, _ = sp.as_staypoints.generate_locations(
+            method="dbscan",
+            epsilon=10,
+            num_samples=2,
+            distance_metric="haversine",
+            agg_level="user",
+            activities_only=True,
+        )
+        assert sp.loc[1, "location_id"] == sp.loc[15, "location_id"]
+        assert sp.loc[2, "location_id"] is pd.NA
+
 
 class TestMergeStaypoints:
     def test_merge_staypoints(self, example_staypoints_merge):
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -17,6 +17,7 @@ def generate_locations(
     agg_level="user",
     print_progress=False,
     n_jobs=1,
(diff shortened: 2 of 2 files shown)


trackintel/preprocessing/staypoints.py
```diff
@@ -17,6 +17,7 @@ def generate_locations(
     agg_level="user",
     print_progress=False,
     n_jobs=1,
+    activities_only=False,
```
hongyeehh, line 20: imo this parameter should appear right after `agg_level` as it affects the generation (and not like `n_jobs` that only control the performance)

trackintel/preprocessing/staypoints.py
```diff
@@ -75,6 +80,10 @@ def generate_locations(
 
     # initialize the return GeoDataFrames
     sp = staypoints.copy()
+    non_activities = None
+    if activities_only:
+        non_activities = sp[~sp["activity"]]
```
hongyeehh, line 85: I am unsure if this will raise an error (and how informative the error message is for this specific function) if `sp` does not have an `activity` column. Do we need to write our own `try...except` statement for the failure situation?

tests/preprocessing/test_staypoints.py
```diff
@@ -454,6 +454,25 @@ def test_method_error(self, example_staypoints):
...
+        sp, _ = sp.as_staypoints.generate_locations(
+            method="dbscan",
+            epsilon=10,
+            num_samples=2,
+            distance_metric="haversine",
+            agg_level="user",
+            activities_only=True,
+        )
+        assert sp.loc[1, "location_id"] == sp.loc[15, "location_id"]
+        assert sp.loc[2, "location_id"] is pd.NA
```
hongyeehh, line 474: Would be great to check if the type of `location_id` is the same for `activity sp not assigned to locations` and `non-activity sp`.

hongyeehh: I just found [All NA-like values are replaced with pandas.NA](https://pandas.pydata.org/docs/user_guide/integer_na.html), as we explicitly transformed the datatype to `Int64` in Line 190. 

So this issue is solved.

trackintel/preprocessing/staypoints.py
bifbof: Yes I agree that would be nicer. 
But my thought process behind that was that it might breaks code that only uses positional arguments and not keyword arguments.
````python
generate_locations(sp, "dbscan", 100, 1, "haversine", "user", True)
````
would behave differently for example.
But I would guess that no code was written that way where that would apply :D

trackintel/preprocessing/staypoints.py
bifbof: Yes, I agree a more informative error would be nicer.

trackintel/preprocessing/staypoints.py
bifbof: I changed the order of the keyword arguments.

hongyeehh (approved):

PR #494, opened 2023-07-29
ENH: separate read_gpd kwargs into own variable

### LINKED ISSUE #409: Cannot rename geometry column with postgis IO functions

bifbof:
closes #409 

Split up kwargs into own variable.
Test already existed for this.

### DIFF
--- a/trackintel/io/postgis.py
+++ b/trackintel/io/postgis.py
@@ -50,7 +50,7 @@ def read_positionfixes_postgis(
     parse_dates=None,
     params=None,
     chunksize=None,
-    **kwargs
+    read_gpd_kws=None,
 ):
     """Reads positionfixes from a PostGIS database.
 
@@ -92,7 +92,7 @@ def read_positionfixes_postgis(
         If specified, return an iterator where chunksize is the number
         of rows to include in each chunk.
 
-    **kwargs
+    read_gpd_kws : dict, default None
         Further keyword arguments as available in trackintels trackintel.io.read_positionfixes_gpd().
         Especially useful to rename column names from the SQL table to trackintel conform column names.
         See second example how to use it in code.
@@ -106,7 +106,7 @@ def read_positionfixes_postgis(
     --------
     >>> pfs = ti.io.read_positionfixes_postgis("SELECT * FROM positionfixes", con, geom_col="geom")
     >>> pfs = ti.io.read_positionfixes_postgis("SELECT * FROM positionfixes", con, geom_col="geom",
-    ...                                        index_col="id", user_id="USER", tracked_at="time")
+    ...                                        index_col="id", user_id="USER", {tracked_at: "time"})
     """
     pfs = gpd.GeoDataFrame.from_postgis(
         sql,
@@ -119,7 +119,7 @@ def read_positionfixes_postgis(
         params=params,
         chunksize=chunksize,
     )
-    return ti.io.read_positionfixes_gpd(pfs, **kwargs)
+    return ti.io.read_positionfixes_gpd(pfs, **(read_gpd_kws or {}))
 
 
 @_handle_con_string
@@ -149,7 +149,7 @@ def read_triplegs_postgis(
     parse_dates=None,
     params=None,
     chunksize=None,
-    **kwargs
+    read_gpd_kws=None,
 ):
     """Reads triplegs from a PostGIS database.
 
@@ -191,7 +191,7 @@ def read_triplegs_postgis(
         If specified, return an iterator where chunksize is the number
         of rows to include in each chunk.
 
-    **kwargs
+    read_gpd_kws : dict, default None
         Further keyword arguments as available in trackintels trackintel.io.read_triplegs_gpd().
         Especially useful to rename column names from the SQL table to trackintel conform column names.
         See second example how to use it in code.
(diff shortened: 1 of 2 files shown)


trackintel/io/postgis.py
```diff
@@ -50,7 +50,7 @@ def read_positionfixes_postgis(
     parse_dates=None,
     params=None,
     chunksize=None,
-    **kwargs
+    read_gpd_kws=None,
```
NinaWie, line 53: I think it would be more clean if we set the defaul to `{}`?  Then in the code below, we could drop the `(read_gpd_kws or {})` and just put `read_gpd_kws` instead.

trackintel/io/postgis.py
```diff
@@ -106,7 +106,8 @@ def read_positionfixes_postgis(
     --------
     >>> pfs = ti.io.read_positionfixes_postgis("SELECT * FROM positionfixes", con, geom_col="geom")
     >>> pfs = ti.io.read_positionfixes_postgis("SELECT * FROM positionfixes", con, geom_col="geom",
-    ...                                        index_col="id", user_id="USER", tracked_at="time")
+    ...                                        index_col="id", user_id="USER",
```
NinaWie, line 109: Shouldn't `user_id="USER"` be part of the `read_gpd_kws` dictionary?? `user_id` is not an keyword argument of the function, so I assume the code would not work like this

trackintel/io/postgis.py
```diff
@@ -92,7 +92,7 @@ def read_positionfixes_postgis(
         If specified, return an iterator where chunksize is the number
         of rows to include in each chunk.
 
-    **kwargs
+    read_gpd_kws : dict, default None
         Further keyword arguments as available in trackintels trackintel.io.read_positionfixes_gpd().
         Especially useful to rename column names from the SQL table to trackintel conform column names.
         See second example how to use it in code.
```
NinaWie, line 98: I'm a bit confused by the name `read_gpd_kws` - can you put any other kwargs instead of renaming? Because if it's only used for renaming columns, then I would call it `rename_columns` or `rename_dict` or so. If we call it `read_gpd_kws`, it sounds like this dictionary is passed on to the `geopandas.from_postgis` function, but it's not, right?

NinaWie (commented):
Thanks for taking care of this! I have a few comments regarding the argument name and the default value, which I would set to `rename_columns={}` or so instead of the current state. Let me know if this makes sense :)

trackintel/io/postgis.py
bifbof: Mutable datatypes as default arguments is a python pitfall.
Functions are only initiated once, thus all function calls share the same object.
This can cause behavior like this
````python
>>> def foo(a=[]):
...     a.append(1)
...     print(a)
...
>>> foo()
[1]
>>> foo()
[1, 1]
````

trackintel/io/postgis.py
bifbof: good catch :)

trackintel/io/postgis.py
bifbof: Yeah, I had some problems naming it. It can be used for all the keyword arguments of the `read_xyz_gpd` functions.
Which is mostly used to rename columns, it can be used for other things but that wouldn't make sense with the interaction between the postgis and the gpd function. But if you want to rename other columns then one has to past an extra mapper in.

trackintel/io/postgis.py
NinaWie: Interesting, didn't know this! I guess it wouldn't matter in our case since we are not mutating the dictionary anywhere in the function, but I'm fine with the current version if you think it's better code style :)

trackintel/io/postgis.py
NinaWie: I see, I guess we can't avoid that users have to look at the read_xzy_gpd functions in order to set these kw arguments. Fine by me

NinaWie (approved):
As said in the comments, I think the current version is fine :)

trackintel/io/postgis.py
bifbof: Yes, I would keep it this way. More as an argument of code style then something else. :)

PR #496, opened 2023-07-30
ENH: Store trips list as JSON in postgis

### LINKED ISSUE #443: List of trip ids in a tour dataframe is unstructured when reading it from postgis.
Tours store a list of the trips they aggregate in the field "trips". Unfortunately the datatype of this list is not stable when writing+reading it from a postgis database. In this case it is transformed to a string of a dictionary, e.g., `'{72066,72067,72068,72069,72070}'`
It would be great if the trackintel reading/writing functions could somehow handle this. If there is no better way, the read tours from postgis function could include a line like this to parse the field properly: 
`tours["trips"] = tours["trips"].apply(lambda x: eval(x.replace("{", "[").replace("}", "]")))`

bifbof:
closes #443

### DIFF
--- a/tests/io/test_postgis.py
+++ b/tests/io/test_postgis.py
@@ -716,6 +716,23 @@ class TestTours:
         finally:
             del_table(conn, table)
 
+    def test_trips_column(self, example_tours, conn_postgis):
+        """Test if list of trips is read correctly."""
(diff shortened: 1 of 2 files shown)


NinaWie (approved):
Nice and clean solution!

PR #497, opened 2023-07-31
BUG: Handle empty data in FREQ location_identifier

### LINKED ISSUE #433: location_identifier raises key error if all staypoints are pre filtered
If the tracking quality is bad and all staypoints are pre-filtered, then the variable `method_val` does not get the column purpose and therefore [the line 74](https://github.com/mie-lab/trackintel/blob/master/trackintel/analysis/location_identification.py#L74) raises a key error. 

I think the behavior should be that the staypoints are returned with a nan-filled or empty purpose column.

bifbof:
closes #433

### DIFF
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -197,6 +197,8 @@ def freq_method(staypoints, *labels):
             group["duration"] = group["finished_at"] - group["started_at"]
         # pandas keeps inner order of groups
         sp.loc[sp["user_id"] == name, "purpose"] = _freq_transform(group, *labels)
+    if "purpose" not in sp.columns:  # if empty sp
+        sp["purpose"] = None
     return sp
 
 
--- a/tests/analysis/test_location_identification.py
+++ b/tests/analysis/test_location_identification.py
@@ -182,6 +182,13 @@ class TestFreq_method:
         assert freq["purpose"].count() == example_freq["purpose"].count()
         assert_geodataframe_equal(example_freq, freq)
 
+    def test_empty_sp(self, example_freq):
+        """Test if empty sp also get purpose column."""
+        example_freq.drop(example_freq.index, inplace=True)
+        freq = freq_method(example_freq)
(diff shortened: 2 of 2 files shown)


trackintel/analysis/location_identification.py
```diff
@@ -197,6 +197,8 @@ def freq_method(staypoints, *labels):
             group["duration"] = group["finished_at"] - group["started_at"]
```
hongyeehh, line 197: It's not related to this PR, but I guess this line can be moved out of the `for name, group in sp.groupby("user_id"):` for better readability?

trackintel/analysis/location_identification.py
```diff
@@ -197,6 +197,8 @@ def freq_method(staypoints, *labels):
...
         # pandas keeps inner order of groups
         sp.loc[sp["user_id"] == name, "purpose"] = _freq_transform(group, *labels)
+    if "purpose" not in sp.columns:  # if empty sp
+        sp["purpose"] = None
```
hongyeehh, line 201: Still related to the data type: shall we use `None`, `pd.NA` or `np.nan`? 

Edit: I think we need to use `pd.NA` for a consistent datatype of `Int64`.

hongyeehh (changes requested):

trackintel/analysis/location_identification.py
bifbof: I think I did it so I wouldn't have to delete it from the `sp` `DataFrame` later, but I'm not sure.

trackintel/analysis/location_identification.py
bifbof: As the data type of "purpose" is `object` all three should work. But as `pd.NA` is intended for `Int64` and `np.nan` for `float`, I thought that `None` is good for `object`.

trackintel/analysis/location_identification.py
hongyeehh: good point, I'll resolve this comment

PR #498, opened 2023-08-01
TST: add test for no extent column in locations_csv

### LINKED ISSUE #391: read_locations_csv and write_locations_csv have no tests for the extent keyword
read_locations_csv and write_locations_csv both use extent as keyword but there is no test assuring it works.

bifbof:
closes #391 
Default tests already check for `extent`, so I added a little extra test to check if it also works without it.

### DIFF
--- a/tests/io/test_file.py
+++ b/tests/io/test_file.py
@@ -231,7 +231,7 @@ def example_locations():
     ]
     locs = gpd.GeoDataFrame(data=list_dict, geometry="center", crs="EPSG:4326")
     locs.index.name = "id"
-    assert locs.as_locations
+    locs.as_locations
     return locs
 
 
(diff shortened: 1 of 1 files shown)


tests/io/test_file.py
```diff
@@ -213,6 +217,24 @@ def test_set_index(self):
...
+    p3 = Point(8.5067847, 47.6)
+
+    list_dict = [
+        {"user_id": 0, "center": p1},
+        {"user_id": 0, "center": p2},
+        {"user_id": 1, "center": p3},
+    ]
+    locs = gpd.GeoDataFrame(data=list_dict, geometry="center", crs="EPSG:4326")
+    locs.index.name = "id"
+    assert locs.as_locations
```
hongyeehh, line 234: We once realized to only write `locs.as_locations` (delete the assert), see #215 and #255

hongyeehh (changes requested):

hongyeehh (approved):

PR #499, opened 2023-08-01
TST: remove unnecessary assert statements from tests

bifbof:
In #498 I copied `assert locs.as_locations` from another test. But we decided in #215 to use `locs.as_locations` to check validity.
To never copy the old style again, this PR removes all of it from the tests (and in addition I removed some `assert True`).
As a sidenote, I think some validity checks don`t work, as discussed in #476, but most are fine.


hongyeehh (approved):
Thanks! Merging!

PR #502, opened 2023-08-02
ENH: warn in `read_postgis_` functions if no index column is provided

### LINKED ISSUE #330: Postgis should warn when no index column is provided
The index of a dataframe plays a very important role in trackintel. Therefore, we included a warning if no explicit index column is provided and the default index is assumed when csv files are read e.g., [here](https://github.com/mie-lab/trackintel/blob/563473b885a91858a8463da9db66353d5adb1089/trackintel/io/file.py#L176)
I think it would be very helpful to have the same functionality for the read postgis functions (forgetting to set the index column actually caused errors for me several times...)

bifbof:
closes #330 
I moved the decorator for the `read_csv` function to `io/util.py` to make it accessible, and then used it for the postgis functions as well.
Alternatively, we could move it to `io/from_geopandas.py` to avoid creating a new file.

### DIFF
--- a/tests/io/test_util.py
+++ b/tests/io/test_util.py
@@ -1,0 +1,23 @@
+import pytest
+from trackintel.io.util import _index_warning_default_none
+
+
+class Test_index_warning_default_none:
+    def test_index_set(self):
+        """Test if a set index creates no warning."""
+
+        @_index_warning_default_none
+        def foo(index_col=None):
+            return
+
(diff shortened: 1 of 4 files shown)


hongyeehh:
I am fine with adding a new file if that helps. @henrymartin1 can review the postGIS parts

henrymartin1 (approved):
Nice and clean solution! 
Thank you

PR #505, opened 2023-08-10
ENH: remove caching in DataFrame accessors

bifbof:
We agreed to remove caching for DataFrame accessors to allow subtyping, as suggested in #490.
This creates overhead every time we call `.as_xyz`, but also provides more security since the model is now truly assured every time we use it this way.
It was necessary to adjust some tests so that they do not violate the trackintel models.


hongyeehh (approved):
Nice and neat, thanks!

PR #513, opened 2023-08-16
ENH: Rework visualization module

### LINKED ISSUE #506: Move `as_positionfixes.plot` method into separate function
with #490 we will subclass `GeoDataFrame` and thus [this method](https://github.com/mie-lab/trackintel/blob/a0e989107868edc8fc8c680920b0a0b9e4754b14/trackintel/model/positionfixes.py#L96) will override the inherited plot function. Access to GeoPanda's plot tools will then only be possible via some direct calls to the GeoDataFrame object.
```python
GeoDataFrame.plot(pfs)(*args, **kwargs)
```
This behavior is awkward.
I would suggest removing the override method from the class and putting it in a separate function.
This will also removes the need for the cascading plot functionality currently employed, drastically simplifying the function.
Alternatively, you could rename the overriding methods.

bifbof:
This one turned out to be bigger than expected. 
closes #506 
My first motivation was to remove the `positionfixes.plot` override. 
But I realized that with all these plot functions cascading calls to each other, it would be much easier to put them into one function that does everything in one go. So I created `ti.plot`.
With only one function to remember, it is also easier to use trackintel.
The rest of the module was so small that I decided to put them in one file. The resulting file is still quite small at 465 LOC.
So a big part of this PR is just collecting and cleaning things up.
But I know I did a lot in one go, so if you want me to split up some changes or decide to do something else, let me know :)


hongyeehh (approved):
Look fine to me; thanks for the cleanup and change! 
I would request another reviewer for this large PR.

trackintel/visualization/plotting.py
```diff
@@ -0,0 +1,465 @@
...
+        radius = meters_to_decimal_degrees(radius_locs, center_latitude)
+        for pt in locations.to_dict("records"):
+            circle = mpatches.Circle(
+                (pt["center"].x, pt["center"].y), radius, facecolor="none", edgecolor="r", zorder=4
+            )
+            ax.add_artist(circle)
+
+
+def plot(
+    *,
```
NinaWie, line 279: Why the star?

NinaWie (approved):
Great effort! I think this cleanup makes the plotting a lot nicer!

trackintel/visualization/plotting.py
bifbof: I wanted to enforce the keyword arguments. But yes that is maybe not necessary. I'll change it :)

PR #515, opened 2023-08-23
DOC: Include `pip` downloads

abcnishant007:
This is just in case we wish to showcase our downloads; Currently it stands at around 32K, which, I think, is a decent number.
Please feel free to reject if it is too early :)


hongyeehh:
Thanks for adding the new badge! Merging now

hongyeehh (approved):

PR #518, opened 2023-09-05
ENH: Add subclassing for trips

bifbof:
Adding the Trips classes to trackintel.
I changed the accessor to the `trips` function to create either the `TripsDataFrame` or `TripsGeoDataFrame` class, depending on whether a `geocolumn` is present or not.
When I implemented these changes, I sometimes thought it would be easier to implement a single class, aka a `Trips` class with no fallback mechanism, as Nina suggested in a meeting. So that it remains a `Trips` class even if it loses the geometry column. I would be open to discuss if this alternative implementation would be more appropriate.
In another PR, I should add documentation explaining this split into two classes and the complicated fallback pattern.

### DIFF
--- a/tests/model/test_util.py
+++ b/tests/model/test_util.py
@@ -252,7 +252,7 @@ class Test_wrapped_gdf_method:
             return gdf.drop(columns=gdf.geometry.name)
 
         foo = _wrapped_gdf_method(foo)
-        assert type(foo(example_positionfixes)) == pd.DataFrame
+        assert type(foo(example_positionfixes)) is pd.DataFrame
 
     def test_failed_check(self, example_positionfixes):
         """Test if _check fails then __class__ is not touched"""
@@ -267,7 +267,7 @@ class Test_wrapped_gdf_method:
             return GeoDataFrame(a)
 
         a = A(example_positionfixes)
-        assert type(foo(a)) == GeoDataFrame
+        assert type(foo(a)) is GeoDataFrame
 
     def test_keep_class(self, example_positionfixes):
         """Test if original class is restored if return value subclasses GeoDataFarme and fulfills _check"""
@@ -282,7 +282,7 @@ class Test_wrapped_gdf_method:
             return GeoDataFrame(a)
 
         a = A(example_positionfixes)
-        assert type(foo(a)) == A
+        assert type(foo(a)) is A
 
 
 class Test_wrapped_gdf_method_fallback:
@@ -295,7 +295,7 @@ class Test_wrapped_gdf_method_fallback:
             return pd.Series(gdf.iloc[0])
 
         foo = _wrapped_gdf_method(foo)
-        assert type(foo(example_positionfixes)) == pd.Series
+        assert type(foo(example_positionfixes)) is pd.Series
 
     def test_failed_check(self, example_positionfixes):
         """Test if _check fails then __class__ is not touched"""
@@ -310,7 +310,7 @@ class Test_wrapped_gdf_method_fallback:
             return GeoDataFrame(a)
 
         a = A(example_positionfixes)
-        assert type(foo(a)) == GeoDataFrame
+        assert type(foo(a)) is GeoDataFrame
 
     def test_keep_class(self, example_positionfixes):
         """Test if original class is restored if return value subclasses GeoDataFarme and fulfills _check"""
@@ -325,7 +325,7 @@ class Test_wrapped_gdf_method_fallback:
             return GeoDataFrame(a)
 
         a = A(example_positionfixes)
-        assert type(foo(a)) == A
+        assert type(foo(a)) is A
 
     def test_fallback(self, example_positionfixes):
         class B(pd.DataFrame):
@@ -343,7 +343,7 @@ class Test_wrapped_gdf_method_fallback:
             return pd.DataFrame(a)
 
         a = A(example_positionfixes)
-        assert type(foo(a)) == B
+        assert type(foo(a)) is B
 
 
 class TestTrackintelGeoDataFrame:
@@ -363,30 +363,30 @@ class TestTrackintelGeoDataFrame:
         """Test if loc on all columns returns original class."""
         a = self.A(example_positionfixes)
         b = a.loc[[True for _ in a.columns]]
-        assert type(b) == self.A
+        assert type(b) is self.A
 
     def test_copy(self, example_positionfixes):
         """Test if copy maintains class."""
         a = self.A(example_positionfixes)
         b = a.copy()
-        assert type(b) == self.A
+        assert type(b) is self.A
 
     def test_merge(self, example_positionfixes):
         """Test if merge maintains class"""
         a = self.A(example_positionfixes)
         b = a.merge(a, on="user_id", suffixes=("", "_other"))
-        assert type(b) == self.A
+        assert type(b) is self.A
 
     def test_constructor_dataframe_fallback(self, example_positionfixes):
         """Test if _constructor gets DataFrame it falls through"""
         a = self.A(example_positionfixes)
         df = example_positionfixes.drop(columns=example_positionfixes.geometry.name)
-        assert type(a._constructor(df)) == pd.DataFrame
+        assert type(a._constructor(df)) is pd.DataFrame
 
     def test_constructor_calls_init(self, example_positionfixes):
         """Test if _constructor gets GeoDataFrame and fulfills test then builds class"""
         a = self.A(example_positionfixes)
-        assert type(a._constructor(a)) == self.A
+        assert type(a._constructor(a)) is self.A
 
 
 class TestTrackintelGeoDataFrameWithFallback:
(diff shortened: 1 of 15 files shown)


trackintel/model/trips.py
```diff
@@ -1,19 +1,25 @@
...
+    _register_trackintel_accessor,
+    TrackintelBase,
+    TrackintelDataFrame,
+    TrackintelGeoDataFrameWithFallback,
+)
 
 
 @_register_trackintel_accessor("as_trips")
-class TripsAccessor(object):
+def trips(*args, **kwargs):
```
NinaWie, line 18: So, the other option would be to make this a class, right? What would be the disadvantage of having this Trips class that handles the TripsDataFrame and TripsGeoDataFrame in the background?

trackintel/model/trips.py
```diff
@@ -101,13 +145,70 @@ def temporal_tracking_quality(self, *args, **kwargs):
...
-        return ti.preprocessing.trips.generate_tours(trips=self._obj, **kwargs)
+        return ti.preprocessing.trips.generate_tours(trips=self, **kwargs)
+
+
+# added GeoDataFrame and DataFrame manually afterwards such that our methods always come first
+class TripsGeoDataFrame(TrackintelGeoDataFrameWithFallback, TripsDataFrame, gpd.GeoDataFrame, pd.DataFrame):
+    """Class to treat a GeoDataFrame as collections of trips.
+
+    Requires at least the following columns:
+    ['user_id', 'started_at', 'finished_at', 'origin_staypoint_id', 'destination_staypoint_id']
```
NinaWie, line 164: shouldn't the geometry be a required column here?

trackintel/model/util.py
```diff
@@ -152,29 +152,103 @@ def _constructor(self):
...
+            else:
+                result.__class__ = self.fallback_class
+        return result
+
+    return wrapper
 
-    # so far we don't have a lot of methods here
-    # but a lot of IO code can be moved here.
+
+# short and memorizable name
```
NinaWie, line 182: Is this still a todo?

trackintel/model/trips.py
```diff
@@ -101,13 +145,70 @@ def temporal_tracking_quality(self, *args, **kwargs):
...
 
         See :func:`trackintel.preprocessing.trips.generate_tours`.
         """
-        assert len(args) == 0, "When calling 'generate_tours' via the accessor all arguments must be keyword arguments"
-        return ti.preprocessing.trips.generate_tours(trips=self._obj, **kwargs)
+        return ti.preprocessing.trips.generate_tours(trips=self, **kwargs)
+
+
+# added GeoDataFrame and DataFrame manually afterwards such that our methods always come first
+class TripsGeoDataFrame(TrackintelGeoDataFrameWithFallback, TripsDataFrame, gpd.GeoDataFrame, pd.DataFrame):
```
NinaWie, line 160: why is it subclassing from both gpd.GeoDataFrame and pd.DataFrame? I guess it's just for readability, but I would only put one of them

NinaWie (changes requested):
I had a look, thanks for the great work on this complicated issue due to the DataFrame / GeoDataFrame fallbacks. In response to your comment, I would still prefer to have a dedicated class Trips, so that the user does not need to care about the dataframe / geodataframe stuff. Some of my comments are related to this. Before changing everything to having a Trips class, I would however wait whether @hongyeehh  or @henrymartin1 agree to that.

trackintel/model/trips.py
NinaWie: I just read your PR description above again, and just to clarify: My suggestion now would be to keep all the code you have with the two classes, but having a Trips class, such that users can use the construtor of Trips and will automatically get an object either of type TripsDataFrame or TripsGeoDataFrame

trackintel/model/util.py
```diff
@@ -152,29 +152,103 @@ def _constructor(self):
...
+class TrackintelGeoDataFrameWithFallback(GeoDataFrame):
```
hongyeehh, line 183: Why do we need both `class TrackintelGeoDataFrame(GeoDataFrame)` and `class TrackintelGeoDataFrameWithFallback(GeoDataFrame)`? Do we have geodataframe that does not fall back?

trackintel/model/trips.py
hongyeehh: In addition to that, why do we need `gpd.GeoDataFrame` on top of `TrackintelGeoDataFrameWithFallback`? Because it is simpler to access the functions in `gpd.GeoDataFrame`?

trackintel/model/trips.py
hongyeehh: Having a `Trip` class that automates the fallback would be great (if possible). 
It will also be great if we can re-initialize a `gdf` to trackintel trips with `Trip(gdf)` once the geometry is (re-)added,

tests/preprocessing/test_trips.py
```diff
@@ -254,7 +254,7 @@ def test_accessor(self, example_trip_data):
         # generate tours using the accessor
         trips_acc, tours_acc = trips.as_trips.generate_tours()
 
-        pd.testing.assert_frame_equal(trips_expl, trips_acc)
+        pd.testing.assert_frame_equal(trips_expl, trips_acc, check_frame_type=False)
```
hongyeehh, line 257: Why do we need to add this line? The returns shall also be of the same datatype?

trackintel/model/trips.py
bifbof: It is not for readability. I added them to avoid some interactions on inheritance.
For example the `to_postgis` method. If we would have
````python
````
If I wouldn't added `gpd.GeoDataFrame` after `TripsDataFrame` we would have the following linear inheritance order:
`````
TrackintelDataFrameWithFallback -> GeoDataFrame -> ... -> TripsDataFrame -> DataFrame -> ...
`````
Meaning we inherit the `to_postgis` method from the

File: tests/model/test_util.py [hunk 7]
```diff
@@ -282,6 +285,67 @@ def foo(a: A) -> GeoDataFrame:
...
+    """Test if _wrapped_gdf_method_fallback conditionals work"""
+
+    def test_no_geodataframe(self, example_positionfixes):
+        """Test if function return value does not subclass DataFrame then __class__ is not touched"""
+
+        def foo(gdf: GeoDataFrame) -> pd.Series:
+            return pd.Series(gdf.iloc[0])
+
+        foo = _wrapped_gdf_method(foo)
+        assert type(foo(example_positionfixes)) is pd.Series
```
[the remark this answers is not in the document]
bifbof: Yes, higher order functions are really confusing. The one liner does not work as `_wrapped_gdf_method` takes and returns a function. So to test it, we need to specify a function.
But I have replaced `foo = _wrapped_gdf_method(foo)` with the equivalent `@_wrapped_gdf_method`, that hopefully makes the intention clearer. 
The line was a bug anyway, as it tested `_wrapped_gdf_method` and not `_wrapped_gdf_method_fallback`. 
I think my naming does not add to the understanding.

tests/model/test_util.py
```diff
@@ -325,6 +389,85 @@ def test_constructor_calls_init(self, example_positionfixes):
         assert type(a._constructor(a)) is self.A
 
 
+class TestTrackintelGeoDataFrameWithFallback:
+    """Test helper class TrackintelGeoDataFrameWithFallback."""
+
+    class A(TrackintelGeoDataFrameWithFallback):
+        """Mimic TrackintelGeoDataFrameWithFallback subclass by taking the same arguments"""
```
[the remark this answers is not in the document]
bifbof: To test the helper class on its own without adding potential bugs from `Trips`.

trackintel/model/trips.py
bifbof: Okay, I changed the code such that `Trips` does pretty much the same thing as `trips` but as a class.

@hongyeehh I don't understand your comment about the automated fallback. Isn't it already automated?
Automatically upcasting if a geometry is added may be a bit hard, but calling `Trips(gdf)` to get `TripsGeoDataFrame` should be trivial.

trackintel/model/trips.py
bifbof: So far we did not included `geometry` as required column in the other classes as well.
But I adapted the information of `MultiPoint` some lines further down, such that it is clear that a geometry is required.

trackintel/model/util.py
bifbof: Yes :). I'll open an issue to remind us to fix this.

trackintel/model/trips.py
bifbof: The `GeoDataFrame` is to avoid annoying inheritance problems, but the `DataFrame` was indeed not necessary and I have removed it.

trackintel/model/util.py
```diff
@@ -152,29 +152,103 @@ def _constructor(self):
...
+
+        return _constructor_with_fallback
+
+
+class TrackintelBase(object):
+    """Class for supplying basic functionality to all Trackintel classes."""
+
+    # so far we don't have a lot of methods here
+    # but a lot of IO code can be moved here.
+    pass
```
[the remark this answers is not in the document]
bifbof: Agree 👍🏻

trackintel/model/util.py
bifbof: Hmm, yes both do fallback. But the `TrackintelGeoDataFrameWithFallback` can additionally fall back to `fallback_class` if `_check` still succeed. 
I updated the docstring to explain the behavior better. 
Mirroring the geopandas fallback + adding an own fallback if the `_check` really increases the complexity. We could simplify the whole thing if we drop the fallback due to `_check` and accept that data integrity is part of the user responsibility.

tests/preprocessing/test_trips.py
bifbof: It added it because `trips_expl` is still a `GeoDataFrame` as the direct call of `ti.preprocessing.trips.generate_tours` does not change the type, while `trips.as_trips` creates a `TripsGeoDataFrame` object. Thus the types are different.

NinaWie (approved):

NinaWie (approved):
Thanks for the explanations and changes, from my side, it is fine now :)

hongyeehh (approved):
Thanks for the additional explanation. This is also ready from my side. Great effort!

PR #522, opened 2023-09-11
ENH: remove `is_datetime64tz_type`

bifbof:
Replaces all calls of `is_datetime64tz_type(obj)` with `isinstance(obj.dtype, pd.DatetimeTZDtype)`, because it is deprecated in pandas `2.1.0`.
Personally I don't understand why they changed that, but apparently it creates some minor performance increase.


hongyeehh (approved):
Thanks for the fix. In the original PR, the comment was "it is ambiguous (based on the name)", but I also did not get the reason :)

Issue #520, opened 2023-09-10
BLD: dependency updates causing tests warning and failing 


hongyeehh:

bifbof:
It seems that some test-jobs of the geopandas library are failing with the same error ([link](https://github.com/geopandas/geopandas/actions/runs/6143355162/job/16666654481)).
I'll try and have a look but it will hopefully be fixed by the geopandas people.

PR #524, opened 2023-09-12
BLD: replicate local environment.yml for CI

bifbof:
I cannot replicate the build error of #520.
Thus I try to change the `environment.yml` file of the test to see if the error remains.

### DIFF
--- a/.github/envs/39-latest-conda-forge.yaml
+++ b/.github/envs/39-latest-conda-forge.yaml
@@ -3,26 +3,21 @@ channels:
   - conda-forge
 dependencies:
 - python
-
-# required
-- matplotlib
 - numpy
-- pandas
+- matplotlib
+- geopandas>=0.12.0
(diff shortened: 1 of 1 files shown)


bifbof:
Ok this just worked :D
closes #520

hongyeehh:
Ahh great, so the change is to delete the `pandas` dependency and rely on its version requirement in `geopandas`?

bifbof:
> Ahh great, so the change is to delete the `pandas` dependency and rely on its version requirement in `geopandas`?

I think the reason was to move geopandas up in the chain. But to be honest, I neither know what caused this nor what fixed it. I simply copied the `environment.yml` file from the root.

hongyeehh:
Well, anyway, it works. The dependency issue is always a pain :)

hongyeehh (approved):

PR #526, opened 2023-09-12
CLN: fix deprecation warnings for `.idxmax` and `.loc[]`

bifbof:
For `idxmax` two things got deprecated.
1. the `axis` keyword, this was an easy fix
2. support of columns with only NaT, we relied on this behavior. Thus I needed to rewrite code to handle this.

Further deprecated is `df.loc[idx, "col"] = "string"`; it creates a new column that defaults to `float64`, that type then gets upcasted to `object` when `"string"` gets filled into the dataframe. To avoid this we have to create the column beforehand with `df["col"] = None`.

### DIFF
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -143,7 +143,7 @@ def pre_filter_locations(
         groupby_loc = ["location_id"]
     else:
         raise ValueError(f"Unknown agg_level '{agg_level}' use instead {{'user', 'dataset'}}.")
-    loc = sp.groupby(groupby_loc).agg({"started_at": ["min", "count"], "finished_at": "max", "duration": "sum"})
+    loc = sp.groupby(groupby_loc).agg({"started_at": [min, "count"], "finished_at": max, "duration": sum})
     loc.columns = loc.columns.droplevel(0)  # remove possible multi-index
     loc.rename(columns={"min": "started_at", "max": "finished_at", "sum": "duration"}, inplace=True)
     # period for maximal time span first visit - last visit.
@@ -214,7 +214,7 @@ def _freq_transform(group, *labels):
     pd.Series
         dtype : object
     """
-    group_agg = group.groupby("location_id").agg({"duration": "sum"})
+    group_agg = group.groupby("location_id").agg({"duration": sum})
     group_agg["purpose"] = _freq_assign(group_agg["duration"], *labels)
     group_merge = pd.merge(
         group["location_id"], group_agg["purpose"], how="left", left_on="location_id", right_index=True
--- a/trackintel/analysis/location_identification.py
+++ b/trackintel/analysis/location_identification.py
@@ -303,12 +303,12 @@ def osna_method(staypoints):
     sp_pivot = sp_agg.unstack()
     # get index of maximum for columns "work" and "home"
     # looks over locations to find maximum for columns
-    # use fillna to such that idxmax works on columns with only NaT
+    # use fillna such that idxmax raises no error on columns with only NaT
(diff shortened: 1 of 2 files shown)


trackintel/analysis/location_identification.py
```diff
@@ -301,24 +301,35 @@ def osna_method(staypoints):
...
-        sp_pivot.loc[redo_work["work"], "work"] = np.nan
-        sp_idxmax_work = sp_pivot.groupby(["user_id"])["work"].idxmax()
-        sp_pivot.loc[sp_idxmax_work.dropna(), "purpose"] = "work"
+    # use fillna to such that idxmax works on columns with only NaT
+    sp_idxmax = sp_pivot.fillna(pd.Timedelta(0)).groupby(["user_id"]).idxmax()
+
+    # preset dtype to avoid upcast (float64 -> object) in pandas (and the corresponding error)
+    sp_pivot["purpose"] = None
+    # assign empty index without any overlap
+    idx_work = idx_home = sp_pivot.iloc[0:0].index
```
hongyeehh, line 312: What is this line doing? The `.iloc[0:0]` makes it quite confusing.

trackintel/analysis/location_identification.py
```diff
@@ -301,24 +301,35 @@ def osna_method(staypoints):
...
-    # The "home" label could overlap with the "work" label
-    # we set the rows where "home" is maximum to zero (pd.NaT) and recalculate index of work maximum.
-    if all(col in sp_idxmax.columns for col in ["work", "home"]):
-        redo_work = sp_idxmax[sp_idxmax["home"] == sp_idxmax["work"]]
-        sp_pivot.loc[redo_work["work"], "purpose"] = "home"
-        sp_pivot.loc[redo_work["work"], "work"] = np.nan
-        sp_idxmax_work = sp_pivot.groupby(["user_id"])["work"].idxmax()
-        sp_pivot.loc[sp_idxmax_work.dropna(), "purpose"] = "work"
+    # use fillna to such that idxmax works on columns with only NaT
+    sp_idxmax = sp_pivot.fillna(pd.Timedelta(0)).groupby(["user_id"]).idxmax()
```
hongyeehh, line 307: Is the `fillna` necessary? What is the behavior when we do not add it? 
Or in other words, in the original comments, you state that there is a bug in pandas. Does it still exist?

bifbof: In future releases this line would raise a `ValueError`, this is also true for the bug mentioned.
They have changed this behavior to be consistent with `np.nanargmax` which is called in `idxmax`.
So the bug is fixed, but for that we need to make sure that no full-NaT groups exist.

trackintel/analysis/location_identification.py
bifbof: I thought `idx_work` and `idx_home` needed the right `MultiIndex` shape to be used in the intersection, but apparently this isn't the case.
I changed the line and updated the comment to make this clearer. :)

trackintel/analysis/location_identification.py
hongyeehh: Just to clarify: what do you mean by `In future releases`? 
I think the ValueError in `np.nanargmax` is what you are referring to, is this correct?

hongyeehh (approved):
This is ready for merging from my side. I only have one clarification issue left. After that, you can merge this PR.

trackintel/analysis/location_identification.py
bifbof: What I mean is that currently only a future warning is issued because the behaviour is deprecated. But in future releases this will change.
Yes exactly, for a pandas Series with only `np.nan` the ValueError in `np.nanargmax` does not trigger, which leads to this inconsistent behaviour.
Therefore, they will check and raise the error themselves.

PR #530, opened 2023-09-18
ENH: enable subclassing for tours

bifbof:
The last class. For reference see #490.

### DIFF
--- a/trackintel/model/tours.py
+++ b/trackintel/model/tours.py
@@ -1,10 +1,13 @@
 import pandas as pd
+
 import trackintel as ti
-from trackintel.model.util import _register_trackintel_accessor
+from trackintel.model.util import _register_trackintel_accessor, TrackintelBase, TrackintelDataFrame
+
+_required_columns = ["user_id", "started_at", "finished_at"]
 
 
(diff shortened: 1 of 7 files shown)


trackintel/model/tours.py
```diff
@@ -49,13 +55,24 @@ def _validate(obj):
...
     def to_csv(self, filename, *args, **kwargs):
         """
         Store this collection of tours as a CSV file.
 
         See :func:`trackintel.io.file.write_tours_csv`.
         """
-        ti.io.file.write_tours_csv(self._obj, filename, *args, **kwargs)
+        ti.io.file.write_tours_csv(self, filename, *args, **kwargs)
 
     def to_postgis(
```
hongyeehh, line 77: We need the `@_copy_docstring` here, right?

hongyeehh (commented):
Looks great from my side, just one small docstring question.

trackintel/model/tours.py
bifbof: Yes that is true 👍🏼

hongyeehh (approved):

PR #533, opened 2023-09-20
ENH: combine `TrackintelGeoDataFrame` and `TrackintelGeoDataFrameWithFallback` into one class

bifbof:
Instead of having two GeoDataFrame subclasses I combined them together into one with some conditionals that check if the `fallback_class` is set. This will hopefully make the code a bit better to read.

### DIFF
--- a/trackintel/model/trips.py
+++ b/trackintel/model/trips.py
@@ -10,7 +10,7 @@ from trackintel.model.util import (
     _register_trackintel_accessor,
     TrackintelBase,
     TrackintelDataFrame,
-    TrackintelGeoDataFrameWithFallback,
+    TrackintelGeoDataFrame,
 )
 
 
@@ -160,7 +160,7 @@ class TripsDataFrame(TrackintelBase, TrackintelDataFrame):
 
 
 # added GeoDataFrame and DataFrame manually afterwards such that our methods always come first
-class TripsGeoDataFrame(TrackintelGeoDataFrameWithFallback, TripsDataFrame, gpd.GeoDataFrame):
+class TripsGeoDataFrame(TrackintelGeoDataFrame, TripsDataFrame, gpd.GeoDataFrame):
     """Class to treat a GeoDataFrame as collections of trips.
 
     Requires at least the following columns:
--- a/trackintel/model/trips.py
+++ b/trackintel/model/trips.py
@@ -159,7 +159,7 @@ class TripsDataFrame(TrackintelBase, TrackintelDataFrame):
         return ti.preprocessing.trips.generate_tours(trips=self, **kwargs)
 
 
-# added GeoDataFrame and DataFrame manually afterwards such that our methods always come first
+# added GeoDataFrame manually afterwards such that our methods always come first
 class TripsGeoDataFrame(TrackintelGeoDataFrame, TripsDataFrame, gpd.GeoDataFrame):
     """Class to treat a GeoDataFrame as collections of trips.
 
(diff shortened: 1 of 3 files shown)


trackintel/model/util.py
```diff
@@ -208,21 +157,23 @@ def _constructor_with_fallback(*args, **kwargs):
             if isinstance(result, GeoDataFrame):
                 # cannot validate geometry as geometry column is maybe not set
                 return self.__class__(result, validate_geometry=False)
```
hongyeehh, line 159: To clarify: this then means the super constructor is a GeoDataFrame, but why for the self class we do not check the geometry? For a class to be the type of GeoDataFrame, you need to set the geometry column, no?

trackintel/model/util.py
```diff
@@ -189,6 +137,7 @@ class TrackintelGeoDataFrameWithFallback(GeoDataFrame):
     Fallback to DataFrame if looses geometry and _check fails.
```
hongyeehh, line 137: TO clarify: as we set fallback_class = None, the 1st and 3rd case at the moment is the same and correspond to the checks in line 154?

hongyeehh (commented):
Thanks for the simplification; this is definitely clearer than before. I only have some clarification questions. Considering the importance of the class change, I would request another maintainer for review.

trackintel/model/trips.py
```diff
@@ -160,7 +160,7 @@ def generate_tours(self, **kwargs):
 
 
 # added GeoDataFrame and DataFrame manually afterwards such that our methods always come first
```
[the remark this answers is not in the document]
bifbof: Yes good catch, I've removed it in the PR but forgot to adapt the comment.

trackintel/model/util.py
bifbof: A GeoDataFrame without **set** geometry is possible. In our `_constructor` we create a new GeoDataFrame without set geometry, thus we cannot validate the geometry, as we cannot access `gdf.geometry`.
The geometry column will only be set, if we get back to the geopandas function that created the `_constructor` call. One example for this would be the GeoDataFrame method [`merge`](https://github.com/geopandas/geopandas/blob/0eb2a5ecdc3b7b595e406c9f7bf1e6435ad39828/geopandas/geodataframe.py#L1556).

trackintel/model/util.py
bifbof: Your understanding is correct. I adapted the docstring to explain the change a bit better. Can you tell me if it is clearer now? :)

bifbof:
The changes look really big, but in reality I changed 6 lines and then deleted the unused code. :)

NinaWie (approved):
It looks okay to me, I mainly did a short test with one of the test data. As you see in the attached screenshot, it worked for my example, but I wondered why the type is a GeoDataFrame when using `ti.io.read_trips_csv`? Is this still a todo?
Screenshot 2023-09-21 at 17 25 06

bifbof:
Yes, the output of all generating functions is still a TODO with #535

hongyeehh:
I could not see new updates. Did you push the commits to this PR? @bifbof

bifbof:
> I could not see new updates. Did you push the commits to this PR? @bifbof

I thought I did. But apparently I forgot it.

trackintel/model/util.py
hongyeehh: Thanks for the example. This is clear now

hongyeehh (approved):

PR #537, opened 2023-09-22
ENH: rework geogr module

bifbof:
I had the following problem:
I wanted to import `Positionfixes` into the file `distances.py` to use the class there. 
But this created a circular import, `positionfixes.py` imports `util.py` which imports `distances.py` which now newly imports `positionfixes.py`, and with that python gave up.
To fix this I moved all the speed calculations out of `util.py` into `distances.py`, thus `util.py` does not have to import `distances.py` anymore. Further I have reworked the `geogr` module a bit. 
What is part of this PR:
- moved `get_speed_positionfixes` and `get_speed_triplegs` to `geogr/distances.py`
- moved `haversine_dist` to `geogr/distances.py` and renamed to `point_haversine_dist` to make clear that is still about points, even without the `ti.geogr.point_distance.haversine_dist` path.
- reimported all distance functions to `ti.geogr` such that we now can use `ti.geogr.check_gdf` vs. `ti.geogr.distances.check_gdf`.


bifbof:
I am sorry for these big PR. Normally I keep them under 400 LOC, but this one is mostly moves code around.

hongyeehh (approved):
Thanks for the rearrangement; this makes the code more structured, especially moving the speed calculation to geogr.

Issue #535, opened 2023-09-20
All generating functions should return Trackintel classes


bifbof:
Currently we return GeoDataFrames / DataFrames from function as `read_positionfixes_csv` etc.
They should return Trackintel classes.

PR #538, opened 2023-09-22
ENH: change copy direction for Positionfixes docstrings

bifbof:
**(This PR should not be reviewed before #537, as the changes are also included here)**

Still trying to fix the circular imports for #535.
The problem here is with `@_copy_docstrings`. For this function we import code from all I/O modules.
This creates the following circular import:`io/file.py` imports `Positionfixes` to return them as value, but `Positionfixes` already imports code from `io/file.py` to copy the docstring in `@_copy_docstring`.
To fix this, this PR includes the following changes:
- I/O functions now copy their docstring from the trackintel classes, so that all imports flow in one direction
- I opted to replace `_copy_docstring` with the new decorator `doc`, it is a really neat function that I copied from pandas
  It allows formatting docstrings nicely, like class method and a separate function definition to format the same docstring to their liking.
- Added variable `_shared_doc` to allow docstrings to be shared beyond pairs. Useful e.g. for `postgis_write` functions that already use the same docstring.

I have only included the changes for `Positionfixes` so as not to make this PR too large. The other classes will follow in another PR.


bifbof:
Will reopen this PR after #538 and #539 went through, like this it makes no sense.

PR #539, opened 2023-09-22
ENH: add doc decorator

bifbof:
Moved the introduction of @doc out of #538 to make the change more reviewable.
`@doc` is a decorator to take docstring templates, concatenate them and perform string substitution on them.


trackintel/model/util.py
```diff
@@ -236,3 +233,104 @@ def decorator(accessor):
...
+        return decorated
+
+    return decorator
+
+
+_shared_docs = {}
+
+# in _shared_docs as all write_postgis_xyz functions use this docstring
+_shared_docs[
+    "write_postgis"
```
hongyeehh, line 293: Should the docstring related to postgis be placed under `model.util`? I understand that we might need to add more docstring here later, but atm the docstrings for `.io` appear in `.model.util`.

trackintel/io/postgis.py
```diff
@@ -11,6 +11,7 @@
 
 import trackintel as ti
 from trackintel.io.util import _index_warning_default_none
+from trackintel.model.util import doc, _shared_docs
```
hongyeehh, line 14: This will not cause a circular import problem?

hongyeehh (commented):
Thanks for introducing the new doc() function. I am just still a bit confused about the import problem.

trackintel/model/util.py
bifbof: Yes, currently that is true. But I moved it here for a later PR, as all the model `to_postgis` methods need to import this docstring and for this it cannot appear in `.io`.

trackintel/io/postgis.py
bifbof: Hopefully not. :D It feels like untying a complicated knot, pulling at one place creates problems elsewhere.
The idea is that `trackintel.model.util` is not dependent on any `trackintel` module. 
This allows it to be imported anywhere without creating any circular dependencies.
Further `trackintel.model.positionfixes` ... `tours` etc. will only import `trackintel.model.util` thus being themselves circle free in any context. At least that is the idea.

hongyeehh (approved):

PR #540, opened 2023-09-23
ENH: remove all potential circular imports from Postfixes

bifbof:
This PR removes all imports that were necessary to copy docstrings to the Positionfixes class. 
The docstrings are still copied, but now in the other direction. 
On some level, I dislike this change, as it removes documentation from the code that relies on it, thus making the source code harder to read. On the other hand `@doc` enables to omit the first argument for all the class methods.
If you find the loss of readability worse than the `@doc` tricks, I would also be happy to simply duplicate the docstrings and add comments as a reminder that dependent docstrings should be updated accordingly

### DIFF
--- a/trackintel/geogr/distances.py
+++ b/trackintel/geogr/distances.py
@@ -284,11 +284,7 @@ def calculate_haversine_length(gdf):
 
 @doc(
     Positionfixes.get_speed,
-    first_arg="""
-Parameters
-----------
-positionfixes : GeoDataFrame (as trackintel positionfixes)
-""",
+    first_arg="\nParameters\n----------\npositionfixes : GeoDataFrame (as trackintel positionfixes)",
 )
 def get_speed_positionfixes(positionfixes):
     pfs = positionfixes.copy()
--- a/trackintel/geogr/distances.py
+++ b/trackintel/geogr/distances.py
@@ -11,6 +11,9 @@ import similaritymeasures
 from scipy.spatial.distance import cdist
 from sklearn.metrics import pairwise_distances
 
+from trackintel import Positionfixes
+from trackintel.model.util import doc
+
 
 def point_haversine_dist(lon_1, lat_1, lon_2, lat_2, r=6371000, float_flag=False):
     """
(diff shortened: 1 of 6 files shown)


trackintel/geogr/distances.py
```diff
@@ -325,25 +282,15 @@ def calculate_haversine_length(gdf):
     return np.bincount((index[:-1])[no_mix], weights=dist[no_mix])
 
 
+@doc(
+    Positionfixes.get_speed,
+    first_arg="""
+Parameters
+----------
```
hongyeehh, line 289: Do we need the `Parameters` line here? All other docstrings do not have it.

trackintel/model/positionfixes.py
```diff
@@ -98,60 +96,266 @@ def center(self):
...
+        --------
+        >>> calculate_distance_matrix(staypoints, dist_metric="haversine")
+        >>> calculate_distance_matrix(triplegs_1, triplegs_2, dist_metric="dtw")
         """
-        return ti.geogr.distances.calculate_distance_matrix(self, *args, **kwargs)
+        return ti.geogr.distances.calculate_distance_matrix(self, Y=Y, dist_metric=dist_metric, n_jobs=n_jobs, **kwds)
 
-    @_copy_docstring(get_speed_positionfixes)
-    def get_speed(self, *args, **kwargs):
+    # @_copy_docstring(get_speed_positionfixes)
```
hongyeehh, line 345: delete

hongyeehh (commented):
As far as I understand, this PR only moved around the docstrings and did not change any functionalities. The docstrings are now defined in each Class and copied into the actual realization functions. 
This also raises another point: we should organize our documentation and tutorials using classes, instead of currently centering around the processing functions. 

I am fine with the changes, just small typos to fix

trackintel/geogr/distances.py
bifbof: For to `pfs.get_speed()` we don't have any parameters so I excluded the section there.
But if we call it with ``ti.geogr.get_speed_positionfixes(pfs)` we now have a parameters and need the parameters section again. &rarr; thus in this special case we need the `Parameters`

hongyeehh (approved):

PR #549, opened 2023-09-26
ENH: vectorize Timestamp calculations

bifbof:
Vectorized some Timestamp calculations. I don't think it increases performance by a large margin, I just found it nice.

### DIFF
--- a/trackintel/analysis/tracking_quality.py
+++ b/trackintel/analysis/tracking_quality.py
@@ -102,7 +102,7 @@ def temporal_tracking_quality(source, granularity="all"):
     elif granularity == "weekday":
         # get the tracked week relative to the first day
         start_date = df["started_at"].min().floor(freq="D")
-        df["week"] = ((df["started_at"] - start_date) // 7).dt.days
+        df["week"] = ((df["started_at"] - start_date)).dt.days // 7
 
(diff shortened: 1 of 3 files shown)


trackintel/analysis/tracking_quality.py
```diff
@@ -100,17 +101,17 @@ def temporal_tracking_quality(source, granularity="all"):
 
     elif granularity == "weekday":
         # get the tracked week relative to the first day
-        start_date = df["started_at"].min().date()
-        df["week"] = df["started_at"].apply(lambda x: (x.date() - start_date).days // 7)
+        start_date = df["started_at"].min().floor(freq="D")
+        df["week"] = ((df["started_at"] - start_date) // 7).dt.days
```
hongyeehh, line 105: Does the change keep the original behavior? i.e., shouldn't it be applying `.dt.days` first and then `//7`?

bifbof:
Ready for review @hongyeehh. I also changed the `_localize_timestamp` function a bit, to work correctly in all cases like when we have some datetimes that have a tz and some that don't.

hongyeehh (approved):

hongyeehh:
Thanks, looks good for me

Issue #551, opened 2023-10-21
Enable empty DataFrames for Trackintel classes


bifbof:
I think we should allow empty Positionfixes, Staypoints, etc. And thus remove the check in `validate`.

PR #554, opened 2023-10-23
ENH: generating functions returns trackintel classes

### LINKED ISSUE #535: All generating functions should return Trackintel classes

bifbof:
closes #535 

All generating function return now trackintel classes by default.
Also added tests everywhere to check that. If I overlooked a function let me know :)

(Also contains commit from #550 because rebasing and squashing lead to many conflicts).


trackintel/preprocessing/positionfixes.py
```diff
@@ -164,7 +159,11 @@ def generate_staypoints(
     # user_id of sp should be the same as ret_pfs
     sp["user_id"] = sp["user_id"].astype(pfs["user_id"].dtype)
 
-    return pfs, sp
+    if len(sp) <= 0:
```
hongyeehh, line 162: How could it be `<0`? should it rather `==` for readability? (Similarly for other changes)

trackintel/preprocessing/positionfixes.py
```diff
@@ -164,7 +159,11 @@ def generate_staypoints(
...
+        warnings.warn("No staypoints can be generated, returning empty sp.")
+        return pfs, sp
```
hongyeehh, line 164: Do we need this line? Can we also use `Staypoints(sp)` even if `sp` is empty? This relates to #551.

trackintel/preprocessing/positionfixes.py
```diff
@@ -164,7 +159,11 @@ def generate_staypoints(
...
+
+    return pfs, Staypoints(sp)
```
hongyeehh, line 166: Should we write `Positionfixes(pfs)` here? (Similar to what is changed in `trackintel/preprocessing/staypoints.py`). I think `pfs` shall be already valid, so rechecking it is redundant, but it would be good to be consistent across functions

trackintel/preprocessing/staypoints.py
```diff
@@ -243,7 +240,7 @@ def merge_staypoints(staypoints, triplegs, max_time_gap="10min", agg={}):
 
     Returns
```
hongyeehh, line 241: There shall be a `Staypoints.validate(staypoints)` line for this function?

trackintel/preprocessing/staypoints.py
```diff
@@ -180,20 +180,17 @@ def generate_locations(
...
     # location_id of staypoints can only be in Int64 (missing values)
     sp["location_id"] = sp["location_id"].astype("Int64")
     # user_id of locs should be the same as sp
     locs["user_id"] = locs["user_id"].astype(sp["user_id"].dtype)
-
-    return sp, locs
+    if len(locs) <= 0:
+        warnings.warn("No locations can be generated, returning empty locs.")
+        return sp, locs
+    return Staypoints(sp), Locations(locs)
```
hongyeehh, line 193: See above comments; either keep the line as it is or change it to `return sp, Locations(locs)`

trackintel/preprocessing/trips.py
```diff
@@ -183,7 +184,7 @@ def generate_tours(
     # trips id (generated by this function) should be int64
     tours.index = tours.index.astype("int64")
 
-    return trips_with_tours, tours
```
hongyeehh, line 186: There should be a `Trip.validate(trips)` at the beginning of this function?

tests/preprocessing/test_positionfixes.py
```diff
@@ -554,3 +560,7 @@ def test_str_userid(self, example_positionfixes_isolated):
         # set user ID to string
         pfs["user_id"] = pfs["user_id"].astype(str) + "not_numerical_interpretable_str"
         pfs, _ = pfs.as_positionfixes.generate_triplegs()
+
+    def test_tpls_type(self, example_positionfixes_isolated):
```
hongyeehh, line 564: A short doctring is missing here

hongyeehh (commented):
Thanks for the detailed changes and PR. I only have supplementary comments, and you can merge once you think they are sufficiently addressed

trackintel/preprocessing/positionfixes.py
bifbof: Yes, `Staypoints.validate` will fail when it tries to access the first element for the geometry type.

trackintel/preprocessing/positionfixes.py
bifbof: The idea was not changing the input type of pfs if pfs is a GeoDataFrame. But as we checked `Positionfixes.valdiate(pfs)` I would be okay with either. What do would you expect as a user using this function?

trackintel/preprocessing/staypoints.py
bifbof: Yes, I think this is a good place. I added it.

trackintel/preprocessing/staypoints.py
bifbof: Here `sp` is created as a new object instance. So I used `Staypoints`. But I agree that this should be consistent with the other functions.

hongyeehh: okay, I see a line `sp= gpd.GeoDataFrame()` above in our code, so its fine to keep it like this

hongyeehh (approved):

PR #563, opened 2023-10-30
ENH: add radius of gyration

bifbof:
I did not know where to put the function, so I put it so far into a new file in `analysis`.
Added branch to calculate planar distances as well.

### DIFF
--- a/trackintel/analysis/metrics.py
+++ b/trackintel/analysis/metrics.py
@@ -1,6 +1,5 @@
-import tqdm
-import pandas as pd
 import numpy as np
+from tqdm import tqdm
 
 from trackintel.geogr import point_haversine_dist, check_gdf_planar
 
--- a/trackintel/analysis/metrics.py
+++ b/trackintel/analysis/metrics.py
@@ -1,0 +1,79 @@
+import tqdm
+import pandas as pd
+import numpy as np
+
+from trackintel.geogr import point_haversine_dist, check_gdf_planar
+
+
+def radius_gyration(gdf, method="count", print_progress=False):
+    """
+    Radius of gyration for individual users.
+
+    Parameters
+    ----------
+    gdf : Geodataframe
+        GeoDataFrame with column "user_id" and Point geometry.
+
+    method: string, {"count", "duration"}
+        Weighting for center of mass and average distance calculation.
+        - "count": assigns each Point the same weight of 1.
+        - "duration": assigns each Point a weight based on duration.
+           Additionally needs Timedelta column "duration" in gdf.
+
+    print_progress: bool, default False
+        Show per-user progress if set to True.
+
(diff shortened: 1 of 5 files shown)


trackintel/analysis/metrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def radius_gyration(gdf, method="count", print_progress=False):
+    """
+    Radius of gyration for individual users.
+
+    Parameters
+    ----------
+    gdf : Geodataframe
+        GeoDataFrame with column "user_id" and Point geometry.
```
hongyeehh, line 14: In trackintel, Rg only makes sense on the staypoint level - can we specify it here?

trackintel/analysis/metrics.py
```diff
@@ -0,0 +1,78 @@
...
+
+    References
+    ----------
+    [1] Gonzalez, M. C., Hidalgo, C. A., & Barabasi, A. L. (2008).
+    Understanding individual human mobility patterns. Nature, 453(7196), 779-782.
+
+    """
+    if method not in ["count", "duration"]:
+        raise ValueError(f'Method unknown. Should be on of {{"count", "duration"}}. You passed "{method}"')
+
```
hongyeehh, line 38: Rg also only makes sense if the sps are in the visited sequence. Since trackintel requires time stamp, I think we should add a `sort_values(by=["user_id", "started_at"])` here.

hongyeehh (commented):
Thanks for the addition. This is much neater than my original version :)
After fixing/answering some minor comments regarding input data, we are ready to merge!

trackintel/analysis/metrics.py
bifbof: I changed it to `Staypoints` everywhere. 
But isn't this metric applicable to Positionfixes and Locations as well?

trackintel/analysis/metrics.py
bifbof: Why should we order it? I don't understand the difference in just leaving it as it is.

hongyeehh: Second thought: we actually do not need to sort for rg. Sorry, Ignore this comment.

bifbof:
PR updated @hongyeehh :)

trackintel/analysis/metrics.py
hongyeehh: It applies to any traces, but the positions shall be with some meanings - I would therefore argue it does not apply for pfs. Locs do not contain frequency information - two visits to the same loc will only be saved once in the locs table - which is also not rg is measuring.

bifbof: okay, thank you for answering :)

hongyeehh (approved):

PR #567, opened 2023-11-01
ENH: Reexport functions from modules

### LINKED ISSUE #546: ENH: Re-export functions at module level
We still support direct access to our functions.
But sometimes the path is very long, like `ti.analysis.labelling.predict_transport_mode`.
We should re-export all public functions to module level so that we can access the above function with `ti.analysis.predict_transport_mode`.
This would reinforce the importance of the modules and not just make them a place where we put code.

bifbof:
closes #546
Changed how we export and import functions.
This makes the module paths clearer and additionally all links in the documentation work now as well.

### DIFF
--- a/benchmarks/preprocessing_benchmarks.py
+++ b/benchmarks/preprocessing_benchmarks.py
@@ -86,7 +86,7 @@ class BM_Generate_TRIPS:
 
     def common_func(self):
         """Generate Trips"""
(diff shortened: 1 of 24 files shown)


hongyeehh (approved):
Thanks for the detailed changes. It is great that the links in the docs are now working!

PR #568, opened 2023-11-01
DOC: replace accessor docs with new class style

### LINKED ISSUE #565: Remove mentions of pandas accessors from docs
In our docs we still talk about pandas accessors, although we work now on classes. 
We should update the documentation.

bifbof:
closes #565 
With the new class style mentioning the pandas accessor thing will lead to confusion. 
So I removed it from all documentation. I've done it rather quickly, so if somebody would take a second look at my formulation, I would be really happy about that. :)

### DIFF
--- a/docs/modules/model.rst
+++ b/docs/modules/model.rst
@@ -39,22 +39,21 @@ GeoPandas Implementation
 
 .. highlight:: python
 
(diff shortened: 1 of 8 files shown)


hongyeehh (approved):
I've double-checked and it seems you covered all occurrences in the doc. Thanks!

PR #570, opened 2023-11-08
ENH: Add gpx reader for single user

bifbof:
That was a lot simpler than I thought.
Currently we assign to all positionfixes `user_id = 0`.


bifbof:
It can generate them for Strava (I haven't tried Henry's) and at first glance the staypoints and triplegs seem fine.

PR #581, opened 2023-11-13
BUG: raise Error for missing files in read_gpx

### LINKED ISSUE #579: Raise error in gpx reader if path does not exist
We had a problem that the gpx reader function still executed even though the path did not exist, and then an error was only thrown in the last line `pd.concat(pfs_list, ignore_index=True)` with the error `no objects to concatenate`. There should be an error `Path does not exist` or so

bifbof:
closes #579


NinaWie (approved):
Perfect, thanks for the quick fix

PR #594, opened 2024-01-04
BUG: `calculate_distance_matrix` computes all pairs

### LINKED ISSUE #592: BUG: `calculate_distance_matrix` does not compute all distance pairs.
In `calculate_distance_matrix` we can either compare an `X` to itself or to another `Y`.
The code does some optimization for `X` in assuming that the computation is symmetric and only doing half of the pairs (e.g. `(x[0], x[1])` has the same result as `(x[1], x[0])`).
But the same thing is done if a `Y` is provided, here `(x[0], y[1])` can be different to `(x[1], y[0])`.  This code triggers the bug:
````python
import geopandas as gpd
import trackintel as ti
a = [(0, Point(0, 0)), (1, Point(1, 1))]
b = [(0, Point(10, 10)), (1, Point(20, 20))]
a = gpd.GeoDataFrame(a, columns=["id", "g"], geometry="g")
b = gpd.GeoDataFrame(b, columns=["id", "g"], geometry="g")
res = ti.geogr.calculate_distance_matrix(a, b)
assert res[0,0] != 0
````

Further the whole implementation is full of bugs and smaller issues, things I found at a quick glance:
- the `n_job` default value is non-valid value 0
- [this](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L139) assert compares a variable to itself
- [this](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L173) assignment makes no sense with non-symmetric matrices
- [this](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L181) else branch is unreachable, because we set `Y` to not None [here](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L136)
- [this](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L182) line forgets to pass `kwds`
- [this](https://github.com/mie-lab/trackintel/blob/e98354d662d6668316dba8d5f14c9448c9d8dc8b/trackintel/geogr/distances.py#L233) is a ValueError and not an AttributeError
- The docstring is confusing in multiple places

Further the test suite for this function is not useful. All except one test only compare the output of the function to different output of the function or to nothing at all. The one exception does not provide a `Y`.

### PR DESCRIPTION
**Pull request by bifbof**
fixes  #592
One of the added tests has a very generous relative tolerance, this is due to #593.
I made this function probably quite a bit slower. I thought that correctness is more important and if nobody experienced this bug in 3 years maybe it isn't called that often either.
Tried to use pairwise distance for everything. That wasn't possible for the LineStrings as sklearn and scipy both enforce float as data type. Thus I copied the code out of sklearn and adapted it a bit.

### DIFF
--- a/trackintel/geogr/distances.py
+++ b/trackintel/geogr/distances.py
@@ -114,7 +114,9 @@ def calculate_distance_matrix(X, Y=None, dist_metric="haversine", n_jobs=None, *
         For LineStrings, we provide the metrics {'dtw', 'frechet'} via the implementation from similaritymeasures.
 
     n_jobs: int, optional
-        Number of cores to use. Only used for Point-distances. Default uses all possible cores
+        The number of jobs to use for the computation. Ignored for LineStrings.
+        None means 1 unless in a joblib.parallel_backend context. -1 means using all processors.
+        See `sklearn.metrics.pairwise_distances` for more informations.
 
     **kwds:
         Optional keywords passed to the distance functions.
--- a/trackintel/geogr/distances.py
+++ b/trackintel/geogr/distances.py
@@ -1,3 +1,4 @@
+import itertools
 import math
 import multiprocessing
 import warnings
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: trackintel/geogr/distances.py [hunk 1]
```diff
@@ -82,44 +83,41 @@ def point_haversine_dist(lon_1, lat_1, lon_2, lat_2, r=6371000, float_flag=False
...
         'sokalsneath', 'sqeuclidean', 'yule']`
 
-        For triplegs, common choice is 'dtw' or 'frechet'. This function uses the implementation
-        from similaritymeasures.
+        For LineStrings, we provide the metrics {'dtw', 'frechet'} via the implementation from similaritymeasures.
 
     n_jobs: int, optional
-        Number of cores to use: 'dtw', 'frechet' and all distance metrics from `pairwise_distance` (only available
-        if only X is given) are parallelized. By default 1.
+        Number of cores to use. Only used for Point-distances. Default uses all possible cores
```
hongyeehh, line 117: `n_jobs=None `(default) means n_jobs=1 and not `all possible cores` (?), see [here](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)

hongyeehh (approved):
Thanks for the fix! The current version is much cleaner and easier to maintain. I only have a minor comment relating to n_jobs docstring

trackintel/geogr/distances.py
bifbof: Ah, yes you are right. I changed it

PR #596, opened 2024-01-07
ENH: remove .apply in predict_transport_mode

bifbof:
Just a small PR I came around to do. 
- replaced an `.apply` that assigns categories based on speed with the `pd.cut` function. This complicates the code a bit, but not that much. Moves basically all computation to pandas
- removed an error if the input dictionary isn't sorted and instead just sort the input.

### DIFF
--- a/tests/analysis/test_label.py
+++ b/tests/analysis/test_label.py
@@ -1,11 +1,9 @@
 import os
 
-import numpy as np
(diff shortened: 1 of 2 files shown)


hongyeehh (approved):
Thanks for the change. imo the changed version is also easy to understand

PR #599, opened 2024-01-14
ENH: Change AttributeError to ValueError

bifbof:
A bit of a pet peeve of mine.
All over trackintel we used `AttributeError` where `ValueError` would make more sense.
I changed it where it made sense for me.

for reference: https://docs.python.org/3/library/exceptions.html
>**AttributeError**
Raised when an attribute reference (see [Attribute references](https://docs.python.org/3/reference/expressions.html#attribute-references)) or assignment fails. (When an object does not support attribute references or attribute assignments at all, [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised.)

e.g.
```python
>> a = object()
>> a.not_existing_field
AttributeError: 'object' object has no attribute 'not_existing_field'
```
> **ValueError**
Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

e.g.
```python
>> chr(0x1100001)
ValueError: chr() arg not in range(0x110000)
```

### DIFF
--- a/tests/analysis/test_label.py
+++ b/tests/analysis/test_label.py
@@ -27,11 +27,11 @@ class TestCreate_activity_flag:
         sp_test = ti.read_staypoints_csv(sp_file, tz="utc", index_col="id")
 
         method = 12345
-        with pytest.raises(AttributeError, match=f"Method {method} not known for creating activity flag."):
+        with pytest.raises(ValueError, match=f"Method {method} not known for creating activity flag."):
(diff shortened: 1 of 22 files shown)


hongyeehh (approved):

hongyeehh:
Thanks for the formalization - these error treatments should be enforced in later enhancements and PRs

Issue #600, opened 2024-01-15
ENH: location generation improvements


hongyeehh:
Our [location_generation](https://github.com/mie-lab/trackintel/blob/4babae92f8d26abf92d80f8b19fd530f2cb4e7e2/trackintel/preprocessing/staypoints.py#L12) relies on the DBSCAN algorithm and enables the generation of `user-` and `dataset-level` locations. Locations are generated from `core` points from the DBSCAN results. When the input `epsilon` is small and locations only consist of 1 or 2 staypoints, we create a buffer with `epsilon` distance around the original point or line geometry to create valid polygon geometry for the column `extent`. This creates inconsistencies in the location results. For example, we create the following locations using `epsilon=20 and num_samples=2`: 

- **a.** locations created from multiple sps are directly connected 
- **b.** locations created from 2 sps are buffered

I suggest also buffering case **a** geometry to ensure consistency.

In addition, we occasionally observe huge locations created in dense sp regions, which overlap with many other locations (case **c** in the figure). We would need to check how these locations are generated and propose solutions to split/avoid them

![pic](https://github.com/mie-lab/trackintel/assets/17105123/4c2a2c3c-4da9-48f5-8ffe-a1cf62496d06)

PR #601, opened 2024-01-17
ENH: add buffer to all generated locs

hongyeehh:
partly solves #600 by adding a buffer to the extent of all generated locations


NinaWie:
I just checked again, the only thing I'm thinking about is that we could add it in the function docstring. Just to make sure that people understand what is returned in the "extent" column

NinaWie (approved):

PR #607, opened 2024-02-23
ENH: overlapping triplegs

munterfi:
This PR introduces a new method `overlap_staypoints` for generating triplegs.

Although the default method `between_staypoints` is suitable for most cases, there are instances where the temporal
resolution of positionfixes is very coarse with an irregular interval between the positionfixes, resulting in large gaps
between staypoints and triplegs. These are not actual missing data in the record, but rather a consequence of the coarse
temporal resolution. Such occurrences can complicate further analysis.

Our use case involves GPS data of locomotives with positionfixes at irregular intervals, ranging from a few seconds to
several hours when active, and possibly reducing to one positionfix per day when the locomotive is inactive for extended
periods.

### Adds

- Option `method="overlap_staypoints"` in `generate_triplegs()`.
- Separate test class `TestGenerate_triplegs_overlap`.
- Updates the relevant docstrings.

### Approach

The logic follows the `"between_staypoints"` method, which already sets the `"finished_at"` time of the staypoint to
the `"tracked_at"` timestamp of the following positionfix (which is the first positionfix of the next tripleg, and
thereby also defines its `"started_at"` time). The `"finished_at"` time of the tripleg is now also extended to
the `"tracked_at"` timestamp of the first positionfix of the next staypoint:

generate_triplegs_methods

To ensure a spatial overlap of the geometries, a new column `"tripleg_id_geom"` had to be introduced for the grouping
the points of the positionfixes into a linestring.

| id  | elevation | tracked_at                | user_id | accuracy | staypoint_id | tripleg_id | tripleg_id_geom | longitude  | latitude  |
|-----|-----------|---------------------------|---------|----------|--------------|------------|-----------------|------------|-----------|
| 591 | 44.196    | 2008-10-23 09:55:16+00:00 | 0       |          |              | 2          | 2               | 116.321781 | 40.009273 |
| 592 | 46.9392   | 2008-10-23 09:55:21+00:00 | 0       |          |              | 2          | 2               | 116.321871 | 40.009313 |
| 593 | 48.4632   | 2008-10-23 09:55:26+00:00 | 0       |          |              | 2          | 2               | 116.321952 | 40.009344 |
| 594 | 49.9872   | 2008-10-23 09:55:31+00:00 | 0       |          | 0            | 2          | 2               | 116.322007 | 40.009356 |
| 595 | 49.6824   | 2008-10-23 09:55:36+00:00 | 0       |          | 0            |            |                 | 116.322103 | 40.009386 |
[...]

PR #609, opened 2024-03-11
DOC/TST: generate tripleg

hongyeehh:
Following #607, this PR does the following:

- Clean up tests for positionfix generation
- Adding docstring to explain triplegs' time and geometry generation
- Adding more tests, in particular `test_stability()`, to ensure consistency between two cases

### DIFF
--- a/trackintel/preprocessing/positionfixes.py
+++ b/trackintel/preprocessing/positionfixes.py
@@ -208,15 +208,11 @@ def generate_triplegs(
 
     Notes
     -----
-    Methods 'between_staypoints' requires either a column 'staypoint_id' on the
-    positionfixes or passing some staypoints that correspond to the positionfixes!
-    This means you usually should call ``generate_staypoints()`` first.
+    The methods require either a column 'staypoint_id' on the positionfixes or passing some staypoints that correspond to the positionfixes! This means you usually should call ``generate_staypoints()`` first.
 
-    The first positionfix after a staypoint is regarded as the first positionfix of the
-    generated tripleg. The generated tripleg will not have overlapping positionfix with
-    the existing staypoints. This means a small temporal gap in user's trace will occur
-    between the first positionfix of staypoint and the last positionfix of tripleg:
-    pfs_stp_first['tracked_at'] - pfs_tpl_last['tracked_at'].
+    Following the assumptions in the function generate_staypoints(), to ensure consistency, the time extend and geometry for triplegs is defined as follows:
+        - 'between_staypoints': The generated tripleg will not have overlapping positionfix with the existing staypoints, thus triplegs' 'geometry' does not have common pfs as sp. 'started_at' is the timestamp of the first pf after a sp, and 'finished_at' is the time of the last pf before a sp. This means a temporal gap will occur between the first positionfix of sp and the last pf of tripleg: pfs_stp_first['tracked_at'] - pfs_tpl_last['tracked_at'] != 0. No temporal gap will occur between sp ends and tripleg starts (as per sp time definition).
+        - 'overlap_staypoints': The generated tripleg will have common start and end positionfix with the existing staypoints, thus triplegs' 'geometry' have common start and end pfs as sp. 'started_at' is the timestamp of the first pf after a sp (same as 'between_staypoints', to be consistent with generate_staypoints()), and 'finished_at' is the time of the first pf of a following sp. Temporal gaps will not occur between sp and triplegs.
 
     Examples
     --------
@@ -288,7 +284,7 @@ def generate_triplegs(
 
     # condition 2: Temporal gaps
     # if there is a gap that is longer than gap_threshold minutes, we start a new tripleg
-    cond_gap = pfs["tracked_at"] - pfs["tracked_at"].shift(1) > datetime.timedelta(minutes=gap_threshold)
+    cond_temporal_gap = pfs["tracked_at"] - pfs["tracked_at"].shift(1) > datetime.timedelta(minutes=gap_threshold)
 
     # condition 3: staypoint
     # By our definition the pf after a stp is the first pf of a tpl.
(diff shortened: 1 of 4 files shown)


hongyeehh:
A major change to the `_generate_triplegs_overlap_staypoints()` function is to assign pfs geometries to triplegs instead of the defined initially staypoint geometries (basically removing line 439-445 in `trackintel/preprocessing/positionfixes.py`). This ensures consistency between the two cases in `generate_triplegs()` (the testing function `test_stability()`). 

@munterfi, does this change affect your use cases or has any consequences for the original code? The tests seem to work fine after corresponding edits.

munterfi:
Hi @hongyeehh, thanks for extending the documentation and streamlining the test cases 👍 

Concerning the LINESTRING geometries of the triplegs: Since the method says the triplegs should overlap with the staypoints, an overlap with the centroid of the staypoint is more intuitive from my point of view. Even if I see the inconsistency between the methods.

What is the advantage if the tripleg starts at the last positionfix of the staypoint and ends at the first positionfix of the next staypoint, if we assume that it is a staypoint where the observed movement only results from GPS inaccuracies but not from real movement of the actual subject?

In our usecase with very coarse GPS tracking data of locomotives, we add a further step after applying trackintel, where we detect gaps in the "lifetime" tracks of the vehicles. Those gaps are then quantified spatially and temporally. If there is no gap, the duration and distance between the staypoint and tripleg should consequently be zero. This assumption would now no longer apply, given the modification.

What do you think?

hongyeehh:
> Hi @hongyeehh, thanks for extending the documentation and streamlining the test cases 👍
> 
> Concerning the LINESTRING geometries of the triplegs: Since the method says the triplegs should overlap with the staypoints, an overlap with the centroid of the staypoint is more intuitive from my point of view. Even if I see the inconsistency between the methods.
> 
> What is the advantage if the tripleg starts at the last positionfix of the staypoint and ends at the first positionfix of the next staypoint, if we assume that it is a staypoint where the observed movement only results from GPS inaccuracies but not from real movement of the actual subject?
> 
> In our usecase with very coarse GPS tracking data of locomotives, we add a further step after applying trackintel, where we detect gaps in the "lifetime" tracks of the vehicles. Those gaps are then quantified spatially and temporally. If there is no gap, the duration and distance between the staypoint and tripleg should consequently be zero. This assumption would now no longer apply, given the modification.
> 
> What do you think?

Thanks for the explanation! My primary concern is ensuring consistency in the code, e.g., obtaining the same tripleg generation results for cases 1 and 2. 

Assigning pfs geom is probably the simplest solution, as it preserves the one-to-one correspondence of the geometries of pf-sp and pf-tpl, and does not lead to tpl table depending on the sp generation result. I agree that assigning sp geometries to triplegs might make more sense, considering the real-world GPS noise. This is more of a design choice, and I am fine with both options as long as the code makes the `test_stability()` pass :). 

Maybe @NinaWie has a preference?

NinaWie:
I agree that there needs to be consistency between the cases whether sp are provided as input or not. But I also agree that it would be nicer to use the actual staypoint geometry for the `overlap` method, since the staypoint geometry should indeed overlap with the tripleg geometry in this case. To implement this, we would have to recompute the staypoint centroid from the pfs, right? So within the tripleg generation functions, in the case where no staypoints are supplied, we need to use the staypoint_id column in pfs and use it to compute the staypoint geometry? This is a bit ugly since it should be done in the staypoint generation. The other workaround would be to add an assertion statement that throws an error when you try to use the `overlap_staypoints` method in the tripleg generation without supplying staypoints.

hongyeehh (commented):
In this version, I returned to assign sp geometry to tpls start and end. The docstrings are changed accordingly. More checks are implemented at the beginning of the function; the "between_staypoint" method now requires staypoint as input. 

However, we still have the inconsistency issue between cases 1 and 2, because, in preprocessing of step 2 (line 268), we have not recovered the one-to-one match between pfs and sp, leading to wrong geometry assignments in line 458. 

We can either:
- Add more lines in the preprocessing to recover the match.
- Remove case 2, and require input pfs to have corresponding "staypoint_id" column with the input sp. 

I favor the latter, as I cannot think of its valid use case: users typically run the generate_triplegs() function after generate_staypoints(). Also, complicating preprocessing makes the function similar to generate_staypoints() as it needs to consider edge cases, e.g., temporal gaps.

munterfi:
Thank you for the suggestions!

I support excluding the case 2 entirely if the `overlap_staypoints` method is chosen.


A general input regarding the application scenario for case 2: Does this case imply that staypoints can be sourced differently, i.e. skipping the `generate_staypoints` step?

Given the high complexity and nesting within the `generate_triplegs` method, would it then make sense to extract case 2 into a separate function, e.g. `assign_pfs_to_sps(pfs, sp)`, and set the `staypoint_id` in the positionfixes as a prerequisite for the `generate_triplegs` function? 

Even though if this is a breaking change for users, it would reduce the input checks and variants that need to be managed, and assign a single responsibility to `generate_triplegs`.

trackintel/preprocessing/positionfixes.py
```diff
@@ -240,6 +233,17 @@ def generate_triplegs(
     # Case 2: False, pfs do not have a column 'staypoint_id' but staypoint are provided
     staypoints_exist = "staypoint_id" in pfs.columns
 
+    if staypoints is not None:
+        Staypoints.validate(staypoints)
+    if (staypoints is None) and (not staypoints_exist):
+        raise TypeError("staypoints input must be provide for pfs without staypoint_id column.")
+    if (staypoints is None) and (method == "overlap_staypoints"):
```
NinaWie, line 240: as discussed, add condition `staypoints_exist` here

hongyeehh:
In the new version, an error will be raised if `pfs` with no `staypoint_id` is provided to the function, such that we only support case 1 for overlap_staypoints.

I like the idea of removing case 2 preprocessing from the function, which we can implement in future PRs. At the moment, I added a deprecation warning for the users. Since it is not important the `print_progress` input argument is removed.

NinaWie (approved):
Thanks for the edits and cleanups! I also like the documentaiton, should be clear now.

PR #621, opened 2024-05-11
Fix deprecation warning osmnx

bifbof:
Fixes small deprecation warning for `osmx` that they deprecate passing `north, south, east, west` in favor for `bbox`


hongyeehh (approved):
Thanks for the fix!

PR #624, opened 2024-05-30
TST: capture expected warning

bifbof:
Just a small fix for `test_duplicate_pfs_warning`.
Was confused about some `empty_pfs` warning which turns out can just be captured and ignored as it is expected that it is empty.


hongyeehh (approved):
Thanks for the warning fix!

PR #626, opened 2024-06-29
TST: ignore `pygrio` default int32 type

bifbof:
Fixes the test failures.
Geopandas reached `1.0.0` 🎉 (hopefully things stabilize now...)
With that Geopandas defaults to `pyogrio` instead of `fiona`. Apparently `pygrio` reads integers as i32 if they are small enough and that differentiates to the `pd.read_csv` which always reads them as i64.
As this behavior is understandable, I choosed to ignore the errors.


hongyeehh:
thanks for the fix, merging now

hongyeehh (approved):

PR #638, opened 2025-02-24
[DOC] Fix incorrect import path of temporal_tracking_quality function in README example

ykangw:
The temporal_tracking_quality function was relocated to ti.analysis in a recent update of Trackintel. This commit updates the example in the 'How it Works' section of the README to reflect the correct function path, ensuring the example runs without errors.


hongyeehh:
Thanks for spotting the error and the fix! Merging the PR now

hongyeehh (approved):

PR #646, opened 2025-06-04
TST: fix inheritance class type error

### LINKED ISSUE #644: Schema validation in model constructors breaks standard GeoPandas >=1.1.0 operations (e.g., sort, merge_asof, sjoin)
#### Code Sample, a copy-pastable example

```python
import pandas as pd
import geopandas as gpd
from shapely.geometry import Point, Polygon
import trackintel as ti
import pytz

tz = pytz.timezone("UTC")

sp = pd.DataFrame({
    'user_id': [1, 1, 2],
    'started_at': pd.to_datetime(['2022-01-01 08:00', '2022-01-01 08:10', '2022-01-01 08:05']).tz_localize(tz),
    'finished_at': pd.to_datetime(['2022-01-01 08:05', '2022-01-01 08:15', '2022-01-01 08:10']).tz_localize(tz),
    'geom': [Point(0, 0), Point(0.1, 0.1), Point(0.2, 0.2)],
})
sp = gpd.GeoDataFrame(sp, geometry='geom', crs='EPSG:4326')
sp = ti.preprocessing.staypoints.Staypoints(sp)

pf = pd.DataFrame({
    'user_id': [1, 1, 2],
    'tracked_at': pd.to_datetime(['2022-01-01 08:02', '2022-01-01 08:12', '2022-01-01 08:07']).tz_localize(tz),
    'geom': [Point(0, 0), Point(0.1, 0.1), Point(0.2, 0.2)],
})
pf = gpd.GeoDataFrame(pf, geometry='geom', crs='EPSG:4326')
pf = ti.preprocessing.positionfixes.Positionfixes(pf)

# Try to sort positionfixes
try:
    pf = pf.sort_values("tracked_at")
except Exception as e:
    print("Exception during sort:", e)


# Try to merge_asof
try:
    merged = pd.merge_asof(
        pf.sort_values('tracked_at'),
        sp.sort_values('started_at'),
        by='user_id',
        left_on='tracked_at',
        right_on='started_at'
    )
    print(merged.head())
except Exception as e:
    print("Exception during merge:", e)


# Create a simple GeoDataFrame of polygons to join against
gdf_polygons = gpd.GeoDataFrame({
    'zone_id': [1],
    'geometry': [Polygon([(-1, -1), (1, -1), (1, 1), (-1, 1)])]
}, geometry='geometry', crs='EPSG:4326')

try:
    joined = gpd.sjoin(sp, gdf_polygons, how='inner', predicate='intersects')
    print(joined)
except Exception as e:
    print("Exception during spatial join:", e)
```

In this case, all raise the same error: 

> You are calling a geospatial method on the GeoDataFrame, but the active geometry column to use has not been set. 
There are columns with geometry data type (['geom']), and you can either set one as the active geometry with df.set_geometry("name") or access the column as a GeoSeries (df["name"]) and call the method directly on it.

#### Bug description
This is a regression in the trackintel models with the recent (today) geopandas 1.1.0 release. In the time between 1.0.1 and 1.1.0, [geopandas improved the ability to subclass `GeoDataframe`. ](https://github.com/geopandas/geopandas/commit/dc66966bbdd24b20ba8b60bb4ca8dad3a1e7c84b) This causes operations with Trackintel classes (as above) to break, due to the validation required in Trackintel model constructors.

#### Expected Output
I'm not sure off the top of my head the best way to address this, but I would think simple geopandas operations with trackintel models should be possible. Setting `validate=False` in all the models by default would resolve this, but then validation is off by default... Happy to help brainstorm ideas, 

#### Remarks
pip freeze in my environment:
```
certifi==2025.4.26
charset-normalizer==3.4.2
contourpy==1.3.2
cycler==0.12.1
fonttools==4.58.1
geoalchemy2==0.17.1
geopandas==1.1.0
greenlet==3.2.2
idna==3.10
joblib==1.5.1
kiwisolver==1.4.8
matplotlib==3.10.3
networkx==3.5
numpy==2.2.6
osmnx==2.0.3
packaging==25.0
pandas==2.2.3
pillow==11.2.1
pyogrio==0.11.0
pyparsing==3.2.3
pyproj==3.7.1
python-dateutil==2.9.0.post0
pytz==2025.2
requests==2.32.3
scikit-learn==1.6.1
scipy==1.15.3
shapely==2.1.1
similaritymeasures==1.2.0
six==1.17.0
sqlalchemy==2.0.41
threadpoolctl==3.6.0
tqdm==4.67.1
trackintel==1.3.2
typing-extensions==4.14.0
tzdata==2025.2
urllib3==2.4.0
```

hongyeehh:
closes #644

The issue is raised due to a recent update in the GeoDataFrame function `_geodataframe_constructor_with_fallback()` (see [here](https://github.com/geopandas/geopandas/commit/dc66966bbdd24b20ba8b60bb4ca8dad3a1e7c84b#diff-2bf1533512765e9a8c8ca4992b8bc2828c4f5567d9a853f160f1dbd0ea473c20L54)), where instead of the hardcoded `GeoDataFrame` (line 54), a class instance `cls` is passed to the function and implicitly initialized (line 2026). This means we cannot pass `self`, but instead need to explicitly pass and call `GeoDataFrame` to prevent initializing a trackintel class. The latter will raise an error because the geometry link is only assigned at the very end of the `GeoDataFrame._constructor_from_mgr()` function (see [here](https://github.com/geopandas/geopandas/blob/c36eba059f6170715b63a9aa065817c64025d263/geopandas/geodataframe.py#L2056)). 

The PR also contains a minor fix to explicitly convert sp into GeoDataFrame before processing, otherwise the `validate` function will be called multiple times.

### DIFF
--- a/trackintel/model/util.py
+++ b/trackintel/model/util.py
@@ -38,10 +38,9 @@ class TrackintelGeoDataFrame(GeoDataFrame):
     @property
     def _constructor(self):
         """Interface to subtype pandas properly"""
-        super_cons = super()._constructor
 
         def _constructor_with_fallback(*args, **kwargs):
-            result = super_cons(*args, **kwargs)
+            result = GeoDataFrame._geodataframe_constructor_with_fallback(*args, **kwargs)
             if isinstance(result, GeoDataFrame):
                 return self.__class__(result, validate=False)
             # uses DataFrame constructor -> must be DataFrame
@@ -53,7 +52,7 @@ class TrackintelGeoDataFrame(GeoDataFrame):
 
     def _constructor_from_mgr(self, mgr, axes):
         """Mirror GeoPandas _constructor_from_mgr method."""
-        return self._constructor(GeoDataFrame._constructor_from_mgr(self, mgr, axes))
+        return self._constructor(GeoDataFrame._constructor_from_mgr(GeoDataFrame, mgr, axes))
 
     # Following methods manually set self.__class__ fix to GeoDataFrame.
     # Thus to properly subtype, we need to downcast them with the _wrapped_gdf_method decorator.
--- a/trackintel/preprocessing/staypoints.py
+++ b/trackintel/preprocessing/staypoints.py
@@ -81,7 +81,7 @@ def generate_locations(
         raise ValueError(f"method '{method}' is unknown. Supported value is ['dbscan'].")
 
     # initialize the return GeoDataFrames
-    sp = staypoints.copy()
+    sp = gpd.GeoDataFrame(staypoints.copy())
     non_activities = None
     if activities_only:
         if "activity" not in sp.columns:


NinaWie:
Okay makes sense. Yes it's a bit mysterious right now why validate is triggered, but we can look at this separately.

